Issue #2, opened 2022-01-01
Use libuv fs_event to watch for changes in the cwd


cseickel:
Currently, this is handled using Buf* autocmds to detect when a file is changed in vim, but that misses events initiated outside outside of vim buffers, such as git actions.

It looks like nvim exposes a proper file watcher that we can use instead:
http://docs.libuv.org/en/v1.x/fs_event.html

`:h watch-file`

cseickel:
This is in main now. If anyone tries it out, please provide feedback!

```lua
      require("neo-tree").setup({
        filesystem = {
          use_libuv_file_watcher = false, -- This will use the OS level file watchers
          -- to detect changes instead of relying on nvim autocmd events.
        }
      })
```

PR #10, opened 2022-01-05
Add Installation

IHaveNoZpu:


cseickel:
Hi @IHaveNoZpu , thanks so much for contributing, but I am going to have to reject this one. Here's why:

1. There is already installation instructions in the Quickstart section below this one, and I don't feel this adds meaningful information. Paq usage is obvious from packer example, and no I don't think many people would use the Nix example you added.
2. If you read the CONTRIBUTING.md page, you'll see that I am asking PRs to go to the **main** branch.
3. Also, the commit message should be something like `docs: add installation section to readme`

#2 and #3 I could fix for you, but I just don't feel this is needed. If you disagree, start a Discussion or create an issue. If enough others agree, we can add more installation instructions.

Thanks again, and please don't hesitate to propose more changes in the future!

PR #24, opened 2022-01-09
Add nospell and nolist by default

danilshvalov:


plugin/neo-tree.vim
```diff
@@ -13,3 +13,6 @@ command! -nargs=? NeoTreeFloatToggle  lua require("neo-tree").float("<args>", tr
 command! -nargs=? NeoTreeFocusToggle  lua require("neo-tree").focus("<args>", true, true)
 command! -nargs=? NeoTreeShowToggle   lua require("neo-tree").show("<args>", true, true, true)
 command! NeoTreeRevealToggle lua require("neo-tree.sources.filesystem").reveal_current_file(true)
+
+set nospell
```
cseickel, line 17: This is not the place to set these. I'm pretty sure doing so here would affect the user's entire session. The places to apply the settings would be in the lua files at neo-tree.ui.renderer.create_window() and neo-tree.ui.popup.

Also, PRs go to main instaead of the v1.x release branch.

danilshvalov:
My bad. I'll fix it now.

danilshvalov:
Does it have to be configurable?

cseickel:
I don't think so, I can't image why someone would want spell turned on in a tree window. _Maybe_ they are doing something with list that they like, but I'd say in most cases it's better to let those few turn it back on for themselves with an autocmd on the `neo-tree` filetype.

cseickel (approved):

Issue #51, opened 2022-01-14
Not showing git status on filesystem source on Windows


tristone13th:
My configuration:
```lua
require("neo-tree").setup({
    enable_git_status = true,
    enable_diagnostics = true,
})
```
Actual behavior:
![c550de978b211a6d58df20db64f2981](https://user-images.githubusercontent.com/17382962/149449330-9de2967d-bb3b-42d2-ac8a-d2bc3ade5a4f.png)

PR #52, opened 2022-01-14
fix: git status doesn"t show on Windows

tristone13th:
#51


cseickel:
Hey @tristone13th, thanks for contributing! There are a few issues:

- PRs should go to main, I fixed that for you
- The space is meaningful, please remove the second commit from the PR
- It failed the StyLua check. I'm not sure why, but it seems to be the "remove space" commit anyway

The leading space would indicate that the changes have been added. It is a two character indicator. `[M ]` means you have changed it but not added it, `[ M]` means you have added the changes. 
https://git-scm.com/docs/git-status#_output 

If you'd like, you can submit another PR that adds a config option to the component to turn it into a single character indicator.

tristone13th:
> 

Thanks for the explanation, I will try to fix these issues soon.

PR #62, opened 2022-01-16
fix: #51 git status doesn't show on Windows

tristone13th:


cseickel (dismissed):

cseickel:
Thanks for updating. I see now why it is failing the StyLua merge check though, it needs to be two space indent instead of 4. Can you correct that?

tristone13th:
> 

OK, thanks for the detailed checking.

> Thanks for updating. I see now why it is failing the StyLua merge check though, it needs to be two space indent instead of 4. Can you correct that?

Thanks for the detailed checking, I have a new commit to update the indent now.

cseickel (approved):

PR #81, opened 2022-01-19
Refactor sources, extract common functionality into a `manager` module

cseickel:


cseickel:
@levouh take a look at this and tell me what you think. With these changes, The `neo-tree.sources.manager` would become the primary interface for the new command logic. Any refactoring we may need for the command work can happen at that level.

PR #83, opened 2022-01-20
fix(files): fixes #78, correctly handle .gitignore

### LINKED ISSUE #78: respect_gitignore is not working correctly
At least as how I would expect it to.

The `plenary.scandir` `respect_gitignore` flag filters out entries in the current scan IFF a .gitignore file is present in the root of the path it scans. The `do_scan` function in `fs_scan.lua` scans with a depth of 1, so each time a folder is opened, a new scan is done with the path of that directory, which has the effect that a .gitignore in the root directory isn't picked up.

An example would be
```
.config/nvim/
|- lua/
   |- packer_compiled.lua
|- .gitignore 
```
with .gitignore containing `lua/packer_compiled.lua`; `packer_compiled.lua` is still visible in the tree.

One solution is to create a gitignore filter in `fs_scan.lua` and do the filtering there. This can easily be done by just copying the method `plenary.scandir` uses.

Another one is just scan the whole tree at once and on any refresh and so on, but that might be detrimental when there are a lot of files.

I have implemented the first solution and it works well. Another benefit of doing it that way is that ignored files could be highlighted differently.

jgottzen:
Plenary's scandir only checks for .gitignore files in the directories it
scans. Since directories are only scanned when opened any .gitignore in
any parent directory isn't applied.

Implements handling of .gitignore files anywhere in the tree. The parsing
and checking logic is copied from the Plenary implementation.

### DIFF
--- a/lua/neo-tree/git/init.lua
+++ b/lua/neo-tree/git/init.lua
@@ -8,7 +8,8 @@ local M = {}
 local function execute_command(cmd)
   local result = vim.fn.systemlist(cmd)
 
-  if vim.v.shell_error ~= 0 or vim.startswith(result[1], "fatal:") then
+  -- An empty result is ok
+  if vim.v.shell_error ~= 0 or (#result > 0 and vim.startswith(result[1], "fatal:")) then
     return false, {}
   else
     return true, result
--- a/lua/neo-tree/git/init.lua
+++ b/lua/neo-tree/git/init.lua
@@ -1,0 +1,176 @@
+local Path = require("plenary.path")
+local utils = require("neo-tree.utils")
+
+local os_sep = Path.path.sep
+
+local M = {}
+
+local function execute_command(cmd)
+  local result = vim.fn.systemlist(cmd)
+
+  if vim.v.shell_error ~= 0 or vim.startswith(result[1], "fatal:") then
+    return false, {}
+  else
+    return true, result
+  end
+end
+
+local function windowize_path(path)
+  return path:gsub("/", "\\")
+end
+
+M.get_repository_root = function(path)
+  local cmd = "git rev-parse --show-toplevel"
+  if utils.truthy(path) then
+    cmd = "git -C " .. path .. " rev-parse --show-toplevel"
+  end
+  local ok, git_root = execute_command(cmd)
+  if not ok then
+    return nil
+  end
+  git_root = git_root[1]
+
+  if utils.is_windows then
+    git_root = windowize_path(git_root)
+  end
+
+  return git_root
+end
+
+local function get_simple_git_status_code(status)
+  -- Prioritze M then A over all others
+  if status:match("U") or status == "AA" or status == "DD" then
+    return "U"
+  elseif status:match("M") then
+    return "M"
+  elseif status:match("[ACR]") then
+    return "A"
+  elseif status:match("!$") then
+    return "!"
+  elseif status:match("?$") then
+    return "?"
+  else
+    local len = #status
+    while len > 0 do
+      local char = status:sub(len, len)
+      if char ~= " " then
+        return char
+      end
+      len = len - 1
+    end
+    return status
+  end
+end
+
+local function get_priority_git_status_code(status, other_status)
+  if not status then
+    return other_status
+  elseif not other_status then
+    return status
+  elseif status == "U" or other_status == "U" then
+    return "U"
+  elseif status == "?" or other_status == "?" then
+    return "?"
+  elseif status == "M" or other_status == "M" then
+    return "M"
+  elseif status == "A" or other_status == "A" then
+    return "A"
+  else
+    return status
+  end
+end
+
+---Parse "git status" output for the current working directory.
+---@return table table Table with the path as key and the status as value.
+M.status = function(exclude_directories)
+  local git_root = M.get_repository_root()
+  if not git_root then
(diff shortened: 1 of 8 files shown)


jgottzen:
This is the draft I mentioned in #78, cleaned up.

It checks any directory scanned for a .gitignore file and only creates a file item if the file/directory isn't ignored.

There are mainly three cases that isn't handled right now:

1. If the tree root is set to a sub-directory of a git repo, no git ignore filter is applied since none is found. This could easily be remedied by traversing upwards until a .git directory is found.
2. Git submodules are treated like any other directory, meaning the parent repo .gitignore will also apply.
3. Dotfiles-management, similar to 1, but since there will most likely be no .git directory upwards in the tree that solution will not suffice. Traversing upwards and stopping at the $HOME directory and apply any .gitignore there can be a solution.

cseickel:
1. To get the gitignore from a sub folder, I do have a function in the utils module which I just modified in a separate branch to accept a path other than cwd: https://github.com/nvim-neo-tree/neo-tree.nvim/blob/aae35cc6b64eee0651610bf4a2b9f73d889e4eac/lua/neo-tree/utils.lua#L171
2. I think using the above method would be effective for submodules as well, but I am not actively using any right now to test this with.
3. I'm not sure how a dotfiles management is different from any other git project. I use dotbot and it's just a standard git repo for me. Would a home directory gitignore take effect if a git project does not have it's own?

cseickel:
One major issue I notice with this is that it is storing the filters as global state. We can easily have multiple trees open in different tabs which each need their own state. I would make `gitignore.parse()` return the patterns table instead of storing it in the module, which can then be stored in the `state` object and passed back as an argument to the `is_ignored()` function.

jgottzen:
Yeah, will do.

I think manually parsing out .gitignore files is problematic, and will cause a lot of issues. Relying on git to do it is a lot safer, and easier, i.e. `git status --porcelain --ignored=matching`.

cseickel:
Absolutely, I think the `git status --porcelain --ignored=matching` is the way to go. I was worried about speed at first but I tried it out on my monorepo from work and it seems pretty efficient. 

You likely noticed this as well, but just to be sure: I noticed is that when a directory is ignored, it is only returning the directory and not the file contents, which is great because something like `node_modules` would be huge. This means that we probably build two kinds of filters out of this result, a file entry that is an exact match and a directory entry which is a prefix match.

Thanks for working on this, I think the end result is going to be great!

levouh:
>I think the git status --porcelain --ignored=matching is the way to go

+1 to this. I've seen `git status` take _ages_ and I am not entirely sure why, but usually doing something like `git fsck` or running it again (assuming there is some caching going on?) results in speedups for subsequent runs.

Regardless, it definitely seems better to let `git` do it as it will obviously handle things that the `plenary.nvim` implementation misses. I made [this](https://github.com/nvim-lua/plenary.nvim/pull/309) (rather naive) _PR_ to `plenary.nvim` as the one there doesn't handle global ignore. Whichever route you end up going here, supporting that would be awesome :wink:

jgottzen:
Changed the implementation to use `git status --ignored=matching` instead.

There's no need to use separate filters for directories and files, it can easily be handled as one filter.

I also moved the status flagging from `utils.lua` to the new `neo-tree/git/init.lua` file. But that should be consolidated into one call since they basically do the same thing, but that's for another time.

cseickel:
It looks great, but when I tried it out I keep getting this error on every navigate/refresh:
```
[neo-tree.nvim] [ERROR 12:27:24] ...packer/start/neo-tree.nvim/lua/neo-tree/events/queue.lua:144: Error in event 
handler for event before_render[filesystem. before_render]: vim/shared.lua:0: s: expected string, got nil
```

I'm not sure what causes it yet.

cseickel:
Of course, the problem went away when I added some log messages, and now it's not repeating. I will keep using this throughout my workday and see what happens.

jgottzen:
@levouh I'm not sure about `git status` taking a long time, but I've included `--no-optional-locks`, see git(1) GIT_OPTIONAL_LOCKS for more info.

cseickel:
I found the problem! The error shows reliably when in a repository with no changes. Classic Heisenbug, as soon as you start working on it, it goes away.

lua/neo-tree/git/init.lua
```diff
@@ -0,0 +1,176 @@
+local Path = require("plenary.path")
+local utils = require("neo-tree.utils")
+
+local os_sep = Path.path.sep
+
+local M = {}
+
+local function execute_command(cmd)
+  local result = vim.fn.systemlist(cmd)
+
```
cseickel, line 10: This should fix the bug with a clean git status:
```suggestion
  if #result == 0 then
    return true, {}
  end

```

cseickel (changes requested):

jgottzen:
Weird, that worked for me previously, but when I tested it now, I had the same error as you. I must have screwed up my testing somehow.

Just pushed a fix for it.

cseickel (approved):

cseickel (approved):

cseickel:
It is so awesome to finally have a functional gitgnore option. I don't think any of the trees I have used in the past did this correctly. I'll use the it for the day and release tonight if there are no issues.

Thanks @jgottzen !

PR #95, opened 2022-01-24
fix(render): Ensure *Show and *Focus commands maintain the previous cursor position within the tree

### LINKED ISSUE #94: Show and Focus actions should maintain the cursor location
@cseickel wanted to double check with you, but right now there isn't a way to "get the previous state from a tree", right?

Specifically consider the case where I reveal a file and close the tree. The tree itself has a state, but so does the cursor. If I then `action=focus` the tree again, the current action only makes an attempt to open and then focus the tree window. The cursor is left alone at the top of the buffer, rather than placed in the same place as it was when the window was previously closed.

I have a relatively simple way to handle this that I use in my own current configuration (basically tracking `Win{Leave,Enter}` for all `neo-tree` buffers and keep track of the cursor state), but wanted to make sure there isn't something that already exists first. The aforementioned method could break in the case that the tree state changes without one of those events firing (not a common thing as far as I'm aware).

Wanted to get your feedback before going any further with this.

_Originally posted by @levouh in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/31#issuecomment-1019544171_

levouh:
Fixes #94

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -317,7 +317,7 @@ M.set_expanded_nodes = function(tree, expanded_nodes)
   end
 end
 
-local create_tree = function(state)
+create_tree = function(state)
   state.tree = NuiTree({
     winid = state.winid,
     get_node_id = function(node)
@@ -356,7 +356,7 @@ local enable_auto_close_floats = function()
   auto_close_floats_is_set = true
 end
 
-local create_window = function(state)
+create_window = function(state)
   local winhl = string.format(
     "Normal:%s,NormalNC:%s,CursorLine:%s",
     highlights.NORMAL,
@@ -405,7 +405,7 @@ local create_window = function(state)
     win.source_name = state.name
     table.insert(floating_windows, win)
 
-    win:map("n", "<esc>", function(bufnr)
+    win:map("n", "<esc>", function(_)
       win:unmount()
     end, { noremap = true })
 
@@ -433,6 +433,16 @@ local create_window = function(state)
     vim.api.nvim_buf_set_var(state.bufnr, "neo_tree_winid", state.winid)
     vim.api.nvim_buf_set_var(state.bufnr, "neo_tree_tabnr", state.tabnr)
   end
+
+  -- Used to track the position of the cursor within the tree as it gains and loses focus
+  --
+  -- Note `WinEnter` is often too early to restore the cursor position so we do not set
+  -- that up here, and instead trigger those events manually after drawing the tree (not
+  -- to mention that it would be too late to register `WinEnter` here for the first
+  -- iteration of that event on the tree window)
+  win:on({ "WinLeave" }, function()
+    state.position.save()
+  end)
 
   win:on({ "BufDelete" }, function()
     win:unmount()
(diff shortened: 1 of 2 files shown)


lua/neo-tree/ui/renderer.lua
```diff
@@ -356,7 +356,7 @@ local enable_auto_close_floats = function()
   auto_close_floats_is_set = true
 end
 
-local create_window = function(state)
+create_window = function(state)
```
cseickel, line 359: Why remove the local here?

lua/neo-tree/ui/renderer.lua
```diff
@@ -510,7 +520,7 @@ end
 ---Draws the given nodes on the screen.
 --@param nodes table The nodes to draw.
 --@param state table The current state of the source.
-local draw = function(nodes, state, parent_id)
+draw = function(nodes, state, parent_id)
```
cseickel, line 523: Same here, without the local, isn't it global?

lua/neo-tree/ui/renderer.lua
```diff
@@ -317,7 +317,7 @@ M.set_expanded_nodes = function(tree, expanded_nodes)
   end
 end
 
-local create_tree = function(state)
+create_tree = function(state)
```
cseickel, line 320: Why remove the local here?

lua/neo-tree/ui/renderer.lua
```diff
@@ -543,6 +553,10 @@ local draw = function(nodes, state, parent_id)
     M.set_expanded_nodes(state.tree, expanded_nodes)
   end
   state.tree:render()
+
+  -- Restore the cursor position/focused node in the tree based on the state
+  -- when it was last closed
+  state.position.restore()
```
cseickel, line 559: If this works in all situations, it's an elegant solution. I want to test it locally and try to break it before merging.

lua/neo-tree/ui/renderer.lua
levouh: It is already declared as a `local` variable at the top of the file.

lua/neo-tree/ui/renderer.lua
levouh: Same as previous; it is declared local at the top of the file (I assume for a forward declaration, but did not look further into it).

cseickel:
There are errors when saving the state. I think it only happens when `follow_current_file` is enabled. At least it happens more when that is enabled. The errors are things like tree is nil, or invalid window id.

cseickel:
I think that last commit fixes the issues I was seeing. I'll test a little more. Let me know if you see any issues with the change.

levouh:
That's more than fine with me, thanks for doing checks with some extra settings.

Looking through the other options, something that comes to mind might be `bind_to_cwd`, where the tree state can change and when we try to focus the old node, it no longer exists. I would need to check it, but your checks might actually accommodate that. Will have to test.

levouh:
Will have to look at this later, but for me the first `NeoTreeReveal` call puts the cursor at the bottom of the window and does not properly reveal the current file.

This is without the changes you added @cseickel. Noting for now, will look into it later when I have a bit of time.

cseickel:
@levouh I think it the interaction between these three separate but related actions the set the current focus of the tree. I think the right thing to do here is to unify them all on the state.position methods you've added. I'm going to take a look at that later.

levouh:
Actually I was using the old version before the commit you added @cseickel, it has worked fine since then.

cseickel:
OK, I'm not seeing any problems either. I do want to refactor all the "set cursor" methods but I can do that separately. Merging it.

cseickel (approved):

Issue #77, opened 2022-01-18
Add testing


cseickel:
We really need to add a full suite of tests to this project, which can be run both locally and as a merge check. A handful of tests were created in #61 using plenary, which looks pretty good.

What needs to be tested?

- [ ] Every documented config option should do what it says
- [ ] Every bug fix should include a regression test (maybe just for future fixes?)
- [ ] All variations of vim commands (`:NeoTree...`) should do what the documentation says.

levouh:
I'll be working on #61 this weekend, and with the new source design it should be pretty trivial. Right now that one focuses on unit tests, however you can do functional tests with `plenary`'s version of `busted` as well.

PR #106, opened 2022-01-30
feat(tests): Add baseline testing using Plenary's Busted hooks

levouh:
@cseickel I split this out from the work I am doing on other branches as it seems to be in a relatively good state, and it would be beneficial to have sooner rather than later.

As an overview of what is here:

- Top level `Makefile` so you can just simply `make test` and run them as you please (also `make format` to run  `stylua`)
- Utility function(s) to setup/teardown a "test filesystem" to do whatever you want in
- Verification functions that help with the fact that a lot of the functions in `neo-tree` are async, so by default if we do something like `vim.cmd("NeoTreeFocus")` and then try to verify stuff, the verification is almost always going to fail as it happens too soon.

To show what this ends up looking like:

```lua
local util = require("tests.helpers.util")
local verify = require("tests.helpers.verify")

describe("Filesystem source command", function()
  local fs = util.setup_test_fs()

  after_each(function()
    util.clear_test_state()
  end)

  it("should reveal the current file in the tree", function()
    local testfile = fs.content[3].abspath
    util.editfile(testfile)

    local start_bufnr = vim.api.nvim_get_current_buf()

    vim.cmd("NeoTree filesystem action=reveal")
    verify.bufnr_is_not(start_bufnr)
    verify.tree_focused()
    verify.tree_node_is(testfile)
  end)

  it("should toggle the reveal-state of the tree", function()
    local testfile = fs.content[3].abspath
    util.editfile(testfile)

    local start_bufnr = vim.api.nvim_get_current_buf()

    vim.cmd("NeoTree filesystem action=reveal toggle=true")
    verify.bufnr_is_not(start_bufnr)
    verify.tree_focused()
    verify.tree_node_is(testfile)

    -- Wait long enough such that the tree _should have_ closed, then assert it is not focused anymore
    vim.cmd("NeoTree filesystem action=reveal toggle=true")
    verify.after(250, function()
      return #vim.api.nvim_tabpage_list_wins(0) == 1 and start_bufnr == vim.api.nvim_get_current_buf()
    end, "Failed to toggle the tree to a closed state with 'action=reveal'")
  end)
  
  util.teardown_test_fs()
end)
```

A few things to note:

- One thing right now is that the teardown stuff will only make sure that all buffers are wiped, but I am not sure that really is enough "cleanup" to ensure state from one test does not effect another. For example, the `state.position` metadata is not wiped. It is pretty easy to just wipe that from existence (the entire `state` object) however I'm not entirely sure what else might need to happen along with that. E.g. bad callbacks, leftover autocommands, etc. so wanted to get your opinion first before going to deep on that. It might be best to just expose a method for testing that "resets" all state (rather than wiping it completely) that any new functionality would need to abide by (meaning providing a way to be reset).
- There is no way to modify the current user configuration, as we only load `neo-tree` once in the `mininit.lua`, we could likely want a way to modify the configuration between different tests. Not sure how much of an effort this is at the moment.
- There are no actual examples of the tests here (other than the one above), so if you want I can push up my _super WIP_ examples that use them from #61 I am more than happy to. That branch also has some more unit-style tests as well, while the example above is more functional.
- I have zero clue about _Github_ actions, but you can definitely setup an action to ensure tests pass for merge requests.

This would be the first step towards #77

### DIFF
--- a/tests/helpers/util.lua
+++ b/tests/helpers/util.lua
@@ -1,0 +1,75 @@
+local utils = {}
+
+local Path = require("plenary.path")
+local testdir = Path:new(vim.env.TMPDIR or "/tmp", "neo-tree-testing"):absolute()
+
+local function rm_test_dir()
+  if vim.fn.isdirectory(testdir) == 1 then
+    vim.fn.delete(testdir, "rf")
+  end
+end
+
+utils.setup_test_fs = function()
+  rm_test_dir()
+
+  -- Need a list-style map here to ensure that things happen in the correct order.
+  --
+  -- When/if editing this, be cautious as (for now) other tests might be accessing
+  -- files from within this array by index
+  local fs = {
+    basedir = testdir,
+    content = {
+      {
+        name = "foo",
+        type = "dir",
+        content = {
+          { name = "foofile1.txt", type = "file" },
+          { name = "foofile2.txt", type = "file" },
+        },
+      },
+      { name = "bar", type = "dir" },
+      { name = "topfile1.txt", type = "file" },
+    },
+  }
+  local function makefs(content, basedir)
+    for _, info in ipairs(content) do
+      if info.type == "dir" then
+        info.abspath = Path:new(basedir, info.name):absolute()
+        vim.fn.mkdir(info.abspath, "p")
+        if info.content then
+          makefs(info.content, info.abspath)
+        end
+      elseif info.type == "file" then
+        info.abspath = Path:new(basedir, info.name):absolute()
+        vim.fn.writefile({}, info.abspath)
+      end
+    end
+  end
+  makefs(fs.content, testdir)
+  vim.cmd("tcd " .. testdir)
+  return fs
+end
+
+utils.teardown_test_fs = function()
+  rm_test_dir()
+end
+
+utils.clear_test_state = function()
+  -- TODO: Clear internal state?
+  vim.cmd("top new | wincmd o")
+  local keepbufnr = vim.api.nvim_get_current_buf()
+  for _, bufnr in ipairs(vim.api.nvim_list_bufs()) do
+    if bufnr ~= keepbufnr then
+      vim.api.nvim_buf_delete(bufnr, { force = true })
+    end
+  end
+  assert(#vim.api.nvim_tabpage_list_wins(0) == 1, "Failed to properly clear tab")
+  assert(#vim.api.nvim_list_bufs() == 1, "Failed to properly clear buffers")
+end
+
+utils.editfile = function(testfile)
+  vim.cmd("e " .. testfile)
+  assert.are.same(vim.fn.fnamemodify(vim.api.nvim_buf_get_name(0), ":p"), vim.fn.fnamemodify(testfile, ":p"))
+end
+
+return utils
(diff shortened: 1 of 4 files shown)


tests/helpers/util.lua
```diff
@@ -0,0 +1,75 @@
...
+end
+
+utils.setup_test_fs = function()
+  rm_test_dir()
+
+  -- Need a list-style map here to ensure that things happen in the correct order.
+  --
+  -- When/if editing this, be cautious as (for now) other tests might be accessing
+  -- files from within this array by index
+  local fs = {
```
levouh, line 19: Should there be a wrapper function for this, such that you can request a file from the tree in a simpler way (that is also more resistant to change in this list)?

Right now you have to:

```lua
local fs = utils.setup_test_fs()
local testfile = fs.content[3].abspath
```

which is obviously not the best thing because as soon as someone messes with the ordering here, all of those references break. One idea might be to give each file a "handle", and access it by the handle, so that a wrapper function can access it like:

```lua
local fs = utils.setup_test_ts()
local testfile = fs.get("top_level_file")
```

which would return a file that suffices the criteria that is being asked for. You could have handles like `"empty_dir"`, `"nested_file"` and whatnot which might make this a bit better? This turns into a "yo dawg" issue, where you need tests for your tests and so on.

tests/helpers/util.lua
```diff
@@ -0,0 +1,75 @@
...
+  vim.cmd("tcd " .. testdir)
+  return fs
+end
+
+utils.teardown_test_fs = function()
+  rm_test_dir()
+end
+
+utils.clear_test_state = function()
+  -- TODO: Clear internal state?
```
levouh, line 58: @cseickel thoughts on what is being done for cleanup here? It doesn't seem like quite enough as a lot of the internal state is still retained.

cseickel:
Thanks so much for this! You're right, we definitely need this now. I'll dig into this soon.

cseickel:
As far as your state concerns:

> It is pretty easy to just wipe that from existence (the entire state object) however I'm not entirely sure what else might need to happen along with that. E.g. bad callbacks, leftover autocommands, etc. 

I can't think of any issues with just wiping the state. I tried to keep everything as functional as possible with that single state object. Even where events are subscribed, they are based on configuration no state is part of those subscriptions. The only other thing to cleanup is the file watcher subscriptions, but that already has an `fs_watch.unwatch_all()` method.

> we could likely want a way to modify the configuration between different tests. Not sure how much of an effort this is at the moment.

So far, the `setup()` methods are all idempotent, so we can just call `require("neo-tree").setup(config)` repeatedly with different options set.

> There are no actual examples of the tests here (other than the one above),

I think that's enough, and i can always pop over to your fork and see what you have in the works

> I have zero clue about Github actions

Don't worry, I got that one!

cseickel (approved):

cseickel:
I think this is a good to merge, unless you want to add anything after reading my responses. I can add the state cleanup in another commit.

levouh:
Thanks for the feedback! I'm off for the night so whatever you want to add (either here or elsewhere) is up to you, I'll adapt this to any changes you make.

Let me change a few things tomorrow (might leave the state clearing to you but the other stuff is simple/useful enough) and rebase my current stuff to ensure the "harness" still works before merging, then I can ping you again for a follow up. Obviously no rush though. Thanks again @cseickel!

PR #107, opened 2022-02-01
feat: indent guides

danilshvalov:
I'm sorry for the delay. There wasn't enough time to finish.


lua/neo-tree/sources/common/components.lua
```diff
@@ -147,4 +148,71 @@ M.name = function(config, node, state)
...
+
+local indent_without_markers = function(config, node, state)
+  local level = node.level
+  local indent_size = state.indent_size or 2
+
+  if level == 0 or indent_size == 0 then
+    return {}
+  end
+
+  if not indent_cache[level] then
```
cseickel, line 161: I don't think this cache is worth the complication. What happens if the user changes the configuration? Then you need to check for that and clear the cache. Also, a user can have different indent settings per source. It would be better to just
```lua
return string.rep(" ", indent_size * level)
```

lua/neo-tree/sources/common/components.lua
```diff
@@ -147,4 +148,71 @@ M.name = function(config, node, state)
...
+  local indent_size = state.indent_size or 2
+  if indent_size == 0 then
+    return {}
+  end
+
+  local level = node.level
+  local with_markers = config.with_markers
+
+  if level < 2 or not with_markers then
+    return indent_without_markers(config, node, state)
```
cseickel, line 188: Actually, the entire function can probably be dropped and replaced with:
```suggestion
    return {
      text = string.rep(" ", indent_size * level),
      highlight = config.highlight or highlights.INDENT_MARKER
    }
```

cseickel (changes requested):

cseickel:
Thanks for making those changes. While testing ti out I realized there is one major item we forgot about:

> 2. We add a config option for a special component called "indent_renderer" which follows the interface of a component but is at the global config level.

I think it needs to be changed to this method of configuration to make it easier to configure and prevent breaking changes. With this strategy, the `indent_renderer` would be at the top level of the config and be implicitly put at the beginning of all renderers.

If you are short on time I can make these changes before merging, just let me know.

danilshvalov:
I have added a global configuration implementation. Now the user can globally configure any component. If the user manually changed the component, the configuration will be overridden (local configuration is more significant than the global one).

For example, user can configure icons:
```lua
require("neo-tree").setup({
      icon_renderer = {
          folder_closed = "-",
      },
     -- other configurations
})
```

cseickel:
I really like the idea of having a global config and we should definitely do that.

What this still doesn't address is the fact that it is a breaking change to take an implicit component and turn it into an explicit one. Everyone that has applied any customization to their renderers will suddenly lose indents with this update,

To prevent that, we have to either:

1. Remove it from the individual renderers completely and just pull the settings from global config instead. This is what happens now.
2. In setup(), check for the existence of the `"indent"` component in each config, and inject it only if it doesn't already exist.

Number two fits better with what you have now. It can probably happen in the `add_global_components_config` function. I would say test for the existence of "indent" in the loop and inject it in the first position only if it wasn't seen. It is is possible someone may have it but not in the first position because they want to have something like diagnostics or git_status all the way on the left before the indent.

lua/neo-tree/sources/common/components.lua
```diff
@@ -147,4 +147,41 @@ M.name = function(config, node, state)
   }
 end
 
+local prevent_indent = {}
```
cseickel, line 150: I didn't notice this before, but this shouldn't be global state. If you want to share some state between components in a render event, put it on the `state` object passed into the function.

lua/neo-tree/sources/common/components.lua
```diff
@@ -147,4 +147,41 @@ M.name = function(config, node, state)
...
+
+  if indent_size == 0 or level < 2 or not with_markers then
+    return { text = string.rep(" ", indent_size * level) }
+  end
+
+  local indent_marker = config.indent_marker or "│"
+  local last_indent_marker = config.last_indent_marker or "└"
+  local highlight = config.highlight or highlights.INDENT_MARKER
+
+  prevent_indent[level] = node.is_last_child
```
cseickel, line 165: Also, should this be named `skip_marker_at_level` instead? Is that what it is actually recording? As far as I can see, it is only being set once and it is entirely determined by whether `i == level and node.is_last_child`. If that is the case, there's no reason to have it all. it could be a local boolean variable or just insert the condition itself.

danilshvalov: I think `skip_marker_at_level` sounds better, I'll rename the variable. Also, I will put `skip_marker_at_level` in the state.

What about making the variable local. The previous level may affect the following levels. So if I use a local variable, I have the following problem:
problem

How it should be and how it works with a non-local variable:
normal

cseickel:
Hey @danilshvalov, I'm going to finish this up and merge it. The only outstanding issue was ensuring the component is always implicitly added even if the users have a custom renderer without an indent component.

cseickel:
Oops, didn't mean to close it.

cseickel:
Content moved to #131 so I could squash and rebase.

PR #159, opened 2022-02-26
fix: windows gitignore

### LINKED ISSUE #158: `filter.respect_gitignore = true` seems not working on WindowsOS
also when I set `filter.show_hidden = false` my `.gitignore` doesn't render, is this intended?
Edit: I did notice that just not works with directories

lopi-py:
fixes #158

### DIFF
--- a/lua/neo-tree/git/ignored.lua
+++ b/lua/neo-tree/git/ignored.lua
@@ -68,7 +68,7 @@ end
 M.is_ignored = function(ignored, path, _type)
   path = _type == "directory" and (path .. sep) or path
   for _, v in ipairs(ignored) do
-    if v:sub(-1) == utils.path_separator then
+    if v:sub(-1) == utils.path_separator or (utils.is_windows and _type == "directory") then
       -- directory ignore
       if vim.startswith(path, v) then
         return true
--- a/lua/neo-tree/utils.lua
+++ b/lua/neo-tree/utils.lua
@@ -347,13 +347,6 @@ M.resolve_config_option = function(state, config_option, default_value)
   end
 end
 
----The file system path separator for the current platform.
-M.path_separator = "/"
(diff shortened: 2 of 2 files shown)


lua/neo-tree/utils.lua
```diff
@@ -347,13 +347,6 @@ M.resolve_config_option = function(state, config_option, default_value)
   end
 end
 
----The file system path separator for the current platform.
```
cseickel, line 350: This part should not be removed, it is used in many places.

lua/neo-tree/git/ignored.lua
```diff
@@ -68,7 +68,7 @@ end
 M.is_ignored = function(ignored, path, _type)
   path = _type == "directory" and (path .. sep) or path
   for _, v in ipairs(ignored) do
-    if v:sub(-1) == utils.path_separator then
+    if v:sub(-1) == utils.path_separator or (utils.is_windows and _type == "directory") then
```
cseickel, line 71: How does this work if `utils.is_windows` was removed?

lua/neo-tree/utils.lua
lopi-py: it's duplicated, see
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/main/lua/neo-tree/utils.lua#L350
and
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/main/lua/neo-tree/utils.lua#L384

lua/neo-tree/git/ignored.lua
lopi-py: Yes, but it's just tested on windows, don't know if will work on linux or macOS

lua/neo-tree/utils.lua
cseickel: OH! Good catch!

cseickel (approved):
Thanks for contributing!

PR #164, opened 2022-03-01
feat: file nesting

### LINKED ISSUE #118: File Nesting a la JetBrains
### Discussed in https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/117

<div type='discussions-op-text'>

<sup>Originally posted by **bennypowers** February  3, 2022</sup>
Jetbrains (webstorm, etc) has a feature where file entries in the tree can collapse into one another. For example, a user can configure the tree to show `*.ts` files as a subtree, with siblings matching `$1.d.ts` `$1.js` and `$1.js.map` as the subtree's children:

example of nesting

This would be a cool feature in neo-tree. I'm wondering if maintainers and contributors think this could be easily accomplished in userland?</div>

lopi-py:
closes #118
image

### DIFF
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -6,6 +6,17 @@ local utils = require("neo-tree.utils")
 local renderer = require("neo-tree.ui.renderer")
 local log = require("neo-tree.log")
 
+---Gets the node parent folder recursively
+---@param tree tree to look for nodes
+---@param node node to look for folder parent
+---@return node
+local function get_folder_node(tree, node)
+  if node.type == "directory" then
+    return node
+  end
+  return get_folder_node(tree, tree:get_node(node:get_parent_id()))
+end
+
 local M = {}
 
 ---Add a new file or dir at the current node
@@ -13,16 +24,13 @@ local M = {}
 ---@param callback function The callback to call when the command is done. Called with the parent node as the argument.
 M.add = function(state, callback)
   local tree = state.tree
-  local node = tree:get_node()
-  if node.type == "file" then
-    node = tree:get_node(node:get_parent_id())
-  end
+  local node = get_folder_node(tree, tree:get_node())
+
   fs_actions.create_node(node:get_id(), callback)
 end
 
 M.close_all_nodes = function(state)
   renderer.collapse_all_nodes(state.tree)
-  state.tree:get_nodes()[1]:expand()
   state.tree:render()
 end
 
@@ -32,13 +40,13 @@ M.close_node = function(state, callback)
   local parent_node = tree:get_node(node:get_parent_id())
   local target_node
 
-  if node.type == "directory" and node:is_expanded() then
+  if node:has_children() and node:is_expanded() then
     target_node = node
   else
     target_node = parent_node
   end
 
-  if target_node then
+  if target_node and target_node:has_children() then
     target_node:collapse()
     tree:render()
     renderer.focus_node(state, target_node:get_id())
@@ -90,12 +98,7 @@ end
 ---@param callback function The callback to call when the command is done. Called with the parent node as the argument.
 M.paste_from_clipboard = function(state, callback)
   if state.clipboard then
-    local at_node = state.tree:get_node()
-    local folder = at_node:get_id()
-    if at_node.type == "file" then
-      folder = at_node:get_parent_id()
-    end
-
+    local folder = get_folder_node(state.tree, state.tree:get_node()):get_id()
     -- Convert to list so to make it easier to pop items from the stack.
     local clipboard_list = {}
     for _, item in pairs(state.clipboard) do
@@ -177,10 +180,20 @@ local open_with_cmd = function(state, open_cmd, toggle_directory)
     log.debug("Could not get node.")
     return
   end
-  if node.type == "directory" then
-    if toggle_directory then
+
+  local function open()
+    local path = node:get_id()
+    utils.open_file(state, path, open_cmd)
+  end
+
+  if utils.is_expandable(node) then
+    if toggle_directory and node.type == "directory" then
       toggle_directory(node)
     elseif node:has_children() then
+      if node:is_expanded() and node.type == "file" then
+        return open()
+      end
+
       local updated = false
       if node:is_expanded() then
         updated = node:collapse()
@@ -191,10 +204,8 @@ local open_with_cmd = function(state, open_cmd, toggle_directory)
         tree:render()
       end
     end
-    return nil
   else
-    local path = node:get_id()
-    utils.open_file(state, path, open_cmd)
+    open()
   end
 end
 
@@ -247,8 +258,6 @@ M.toggle_directory = function(state)
       updated = node:expand()
     end
     if updated then
-      tree:render()
-    else
       tree:render()
     end
   end
--- a/README.md
+++ b/README.md
@@ -78,14 +78,20 @@ use {
           indent = {
             indent_size = 2,
             padding = 1, -- extra padding on left hand side
+            -- indent guides
             with_markers = true,
             indent_marker = "│",
             last_indent_marker = "└",
             highlight = "NeoTreeIndentMarker",
+            -- expander config, needed for nesting files
+            with_expanders = nil, -- if nil and file nesting is enabled, will enable expanders
+            expander_collapsed = "",
+            expander_expanded = "",
+            expander_highlight = "NeoTreeExpander",
           },
           icon = {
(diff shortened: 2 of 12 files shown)


lua/neo-tree/sources/common/commands.lua
```diff
@@ -22,7 +22,6 @@ end
 
 M.close_all_nodes = function(state)
   renderer.collapse_all_nodes(state.tree)
-  state.tree:get_nodes()[1]:expand()
```
lopi-py, line 25: `render.collapse_all_nodes` also does this

cseickel:
Thanks for working on this! I tried it out and it works for some files, but not all. It seems like a pattern matching issue. Given this config:
```lua
        nesting_rules = {
          ["ts"] = {"d.ts", "js", "spec.ts", "html", "css", "scss" }
        },
```

..this will work:
```
test.ts
test.js
```

but this won't:
```
test-test.ts
test-test.js
```

-----

>  Maybe add it to README.md?

I think this is an interesting enough feature to highlight on the README.

> Handle entire name rules?

I think the extension logic is probably the right way to go, keeps it simple.

> How to handle expand and collapse (at the moment just keeps expanded all the time)

I would expect it to behave like a folder. Personally, I would also want them to be collapsed by default. We might need to replace some `node.type == "directory"` checks throughout the codebase with `node:has_children()` now that non-folders can be expanded.

lopi-py:
I'm thinking on adding a `link_resolver` module
> I would expect it to behave like a folder. Personally, I would also want them to be collapsed by default.

I thought it, but what if you want to open the file itself?, also I should add a component that tells you if the file have a nesting files
> Thanks for working on this! I tried it out and it works for some files, but not all. It seems like a pattern matching issue. Given this config:

and yes, I think the problem is [here](https://github.com/nvim-neo-tree/neo-tree.nvim/pull/164/files#diff-feffd4641239b2931092edb6184a28be846cb5b9aafd9a58ba36d67e1cc3256cR58), I'll look for all naming cases to add to the pattern

lopi-py:
image
pattern fixed 👍 (I did notice that git integration doesn't works with that names, lmao)

bennypowers:
FWIW, the default UI I'd expect would collapse these "dirs", but add a `+` icon next to (preferably overlaying, but I understand we're limited here) the file icon

cseickel:
> FWIW, the default UI I'd expect would collapse these "dirs", but add a `+` icon next to (preferably overlaying, but I understand we're limited here) the file icon

I noticed in the original request, there was a screenshot showing the use of both expand/collapse icons AND the item type icons:
image

That seems like the only way to do it. The `icon` component would need to be aware that nesting is enabled and switch to the combo expand/collapse icon with a second filetype icon or folder icon.

bennypowers:
Sounds legit

lopi-py:
Hi, sorry inactivity, then should I set a custom icon(instead of the file icon itself) if file is file has nesting files?

bennypowers:
> Hi, sorry inactivity, then should I set a custom icon(instead of the file icon itself) if file is file has nesting files?

sounds like there should be a new icon column to the left of the existing icon, which (for now) holds the disclosure widget (`>`)

cseickel:
> > Hi, sorry inactivity, then should I set a custom icon(instead of the file icon itself) if file is file has nesting files?
> 
> sounds like there should be a new icon column to the left of the existing icon, which (for now) holds the disclosure widget (`>`)

Yes, I think that's the way. When file nesting is enabled, the folder icon should look like a folder instead of the simple triangle, and anything that `has_children()`, like a folder or nested file container, should have the additional triangle style expand/collapse icon to the left of the normal icon.

lopi-py:
Sounds good

lopi-py:
It's ok if markers are disabled(if file nesting is enabled)?, looks a bit weird
without markers:
image
with markers:
image

cseickel:
I think we can just fix the markers so that they extend the extra space when they need to. I'm happy to work on that separately if you want to submit this as is.

lopi-py:
I'll try to fix it first, thanks

lopi-py:
@cseickel I had to put the arrows into the indent component, so is fine if the arrows uses the same highlight group as the indent markers or should have a specific highlight group(might be a breaking change because it will need to change the `highlight` to `marker_highlight` & `arrow_highlight` config for the `indent` component)?
image

cseickel:
I'm not crazy about having the indent component doing so many different things, but I do like the way that looks. I think in the future I might try to find a way to configure overlapping components to handle things like the indent / guides / nesting component. Definitely a problem for another day though...

For now, I think you can go ahead and add the two new highlight groups, but also keep the existing highlight group which can serve as the default for both.

```lua
  local marker_highlight = config.marker_highlight or config.highlight
  local arrow_highlight = config.arrow_highlight or config.highlight
```
That way there's no breaking change.

lopi-py:
Oh, yeah

lopi-py:
@cseickel could you check it if works fine?, then I just need add it to README.md and update the docs
also I have fixed the first issue of #186

cseickel:
I gave it a quick try and it works well. 

One thing I would change is that it didn't feel intuitive to have to turn on the `with_arrows` property in the indent component. I think that can be implicit and can just be set to true in `setup()` if the value is currently nil and `nesting_rules` is truthy. On the off chance that someone wants nesting but not expanders, they can explicitly set that value to false. This option can be a side note in the vim docs.

Also, on the same topic, I think I prefer the term `expanders` over `arrows`, because someone may use a different icon that doesn't happen to be an arrow.

All in all, this looks really awesome!

lopi-py:
> All in all, this looks really awesome!

Thanks!

> Also, on the same topic, I think I prefer the term `expanders` over `arrows`, because someone may use a different icon that doesn't happen to be an arrow.

Something like this?
```lua
require("neo-tree").setup({
  expander = {
    enabled = true, -- if nil and file nesting is enabled, will set to true
    expanded = "", -- expanded icon
    collapsed = "", -- collapsed icon
    highlight = "NeoTreeExpander",
  }
})
```
(still all the logic will be into the indent marker component)

cseickel:
> Something like this?
> 
> ```lua
> require("neo-tree").setup({
>   expander = {
>     enabled = true, -- if nil and file nesting is enabled, will set to true
>     expanded = "", -- expanded icon
>     collapsed = "", -- collapsed icon
>     highlight = "NeoTreeExpander",
>   }
> })
> ```

I was originally just thinking of replacing the word "arrow" with "expander" in the indent component config:
```lua
      with_arrows = true,
      arrow_collapsed = "",
      arrow_expanded = "",
```
changed to:
```lua
      with_expanders = true,
      expander_collapsed = "",
      expander_expanded = "",
```

...but having a completely different section is probably cleaner. That entire section can just be copied right onto the indent components config during setup.

I'm just not sure if the name "expander" is obvious enough in that context. Maybe "expander_icons" or "expander_controls"? @bennypowers do you have any thoughts on this?

lopi-py:
I have an idea, it's a bit complex(I think) but is making the `expander`(maybe `markers` too) component like other components but with "metadata"(`_target = "indent"`), then in setup copy it into de indent component, I think it will looks more cleaner

cseickel:
> I have an idea, it's a bit complex(I think) but is making the `expander`(maybe `markers` too) component like other components but with "metadata"(`_target = "indent"`), then in setup copy it into de indent component, I think it will looks more cleaner

YES!

lopi-py:
@cseickel check it again please(check `default.lua` too), btw the exntensions thing could be a migration, the sadly part is the documentation ;_;

bennypowers:
wrt naming, i don't mind "expander". another name for the top level config key that i think makes sense is "nesting"

cseickel:
Hey @lopi-py , sorry for all the back and forth on how to configure this. I think that the whole concept of hot to handle these overlapping elements is a bigger discussion that we should take our time with, and not hold up this feature. For now, can we just go back to the simple case of adding this config to the `indent` component and finish it off?

```lua
    indent = {
      indent_size = 2,
      padding = 1,
      -- indent guides
      with_markers = false,
      indent_marker = "│",
      last_indent_marker = "└",
      highlight = "NeoTreeIndentMarker",
      -- expander config, needed for nesting files
      with_expanders = true,
      expander_collapsed = "",
      expander_expanded = "",
      expander_highlight = "NeoTreeExpander"
    },
```

We can refactor later once discussion #191 has had some time to hopefully gather input from more people.

lopi-py:
@cseickel This can be merged I guess

cseickel (approved):

cseickel:
Looks good, thanks!

PR #169, opened 2022-03-03
add NeoTreeFloatTitle highlight group

hbiel:
This fixes the issue described in https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/35. I cleaned everything up and updated the documentation accordingly.

If there's something missing please let me know!

Best regards!

### DIFF
--- a/lua/neo-tree/ui/popups.lua
+++ b/lua/neo-tree/ui/popups.lua
@@ -17,7 +17,6 @@ M.popup_options = function(title, min_width, override_options)
   end
 
   local nt = require("neo-tree")
-  local popup_border_text = NuiText(" " .. title .. " ", highlights.FLOAT_TITLE)
   local popup_border_style = nt.config.popup_border_style
   local popup_options = {
     relative = "cursor",
@@ -28,7 +27,7 @@ M.popup_options = function(title, min_width, override_options)
     size = width,
     border = {
       text = {
-        top = popup_border_text,
+        top = NuiText(" " .. title .. " ", highlights.FLOAT_BORDER),
       },
       style = popup_border_style,
       highlight = highlights.FLOAT_BORDER,
@@ -49,7 +48,7 @@ M.popup_options = function(title, min_width, override_options)
       style = { "▕", blank, "▏", "▏", " ", "▔", " ", "▕" },
       highlight = highlights.FLOAT_BORDER,
       text = {
-        top = popup_border_text,
+        top = NuiText(" " .. title .. " ", highlights.TITLE_BAR),
         top_align = "left",
       },
     }
--- a/lua/neo-tree/ui/popups.lua
+++ b/lua/neo-tree/ui/popups.lua
@@ -17,7 +17,7 @@ M.popup_options = function(title, min_width, override_options)
   end
 
   local nt = require("neo-tree")
-  local popup_border_text = NuiText(" " .. title .. " ", highlights.TITLE_BAR)
+  local popup_border_text = NuiText(" " .. title .. " ", highlights.FLOAT_TITLE)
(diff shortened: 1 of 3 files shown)


cseickel:
This looks good, I'll just try it out later and I'll merge if there are no issues.

cseickel:
The problem with this is that I don't think the TitleBar highlight group looks good with a normal border. It tried using the FloatBorder group for the title and that's better but still not great. I think the right thing to do is to add another highlight group which uses the same background as float border but with the Normal foreground.

cseickel:
I just added a commit to show you what I mean, what do you think about that? Docs would still need to be updated.

hbiel:
I think the 'NC' style actually looks better on my initial commit. Although 'rounded' looks better on yours.
I'm using kanagawa.nvim as my theme so you might be getting different results.
Here is a comparison for both cases on my current setup:

'NC' style my commit: 6d46faf0eb43b46ead7c04ff54a3e39c10369d05_NC_border
'NC' style your commit: 093a1fe20abd7080b3b5a4e538327a48ba953ff7_NC_border

'rounded' style my commit: 6d46faf0eb43b46ead7c04ff54a3e39c10369d05_rounded_border
'rounded' style your commit: 093a1fe20abd7080b3b5a4e538327a48ba953ff7_rounded_border

I added another commit in which different highlight groups are used as each case actually needs different highlighting.
In 'NC' style the title border is drawn by the background highlighting of blank characters.
In the other styles the border is drawn by actual glyphs and as such the foreground highlighting is used.

It now looks like this for me:
'NC' style uses the established TitleBar highlight group and looks like this: NC style
'rounded' style uses the FloatBorder highlight group and looks like this (just like telescope does for me): rounded style
 
The additional highlight group isn't needed anymore as far as i can tell so i removed it.

Does this also look like this on your end or are you seeing different results?

cseickel:
I didn't even check the NC style because I didn't expect it to be affected. If you back out the changes you made to that branch then it should work out correctly.

I'd like to keep the added highlight group because I think it's something that others would want to be able to change. I know I personally don't like the way it looks when the title text is the same color as the border.  We can, however, make the default for that new group just be linked to the FloatBorder group so the default look is still what you are looking for.

lua/neo-tree/ui/popups.lua
```diff
@@ -44,12 +44,11 @@ M.popup_options = function(title, min_width, override_options)
 
   if popup_border_style == "NC" then
     local blank = NuiText(" ", highlights.TITLE_BAR)
-    local text = NuiText(" " .. title .. " ", highlights.TITLE_BAR)
```
cseickel, line 47: This should not have been removed

lua/neo-tree/ui/popups.lua
```diff
@@ -44,12 +44,11 @@ M.popup_options = function(title, min_width, override_options)
...
     popup_options.border = {
       style = { "▕", blank, "▏", "▏", " ", "▔", " ", "▕" },
       highlight = highlights.FLOAT_BORDER,
       text = {
-        top = text,
```
cseickel, line 52: This also should be reverted.

hbiel:
> I'd like to keep the added highlight group because I think it's something that others would want to be able to change. I know I personally don't like the way it looks when the title text is the same color as the border. We can, however, make the default for that new group just be linked to the FloatBorder group so the default look is still what you are looking for.

Yeah, that's nice to have. I have restored this in the latest commit and kept the defaults you set initially as they seem perfectly fine to me. I'm more than happy to be able to set my own colors in my personal configuration.

I have also added documentation for the new highlight group and have grouped the three groups together.

cseickel (approved):
Perfect, thanks!

PR #178, opened 2022-03-12
fix(files): set tree root correctly when revealing in split

marcuscaisey:
When using `:NeoTreeRevealInSplit`, the tree's root never changes from the working directory. This PR fixes that. I would have thought that `utils.split_path(path_to_reveal)[1]` would get the first out of the two things returned by `split_path` (like when you return tuples in Python), but it's actually indexing the first string returned instead (which returns `nil`).

### DIFF
--- a/lua/neo-tree/sources/manager.lua
+++ b/lua/neo-tree/sources/manager.lua
@@ -409,7 +409,7 @@ M.reveal_in_split = function(source_name, callback)
     cwd = M.get_cwd(state)
   end
   if cwd and not utils.is_subpath(cwd, path_to_reveal) then
-    state.path = utils.split_path(path_to_reveal)[1]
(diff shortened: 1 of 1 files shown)


cseickel:
> I would have thought that `utils.split_path(path_to_reveal)[1]` would get the first out of the two things returned by `split_path` (like when you return tuples in Python), but it's actually indexing the first string returned instead (which returns `nil`).

Huh, that's just nonsense, but I should have known that!

The issue with this PR is that it is targeting v1.x and PRs should go to **main**.  Can you recreate it? I tried ti change it but the protection does not update when I do that.

marcuscaisey:
> The issue with this PR is that it is targeting v1.x and PRs should go to **main**. Can you recreate it? I tried ti change it but the protection does not update when I do that.

aha! yeah will do, thanks

PR #179, opened 2022-03-12
fix(files): set tree root correctly when revealing in split

marcuscaisey:
Rework of #178 against `main` branch


cseickel (approved):

cseickel:
Thanks!

PR #210, opened 2022-03-27
Suppress non-critical errors on opening file (e.g E325)

smhc:


lua/neo-tree/utils.lua
```diff
@@ -317,7 +317,7 @@ M.open_file = function(state, path, open_cmd)
       return
     end
     if state.current_position == "current" then
-      vim.cmd(open_cmd .. " " .. escaped_path)
+      pcall(vim.cmd, open_cmd .. " " .. escaped_path)
```
cseickel, line 320: You can't just swallow errors without informing the user. Is there a particular error that could happen where the file is actually opened properly but vim outputs a useless error message anyhow? If there is, you need to handle that one specific case and then re-throw or log.error(...) if there was any other reason for the error.

lua/neo-tree/utils.lua
```diff
@@ -352,7 +352,7 @@ M.open_file = function(state, path, open_cmd)
         vim.cmd("vsplit " .. escaped_path)
         vim.api.nvim_win_set_width(winid, width)
       else
-        vim.cmd(open_cmd .. " " .. escaped_path)
+        pcall(vim.cmd, open_cmd .. " " .. escaped_path)
       end
     end
     events.fire_event(events.FILE_OPENED, path)
```
cseickel, line 358: By swallowing the error above, this FILE_OPENED event will now fire even if the file was not actually opened, creating more errors.

cseickel (commented):
Is E325 about the swap file? If it is and you are trying to ignore swapfile warnings, you should just turn it off.

smhc:
> Is E325 about the swap file? If it is and you are trying to ignore swapfile warnings, you should just turn it off.

I don't wish to turn the swap file off - I want to get the warning that the file is opened in another session or recovered from a crash (or instance shutdown) and to get the opportunity to open in read-only mode instead.

The error still occurs but it is presented by neovim rather than bubbled up to neo-tree to cause an error. We could check the status of the pcall and prevent the file opened event - but in the case of E325 the file is still opened. I'm not sure why that is an error but it seems like it requires every caller of this api to have a long list of valid and not-valid error codes to handle it gracefully.

cseickel:
> The error still occurs but it is presented by neovim rather than bubbled up to neo-tree to cause an error.

OK, in that case you'll just need to figure out how to positively identify that you are in this specific situation and not some other valid error relating to opening a file.

smhc:
I'm not sure how many more of these errors there are. Also selecting 'abort' on the prompt also throws e325 even though the file is not opened.

- Current implementation is broken as it presents an error and doesn't process the file open even
- Handling the specific e325 error is broken as it may process the file opened event even if it doesn't

Perhaps the file opened event should be driven by an autocmd to know when a file was really openend.

fyi nvm.tree had a similar bug and fix : https://github.com/kyazdani42/nvim-tree.lua/issues/880

I do think this is probably an neovim bug. There shouldn't be an error if the user presses 'Edit anyway' or 'Open readonly'. Or at the very least there should be different error codes depending on what they select.

smhc:
I've added handling for E325 and a 'print' of the error, but I'm not sure either is the correct thing to do.
Looking at the list of errors for neovim there are a number of 'harmless' errors - I doubt there is a 'did_this_error_block_the_file_open()' api call that can be used.

I'm also not sure triggering an 'error' or even a 'print' is useful either. e.g in the case of a selecting 'abort' on the swap file prompt. It's not an error from neo-tree - just something the user chose to do. 

I think the correct solution is to pass the result, err to the file_open event as parameters.

lua/neo-tree/utils.lua
```diff
@@ -349,13 +351,17 @@ M.open_file = function(state, path, open_cmd)
...
       else
-        vim.cmd(open_cmd .. " " .. escaped_path)
+        result, err = pcall(vim.cmd, open_cmd .. " " .. escaped_path)
       end
     end
-    events.fire_event(events.FILE_OPENED, path)
+    if (result or err == "Vim(edit):E325: ATTENTION") then
+      events.fire_event(events.FILE_OPENED, path)
+    else
+      print(err)
```
cseickel, line 363: There is an internal logging mechanism that should be used here, something like:
```suggestion
      log.error("Error opening file:", err)
```

cseickel (commented):
This looks good except for the one `print` to `log.error` change I requested. 

Also, PRs need to target `main` instead of the release branch. Can you change that and rebase if needed please?

smhc:
I created a new PR as I wasn't sure how to rework this one against a different branch.

I am still convinced that neo-tree should pass the result of the call through to the event handler rather than it trying to handle every perceivable 'harmless' error that neovim may generate. Or it should call a different handler such as 'file_opened_with_error' or something like that.

cseickel:
> I am still convinced that neo-tree should pass the result of the call through to the event handler rather than it trying to handle every perceivable 'harmless' error that neovim may generate. Or it should call a different handler such as 'file_opened_with_error' or something like that.

It seems unusual to have a "harmless" error. Are there other errors that could happen where the file would actually be opened correctly?

smhc:
I agree it is unusual - but as I mentioned above vim doesn't really have a way of indicating 'command succeed but with caveats' other than having an error. (although I would think a warning would be more appropriate). The only example I can see that might occur on open is E303, but this is far from an exhaustive search. There's a number that seem 'harmless' on close however.

I guess if we are now logging the error instead of "throwing" an error it will at least not interrupt the user. They'll just miss out on that file open event.

PR #214, opened 2022-03-27
fix: handle reveal_force_cwd correctly when cwd doesn't need to change

marcuscaisey:
This PR fixes a bug where the tree will not open if `reveal_force_cwd` is given but the cwd doesn't need to change. i.e. the cwd is already a parent of the file being revealed. As you can see from the changes, the bug was just caused by an if branch being taken and then not doing anything.

I've added a unit test for this case and also tested manually.


marcuscaisey:
diffed against the wrong branch again, oops

PR #215, opened 2022-03-27
fix: handle reveal_force_cwd correctly when cwd doesn't need to change

marcuscaisey:
This PR fixes a bug where the tree will not open if `reveal_force_cwd` is given but the cwd doesn't need to change. i.e. the cwd is already a parent of the file being revealed. As you can see from the changes, the bug was just caused by an if branch being taken and then not doing anything.

I've added a unit test for this case and also tested manually.

### DIFF
--- a/tests/neo-tree/command/command_current_spec.lua
+++ b/tests/neo-tree/command/command_current_spec.lua
@@ -1,3 +1,4 @@
+local Path = require('plenary.path')
 local util = require("tests.helpers.util")
 local verify = require("tests.helpers.verify")
 local config = require("neo-tree").config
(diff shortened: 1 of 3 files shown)


tests/neo-tree/command/command_current_spec.lua
```diff
@@ -50,14 +51,20 @@ describe("Command", function()
       verify.buf_name_is(testfile)
     end)
 
-    it("`:Neotree current reveal_force_cwd reveal_file=xyz` should reveal file current window", function()
+    it("`:Neotree current reveal_force_cwd reveal_file=xyz` should reveal file current window if cwd is not a parent of file", function()
       vim.cmd("cd ~")
-      vim.cmd("tcd ~")
-      vim.cmd("lcd ~")
```
marcuscaisey, line 56: removed these two lines as they have no effect

cseickel (approved):

cseickel:
Thanks!

PR #222, opened 2022-03-29
feat: support git base

lukas-reineke:
I'd like to add support for changing the git base.
With the current implementation it just defaults to `HEAD`, but there are cases where it would be nice to see the diff against other bases as well.
For example, when you review a pull request, or you are working on a feature branch and want to compare it against the main branch.

The implementation is not super complex, instead of using `git status . --porcelain=v1`, you use `git -C . diff --name-status HEAD --` where `HEAD` can be any commit/branch.
You get the same information, but the formatting of the output is a little bit different.

This is just a hacky PoC, so you can try with setting `vim.g.git_base` to different values. I wanted to ask if you are okay with this, before I make a proper PR.

Also `vim.g.git_base` needs to be handled correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ What to show. Can be one of:
 |--------|-------------|
 | filesystem | Show a file browser. DEFAULT |
 | buffers    | Show a list of currently open buffers. |
-| git_status | Show the ouput of `git status` in a tree layout. |
+| git_status | Show the output of `git status` in a tree layout. |
 
 #### `position`
 Where to show it, can be one of:
@@ -311,6 +311,22 @@ current file. For example:
 :Neotree /home/user/relative/path
 :Neotree dir=/home/user/relative/path
 :Neotree position=current dir=relative/path
+```
+
+#### `git_base`
+The base that is used to calculate the git status for each dir/file.
+By default it uses `HEAD`, so it shows all changes that are not yet committed.
+You can for example work on a feature branch, and set it to `main`. It will
+show all changes that happened on the feature branch and main since you 
+branched off.
+
+Any git ref, commit, tag, or sha will work.
+
+```
+:Neotree main
+:Neotree v1.0
+:Neotree git_base=8fe34be
+:Neotree git_base=HEAD
 ```
 
 #### `reveal`
@@ -367,7 +383,7 @@ that would open a directory will open neo-tree in the specified window.
 
 Neo-tree is built on the idea of supporting various sources. Sources are
 basically interface implementations whose job it is to provide a list of
-hierachical items to be rendered, along with commands that are appropriate to
+hierarchical items to be rendered, along with commands that are appropriate to
 those items.
 
 ### filesystem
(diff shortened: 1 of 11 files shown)


cseickel:
I was just thinking about this! I would sometimes prefer to show status as it relates to `origin` for the current branch, or from the default branch for the repo like `dev` or `main`.

The difficult part at this point is how to set the base?

1. It could be a global config that can be overridden in each source.
2. and/or could be set at runtime.
3. if you do set it at runtime, does it apply to all trees or just the one you set it for? Maybe there are two commands, one for global and one for this tree? or maybe that's over complicating it?
4. if you set it at runtime, do we just take text input 
5. or present a list of commits and let the user choose the base?
6. or we can have two or three pre-selected base choices and map commands that set them, like `gbo` to change git status base to origin, or `gbd` to change git status base to the default branch, or `gbh` to switch back to head?

lukas-reineke:
I think it should definitely be settable at runtime.

fugitive for example just takes it as an argument for commands.
gitgutter uses a global variable https://github.com/airblade/vim-gitgutter#the-base-of-the-diff

Either would work.

As for the input, just a string should be good enough. fugitive implements a completion function for this [fugitive#CompleteObject](https://github.com/tpope/vim-fugitive/blob/321328c6c5901a597348155fc0e83b800544dcb0/autoload/fugitive.vim#L2412), but it is pretty complex.

cseickel:
Ok, so we add a command to set the git base, which will default to showing the current value so it can double as a way to display the current base.

I would think that the scope of this setting should be either based on the current git root and apply to any tree within that repo, or the current tree, or the combination of the two. Current tree is the easy because you can just set it on the `state` arg that is passed to pretty much every function.

lukas-reineke:
It ended up being a pretty big PR :sweat_smile: 

Instead of adding a new command, I added it as an argument to `NeoTree`. I think this is the most seamless integration.
It works both with `:NeoTree git_base=master`, or just with `:NeoTree master`. I also implemented completion for it.

cseickel:
This looks like an awesome feature! I played around with this a little and the completion works well, but I noticed when I enabled the staged/unstaged symbols that everything shows as unstaged. I didn't have to time to dig in to this, but shouldn't they be staged if they were already committed?

image

```
:Neotree origin/main
```

image
(ignore the clipped icons, that's an issue with Windows Terminal)

lukas-reineke:
As far as I can tell, git diff does not have an option to show if a file is staged or not orz.

I had to split it into 2 commands, one runs the diff against whatever `git_base` is with `--staged`. The other runs diff just against the current unstated HEAD.
It should work now.

cseickel:
This was more than you bargained for, wasn't it? 😄 

It works well, except that if Neotree is already open, running `:Neotree <gitref>` does not refresh the tree with the new git base as I would expect. I have to press R to refresh to see the results of changing the git root. This can be handled in the `command.execute()` function like it does for changing the `dir`. 

This can also use a rebase against main now that you are almost done. If you've had enough, this is the one thing I need before I can merge, the other two I can add in myself.

It will also need documentation in the help file and README.

Thank you for this awesome new feature!

lukas-reineke:
Definitely was more work than I thought, but it was fun to implement. The code is structured well :+1: 

I rebased, implemented the refresh and wrote some docs

cseickel:
> but it was fun to implement. The code is structured well 👍

Thanks!

cseickel (approved):

PR #233, opened 2022-04-01
feat: add tabnew command for opening via a new tab.

tylersaunders:
I'd like to add support for opening a file from the filesystem source in a new tab.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ use {
             ["<cr>"] = "open",
             ["S"] = "open_split",
             ["s"] = "open_vsplit",
+            ["t"] = "open_tabnew",
             ["C"] = "close_node",
             ["<bs>"] = "navigate_up",
             ["."] = "set_root",
--- a/lua/neo-tree/sources/filesystem/commands.lua
+++ b/lua/neo-tree/sources/filesystem/commands.lua
@@ -94,6 +94,9 @@ end
 M.open_vsplit = function(state)
   cc.open_vsplit(state, utils.wrap(fs.toggle_directory, state))
 end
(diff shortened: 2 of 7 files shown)


lua/neo-tree/sources/filesystem/commands.lua
```diff
@@ -94,6 +94,9 @@ end
 M.open_vsplit = function(state)
   cc.open_vsplit(state, utils.wrap(fs.toggle_directory, state))
 end
+M.open_tabnew = function (state)
```
cseickel, line 97: This command should also be added to the `buffers` and `git_status` sources to be consistent with the other open commands.

README.md
```diff
@@ -128,6 +128,7 @@ use {
             ["<cr>"] = "open",
             ["S"] = "open_split",
             ["s"] = "open_vsplit",
+            ["t"] = "open_tabnew",
```
cseickel, line 131: I try to keep the README mappings as a reflection of the default mappings. to that end, can you add this to the `defaults.lua` file as well?

It should also be listed in the help file in `docs/`
See `:h neo-tree-mappings`

cseickel (commented):
Thanks for adding, I think it's a good idea to add this. Can yoiu just add the things mentioned in my review comments?

lua/neo-tree/sources/filesystem/commands.lua
tylersaunders: Done

README.md
tylersaunders: Done

cseickel (approved):
Looks good, thanks!

PR #240, opened 2022-04-05
Make async_scan scan each opened directory asynchronously

mauriciopoppe:
This is an implementation of the comment in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/236#issuecomment-1088210652, pasting the algorithm for future reference:

```
local function async_scan(context, callback /* new */)
  for every path in context.paths_to_load
    async_scan(path) ->
        scan.scan_dir_async(parent) -> 
        wait for on_exit callback ->  increase counter and call `callback` when all the paths in context_paths are done
```

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/fs_scan.lua
+++ b/lua/neo-tree/sources/filesystem/lib/fs_scan.lua
@@ -1,21 +1,25 @@
 -- This files holds code for scanning the filesystem to build the tree.
-local vim = vim
+local uv = vim.loop
+
 local renderer = require("neo-tree.ui.renderer")
 local utils = require("neo-tree.utils")
-local scan = require("plenary.scandir")
 local filter_external = require("neo-tree.sources.filesystem.lib.filter_external")
 local file_items = require("neo-tree.sources.common.file-items")
 local log = require("neo-tree.log")
 local fs_watch = require("neo-tree.sources.filesystem.lib.fs_watch")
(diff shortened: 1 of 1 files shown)


cseickel:
Oh cool, you completely dropped plenary's scan_dir from this. I'll give this a try and merge it if there are no problems. I do have a lot of other things up in the air though so this might sit a bit, now that we know this is not the primary slowdown.

lua/neo-tree/sources/filesystem/lib/fs_scan.lua
```diff
@@ -9,13 +10,17 @@ local log = require("neo-tree.log")
...
   local state = context.state
-  local paths_to_load = context.paths_to_load
-  local folders = context.folders
-
+  local scanned_folder = context.folders[current_dir]
+  if scanned_folder then
+    scanned_folder.loaded = true
+  end
   if state.use_libuv_file_watcher then
     local root = context.folders[dir_path]
```
cseickel, line 25: I think `dir_path` needs to be updated to `current_dir`. You'll need to test with the `use_libuv_file_watcher = true` to make sure it is working.

mauriciopoppe: good catch, thanks

cseickel (approved):

cseickel:
Looks good, thanks!

PR #252, opened 2022-04-08
performance work

cseickel:
- refactor: try to speed up render
- feat: allow disabling resize timer
- refactor: move all tree:render calls to a central location
- fix: turn off undo in tree buffers, fixes memory leak and maybe speed issues


nyngwang:
Thank you!

PR #270, opened 2022-04-12
fix possessive its

Liquidmantis:
Minor niggle on its vs it's contraction in README


cseickel:
Lol, thank you grammer police!

PR #274, opened 2022-04-13
236 gitignored slowdown

### LINKED ISSUE #236: filesystem redraws/delays
Hi, apologies for poor issue here but don't have much time at the moment so putting it here in case it will easily indicate what the problem is. Will try to update this issue with more meaningful information next week.

When opening filesystem tree there is a very small but noticeable delay (before window opens) during which my cursor blinks few times (I don't have blinking cursor in my terminal). Delay is of about half a sec or even less, hard to tell and am not sure how to measure it. For reference both git_status and buffers trees open immediately without blinking the cursor (or at least it happens so fast that I can't see it). I have tested nvim-tree, which opens immediately for the same file in the same repository so I doubt it's IO issue. Perhaps some redrawing for new conteiners? I am not sure if it was like that before though, can try to bisect next week.

cseickel:
- chore: add extra logging for giti ignored work, part of #236
- fix: remove git root lookup from git check ignored, may fix #236

### DIFF
--- a/lua/neo-tree/git/ignored.lua
+++ b/lua/neo-tree/git/ignored.lua
@@ -48,36 +48,29 @@ local get_root_for_item = function(item)
 end
 
 M.mark_ignored = function(state, items)
-  local git_roots = {}
+  local folders = {}
   log.trace("================================================================================")
   log.trace("IGNORED: mark_ignore BEGIN...")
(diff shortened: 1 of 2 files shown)


mauriciopoppe:
Thanks a lot! It's gotten way faster with lots of directories

PR #275, opened 2022-04-13
fix: missing signcolumn highlight group

jaeheonji:
Hi! recently I found one missing part while customizing the theme of the neo-tree.

neotree

Like the picture above, The background color of the neo-tree can be changed using `NeoTreeNormal`. And this overrides `Normal`.
However, because there were no highlight groups defined in the neo-tree for `SignColumn`, `NeoTreeNormal` was unable to create a consistent theme.


cseickel (approved):

cseickel:
Thanks!

PR #277, opened 2022-04-13
fix gitignore on Windows and File Nesting

lopi-py:
Tested on Windows

### DIFF
--- a/lua/neo-tree/git/ignored.lua
+++ b/lua/neo-tree/git/ignored.lua
@@ -6,20 +6,11 @@ local M = {}
 local sep = utils.path_separator
 
 M.is_ignored = function(ignored, path, _type)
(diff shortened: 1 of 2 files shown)


lua/neo-tree/git/ignored.lua
```diff
@@ -74,13 +65,22 @@ M.mark_ignored = function(state, items)
       result = {}
     end
 
+    -- on Windows, git seems to return quotes and double backslash "path\\directory"
```
cseickel, line 68: Since this part is Windows specific, can you wrap it in an `if utils.is_windows` block?

lopi-py: Done 👍

cseickel (approved):
Thanks!

Issue #314, opened 2022-04-21
split_with_window_picker


nagy135:
Hello, I m curious if this feature would be possible (or is already?).

Basically `w` (default bind) allows us to select where the window opens, but if I want it to appear in split, I have to create split, then close neo-tree, open it again (because that way the last split that had focus is that new split ...or simply make sure that I get to neo-tree from that new split) and then use `w` to place new file there.

Those are many steps I do quite often. Is there a better way? I would even suggest to take approach that nvim-tree took, and basically defaulting to pick destination every time when destination is not 100% determined, when you have more than one split opened.

The same thing should apply to vsplit.

Subjective opinion:
I understand that it has some advantages to have this picker as a optional dependency, but imho this should be default because the way it "chooses" where to open is something I have to keep in mind every time I open split. I even made `open_with_window_picker` my default opener, because that does exactly what I m proposing here with splits.

cseickel:
I think it would actually be a simple change to create a "split from chosen window" command. I'll put it on the list. There's probably enough information in the wiki to build one in your config as well if you want to take a shot at it.

PR #315, opened 2022-04-22
feat: use window_picker for split and vsplit as well

nagy135:
Adds option to bind split and vsplit versions of "open with window picker".
Not sure if we wanna bind this to some specific key, as I suggested this should be default behavior ...maybe it should be just suggested as an alternative to split/vsplit as a comment after ?
Have to admit that i m in no sense "lua programmer" so please do some testing, but on my setup it works flawlessly


nagy135:
#314

cseickel:
This looks good except for:

- You need to target the `main` branch instead of a release branch.
- You should document the new commands in `doc/neot-tree.txt` / `:h neo-tree-mappings`
- You might also want to include a reference to them in `deafults.lua` and `README.md`, even if they are just commented out by default.

nagy135:
it looks like I m little tired at the end of the week ...I also created it from release branch, stripping your netrw fix. Since rebase on main doesnt seem to register here after pull request is already created, I will create a new one fixing these fck-ups

PR #316, opened 2022-04-22
Feat/window picker split vsplit

nagy135:
#314 

Adds option to bind split and vsplit versions of "open with window picker".
Since these binds are replacement for split/vsplit, they are commented out so that users can choose.


cseickel (approved):

cseickel:
Looks good, thanks!

Issue #328, opened 2022-04-27
NeoVim 0.7 superseded VertSplit with WinSeparator


m-lima:
This means that https://github.com/nvim-neo-tree/neo-tree.nvim/blob/01d54bdc316aa36b4eefc8cc456f605cd2e267cc/lua/neo-tree/setup/init.lua#L77 does not work as intended with `VertSplit`.
A new (maybe transitional) hl needs to be created to work as an override of `WinSeparator`

cseickel:
Do you happen to have a link to any official documentation about this? I see it mentioned in the help but nothing about deprecation period or similar.

m-lima:
> Do you happen to have a link to any official documentation about this? I see it mentioned in the help but nothing about deprecation period or similar.

I don't have anything besides the help.
What I can add is that 0.7 added the `laststatus = 3` mode, which allows a single statusline. Because of this, the separators are not only vertical anymore, they can also be horizontal. Therefore, they needed a more fitting name for the hl
See also `fillchars` to see the new characters added for the horizontal split and the intersections

PR #329, opened 2022-04-27
Add NeoTreeWinSepartor hl

m-lima:
Fix https://github.com/nvim-neo-tree/neo-tree.nvim/issues/328

This PR only adds the `NeoTreeWinSeparator` so that pre 0.7 users can continue using their setup.
In the future, `NeoTreeVertSplit` could be considered for removal? And the local `winhighlight` option would override both `WinSeparator` and `VertSplit` with `NeoTreeWinSeparator`, thus allowing pre 0.7 and post 0.7 users to work

### DIFF
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -863,7 +863,7 @@ NeoTreeSignColumn         |hl-SignColumn| override in Neo-tree window.
 NeoTreeStatusLine         |hl-StatusLine| override in Neo-tree window.
 NeoTreeStatusLineNC       |hl-StatusLineNC| override in Neo-tree window.
 NeoTreeVertSplit          |hl-VertSplit| override in Neo-tree window.
-NeoTreeWinSeparator       |hl-WinSeparator override in Neo-tree window.
+NeoTreeWinSeparator       |hl-WinSeparator| override in Neo-tree window.
 NeoTreeEndOfBuffer        |hl-EndOfBuffer| override in Neo-tree window.
 NeoTreeRootName           The name of the root node.
 NeoTreeSymbolicLinkTarget Symbolic link target.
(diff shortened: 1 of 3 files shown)


doc/neo-tree.txt
```diff
@@ -863,6 +863,7 @@ NeoTreeSignColumn         |hl-SignColumn| override in Neo-tree window.
 NeoTreeStatusLine         |hl-StatusLine| override in Neo-tree window.
 NeoTreeStatusLineNC       |hl-StatusLineNC| override in Neo-tree window.
 NeoTreeVertSplit          |hl-VertSplit| override in Neo-tree window.
+NeoTreeWinSeparator       |hl-WinSeparator override in Neo-tree window.
```
cseickel, line 866: This is missing a closing `|` at the end of the link.

cseickel (commented):
This needs to target `main` instead of `v2.x`.

doc/neo-tree.txt
m-lima: 🤦

lua/neo-tree/setup/init.lua
```diff
@@ -297,7 +297,7 @@ local function merge_global_components_config(components, config)
   return merged_components
 end
 
-local merge_renderers = function (default_config, source_default_config, user_config)
+local merge_renderers = function(default_config, source_default_config, user_config)
```
m-lima, line 300: This change was just sumneko-lua auto-formatting the code. I suppose it is ok to keep it?

cseickel: That's as it should be. I'm not sure how these formatting issues sneak in, but it should be without the space.

cseickel (approved):

cseickel:
Thanks!

PR #345, opened 2022-05-04
Change `git commit` quotes context

jorgectf:
By using single quotes, commit messages can contain special characters like ` (backtick) without being interpreted.

For example, providing [this is a \`test\`] without brackets results now in [this is a ] instead of [this is a `test`].


cseickel (approved):
Thanks for making Neotree better!

PR #363, opened 2022-05-28
Fixed problems with shellslash

andreadev-it:
These changes should fix most of the issue about using shellslash on windows. However, maybe it's better to wait for [this PR on plenary](https://github.com/nvim-lua/plenary.nvim/pull/371) to be merged before accepting this. Without the changes on plenary, many things will just not work anyway.

### DIFF
--- a/lua/neo-tree/utils.lua
+++ b/lua/neo-tree/utils.lua
@@ -707,7 +707,10 @@ M.is_expandable = function(node)
 end
 
 M.windowize_path = function(path)
-  return path:gsub("/", "\\")
+  if not vim.o.shellslash then
+    return path:gsub("/", "\\")
+  end
+  return path
 end
 
 M.wrap = function(func, ...)
--- a/lua/neo-tree/utils.lua
+++ b/lua/neo-tree/utils.lua
@@ -511,7 +511,13 @@ end
 ---The file system path separator for the current platform.
 M.path_separator = "/"
 M.is_windows = vim.fn.has("win32") == 1 or vim.fn.has("win32unix") == 1
-M.use_shellslash = vim.o.shellslash
+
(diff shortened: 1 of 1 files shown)


andreadev-it:
Found a problem: on linux the shellslash option is not disabled, it doesn't exists at all, and the code breaks. I'm making a change on that regard (using pcall to handle the eventual error), but I'm not sure how to update the PR.

andreadev-it:
Tested on my linux laptop, that issue should be fixed now

cseickel:
Hi @andreadev-it, thanks for submitting this PR. I'm OK with it as is, but I'm going to wait and see what happens on Plenary before merging. If it is accepted there and turned into a utility function, it might be better to just reuse that final Plenary implementation here instead of duplicating it.

cseickel:
Hi @andreadev-it, it looks like the plenary one is stalled and I don't think it is relevant anymore because I have stopped using plenary for reading the filesystem.

Are you still interested in merging this?

I have to say, it seems like there should be more code changes/refactoring to make this work universally. I did a reference search for `is_windows` and came up with a lot of results with inconsistent methods of adjusting paths.

andreadev-it:
Hi @cseickel ,
Sorry for being out of touch these days... 
From what I've experienced, enabling the shellslash option on windows causes way more harm than good, and I personally would not suggest it to anyone.  
I agree that there might be the need to do more changes than these to allow the use of shellslash without generating issues. Actually, From what I've seen when I had shellslash enabled, even changing the code accordingly doesn't always work, since neovim also has a weird way of handling shellslash, IIRC.

Personally, I've stopped using shellslash, but if you think there might be the need to adjust it, I can look it up a little deeper and try to come up with a better fix

cseickel:
No problem @andreadev-it. No one else has asked for shellslash related improvements, so if you are no longer using it then I will close this PR.

Issue #114, opened 2022-02-02
Moving files/folders are inconvenient currently


nyngwang:
For some reason, I have to disable `<Up>`,`<Down>`,`<Left>`,`<Right>` in insert mode. So when I want to rename a file for moving it to another folder, I will have to retyping it's filename:

<img src="https://user-images.githubusercontent.com/24765272/152228253-e80524d6-ea48-4de8-8e40-3be44c6f19fa.png" width="700">

Is it possible to rename the file in the cmdline (using `: ...`) and with the full-path of the file being rename pre-filled? Since I didn't disable the four keys in command mode, this will be convenient for me. (Actually this is the behaviour provided by nvim-tree.lua)

cseickel:
Not right now. If more people were interested I might add the option.

I have to ask, why not fix your problem with using arrow keys in insert mode? That's a pretty basic feature to just live without.

nyngwang:
@cseickel The input method I'm using needs to facilitate these four arrow keys to choose a word from candidates. Inside insert mode, these arrows **move** the insert point which makes typing my mother language in (Neo)Vim impossible. It's OK that you don't want to implement this, but then what I can tell is this indeed an important feature for who need to typing non-english in (Neo)Vim.

nyngwang:
As for other idea: is it possible to simulate `y`(yank) or `d`(delete) and paste the file/folder one needs to move?

cseickel:
If it might affect a large number of people then that makes it a higher priority.

What do mean by simulating yank/delete?

nyngwang:
> If it might affect a large number of people then that makes it a higher priority.

Okay.

> What do mean by simulating yank/delete?

I imagine one can `yy` on the file he/she wants to copy. So just move the cursor to the correct position and then `p`. Same argument for `dd`.

cseickel:
I just pushed this to **main**, add this option to disable the popup inputs:

```lua
      require("neo-tree").setup({
        use_popups_for_input = false, -- If false, inputs will use vim.ui.input() instead of custom floats.
      })
```

nyngwang:
@cseickel: Insane! Really appreciate this option!!

PR #367, opened 2022-06-03
Feat: add option to show none/relative/absolute for move and copy (and fix copy to new directory)

Frydac:
Hi,

This feature is related to https://github.com/nvim-neo-tree/neo-tree.nvim/issues/114

I added similar to `fs_actions.create_node()` (add node) the option to show "none", "relative" or "absolute" path when moving or copying a node/file. I tried to implement it in a similar way to the add feature, using similar naming conventions.

My motivation for adding this option/feature:
I regularly have to move a file from one subtree to another, which sometimes doesn't exist yet. And showing the path to edit can really help a lot, e.g. I want to move
`/path/dir1/aaa/bbb/ccc/file.cpp`
to 
`/path/dir2/aaa/bbb/ccc/file.cpp`
This feature allows for editing only 1 dir name, any other method to accomplish this task is more work atm, I think.

Fix issue with copy:
Also while doing this, I noticed that the copy didn't work when trying to copy to a non-existing directory. It doesn't give an error message but the file wasn't actually copied. 
e.g. to reproduce the issue:
nvim /tmp/test.txt
:Neotree reveal
copy the node with `c` and enter e.g. `dir/test.txt`
-> no error and `/tmp/dir/test.txt` is not created or copied to.
Plenary's path.copy() was used for this feature, I haven't further investigated into plenary what is up with that. I saw that the move uses vim.loop.fs_rename, so I changed the copy implementation with vim.loop.fs_copyfile in a similar manner, which does seem to work now. Removing this usage means I could remove the require of plenary in that file, as it was only used for that path.copy

Maybe these should be 2 separate pull requests, one for the feature one for the fix? Or anything else..

Note: I use this with `use_popups_for_input = false` which works absolutely great, but if floating windows are used, their size doesn't depend on the content but on the title (afaik), the title is still just the basename and not the full path (maybe this would be better.. though in the non-popup the input line could get very long as it also prints the title there). Anyway when the floating window is small, and the path long, it seems not ideal.

### DIFF
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -238,8 +238,8 @@ local config = {
       ["y"] = "copy_to_clipboard",
       ["x"] = "cut_to_clipboard",
       ["p"] = "paste_from_clipboard",
-      ["c"] = "copy", -- takes text input for destination
-      ["m"] = "move", -- takes text input for destination
+      ["c"] = "copy", -- takes text input for destination, also accepts the config.show_path option
+      ["m"] = "move", -- takes text input for destination, also accepts the config.show_path option
       ["q"] = "close_window",
       ["?"] = "show_help",
     },
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -225,9 +225,9 @@ local function get_using_root_directory(state, node)
   -- default to showing only the basename of the path
   local using_root_directory = in_directory
   if state.config.show_path == "absolute" then
-      using_root_directory = ""
+    using_root_directory = ""
   elseif state.config.show_path == "relative" then
-      using_root_directory = state.path
+    using_root_directory = state.path
   end
   return using_root_directory
 end
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -218,15 +218,30 @@ M.paste_from_clipboard = function(state, callback)
   end
 end
 
+---The root_directory is used to decide what part of the filename to show the
+-- user when asking for a new filename to e.g. copy or move to.
+local function get_using_root_directory(state, node)
+  local in_directory, _ = utils.split_path(node:get_id())
+  -- default to showing only the basename of the path
+  local using_root_directory = in_directory
+  if state.config.show_path == "absolute" then
+      using_root_directory = ""
+  elseif state.config.show_path == "relative" then
+      using_root_directory = state.path
+  end
+  return using_root_directory
+end
+
 ---Copies a node to a new location, using typed input.
 ---@param state table The state of the source
 ---@param callback function The callback to call when the command is done. Called with the parent node as the argument.
 M.copy = function(state, callback)
   local node = state.tree:get_node()
+  local using_root_directory = get_using_root_directory(state, node)
   if node.type == "message" then
     return
   end
-  fs_actions.copy_node(node.path, nil, callback)
+  fs_actions.copy_node(node.path, nil, callback, using_root_directory)
 end
 
 ---Moves a node to a new location, using typed input.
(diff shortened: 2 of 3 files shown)


lua/neo-tree/defaults.lua
```diff
@@ -238,8 +238,8 @@ local config = {
       ["y"] = "copy_to_clipboard",
       ["x"] = "cut_to_clipboard",
       ["p"] = "paste_from_clipboard",
-      ["c"] = "copy", -- takes text input for destination
-      ["m"] = "move", -- takes text input for destination
+      ["c"] = "copy", -- takes text input for destination, also accepts the config.show_path option
+      ["m"] = "move", -- takes text input for destination, also accepts the config.show_path option
```
cseickel, line 242: Can you also add this comment to the README?
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/6dac7b5b46c27af921369dc5ea1390025f0e58ae/README.md?plain=1#L212-L213

lua/neo-tree/sources/common/commands.lua
```diff
@@ -218,26 +218,42 @@ M.paste_from_clipboard = function(state, callback)
...
+  end
+  return using_root_directory
+end
+
 ---Copies a node to a new location, using typed input.
 ---@param state table The state of the source
 ---@param callback function The callback to call when the command is done. Called with the parent node as the argument.
 M.copy = function(state, callback)
   local node = state.tree:get_node()
+  local using_root_directory = get_using_root_directory(state, node)
```
cseickel, line 240: Since you created the utility function, can you update `add` and `add_directory` to also use this `using_root_directory` function?

cseickel:
Looks good. Thanks for contributing!

cseickel (approved):

Frydac:
Oops, I made a mistake in the logic for the warning log statement, added it quite late and didn't test again with nil/"none" only with a misspelled word.
Not sure how to practically handle this, I made a commit on my branch, it is only 3 chars difference:
https://github.com/Frydac/neo-tree.nvim/commit/b1fe8c20f8030734fbe3959544e34faf97385c9b

cseickel:
Can you open a new PR?

nyngwang:
Got error with this setup:

```lua
      require('neo-tree').setup {
        -- omitted options
        show_path = 'relative',
```

Frydac:
Thanks for the feedback.

1. The option is modeled after the already existing show_path option documented for the "add" mapping:
     https://github.com/nvim-neo-tree/neo-tree.nvim/blob/main/README.md?plain=1#L199-L205
    The "Add" mapping right below it refers to this, where it is more clear as it is right below it, the copy and move 'documentation' copies the description of the "Add", which indeed only makes sense if you also looked add the "add" mapping and can be more clear.
    So the idea is to set this option for each action you want it, this PR adds it for the "copy" and "move" action and expects the option to be defined similar to:
```
        require("neo-tree").setup({
            window = {
                mappings = {
                    ["r"] = {
                        "move",
                        config = {
                            -- show_path = "absolute"
                        }
                    },
            }
        }
```

2. The error you get is caused by an error in the implementation I mentioned in my previous comment, I will make a PR for that in an hour or so. For now you have to set the option to "absolute" or "relative" to get no error and no warning, if you set it to "none" or any other non-nil value you get a warning, and nil (or not setting the option) results in this error.

nyngwang:
After the change I got this, seems like that UI has been broken (might not related to this PR though):

image

And if I change to `absolute` I got this:

image

I think the UI width settings is probably wrong. @cseickel You might need to take a look.

Frydac:
Indeed, I noted this in my initial post:
> Note: I use this with use_popups_for_input = false which works absolutely great, but if floating windows are used, their size doesn't depend on the content but on the title (afaik), the title is still just the basename and not the full path (maybe this would be better.. though in the non-popup the input line could get very long as it also prints the title there). Anyway when the floating window is small, and the path long, it seems not ideal.

So, if you want to use the relative or absolute path setting the
```      require("neo-tree").setup({
        use_popups_for_input = false, -- If false, inputs will use vim.ui.input() instead of custom floats.
      })
```
helps imo (at least for now).
And what I really like about using the command line/mode to edit the path: you can press `ctrl-f` to open the command-line-window which allows to edit the path in a window where you can switch to normal mode to move around and edit the filename more easily.

PR #374, opened 2022-06-08
fix: don't close last window

smoka7:
when neotree is the only window; `NeoTreeShowToggle` tries to close last window resulting in following error:
```
E5108: Error executing lua .../packer/start/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:92: Vim:E444: Cannot close last window
stack traceback:
        [C]: in function 'nvim_win_close'
        .../packer/start/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:92: in function 'close'
        ...im/site/pack/packer/start/neo-tree.nvim/lua/neo-tree.lua:226: in function 'show'
        [string ":lua"]:1: in main chunk
```

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -88,7 +88,10 @@ M.close = function(state)
           end
           vim.api.nvim_win_set_buf(state.winid, new_buf)
         else
-          vim.api.nvim_win_close(state.winid, true)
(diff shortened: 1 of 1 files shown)


cseickel:
Thanks for the fix, but can you please open a new PR that targets `main` instead? `main` is currently a few commits ahead of `v.2.x` so you will need to rebase, but there shouldn't be any conflicts to worry about.

PR #376, opened 2022-06-09
feat: add <C-w> mapping for inputs, fixes #366

### LINKED ISSUE #366: Pressing <C-w> causes input to switch window
I always use `<C-w>` in insert mode to erase input text. But when I do it in neo-tree when moving files, it just switches windows

[asciicast](https://asciinema.org/a/5gNLURaPax3Yi5S2FAjYQPUWd)

smoka7:


cseickel (approved):
I have never heard of <C-S-w> for deleting a word, and I can't find any documentation around it, but it seems to work.

Thanks!

yujinyuz:
Thanks for working on this @smoka7 

I still experience the issue on the first prompt. After that, the problem goes away.

https://user-images.githubusercontent.com/10972027/173069108-6fda4bc4-dd46-486a-8b99-2187db19f131.mp4

Although the solution looks good by remapping `<C-w>` but I don't seem to understand why `<C-w>` doesn't behave the way it should.

Were you able to identify why `<C-w>` mapping was behaving unexpectedly?

Also if you look closely at the statusline (bottom right) when I press `<C-w>` for the first time, it looks like it's waiting for an input. Not sure if it might help.

smoka7:
I couldn't find the reason that why only `<c-w>`  doesn't work.
I don't have the first prompt issue. in input use `<C-o>:Telescope keymaps` to see if it's mapped in the first time

cseickel:
I can confirm that it works fine for me the first time as well. I also confirmed that it works with the minimal config I use for testing. The issue of it not working the first time is probably something in your personal config @yujinyuz.

I also have no idea why the native <c-w> mapping does not work.

yujinyuz:
@smoka7 @cseickel I found the issue why you weren't experiencing it with minimal config. My `nvim-autopairs` looked something like this

```lua
require('nvim-autopairs').setup {
  disable_filetype = { 'TelescopePrompt', 'vim', 'markdown' },
  map_c_w = true,
  check_ts = true,
}
```

Adding `neo-tree-popup` to the disable_filetype solves the "first prompt" issue

PR #377, opened 2022-06-09
feat: add a custom function for sorting files and directories

smoka7:


lua/neo-tree/sources/common/file-items.lua
```diff
@@ -27,6 +27,8 @@ local function deep_sort(tbl, sort_func)
     local config = require("neo-tree").config
     if config.sort_case_insensitive then
       sort_func = sort_items_case_insensitive
+    elseif config.sort_function ~= nil then
```
cseickel, line 30: I would think that the `config.sort_function` check should be first, so that if both `sort_case_insensitive = true` and `sort_function` is set, `sort_function` will win and be the one that is used.

I would also make the check `type(config.sort_function) == 'function'` instead of just a nil check, in case someone messes up the config. Either that or put that type check in the setup function and if it's the wrong type, set it to nil and log a warning.

cseickel (approved):

PR #406, opened 2022-07-09
feat: add support for external sources, from discussion #403

cseickel:


lua/neo-tree/setup/init.lua
```diff
@@ -408,13 +398,44 @@ M.merge_config = function(user_config, is_auto_config)
...
+  local all_source_names = {}
+  for _, source in ipairs(user_config.sources or default_config.sources) do
+    local parts = utils.split(source, ".")
+    local name = parts[#parts]
+    if #parts > 1 then
+      -- external source, fully qualified module name
+      all_sources[name] = source
+    else
+      -- might be internal source
+      local is_builtin, _ = pcall(require, "neo-tree.sources." .. source)
```
ElPiloto, line 413: What do you think about having external sources with a structure as shown below.  That way there's no need to differentiate between internal and external sources?

`my_external_repo/lua/neo-tree/sources/my_custom_source_name/init.lua`

cseickel: That would work fine the way it's written. It might be clearer if the code called it "internal namespace" instead of "internal source".

What I'm not sure about is if there should be a recommended convention or not. One reason you might want to have a separate namespace is to overwrite a built-in source with your own implementation. It would also prevent potential namespace collisions. Whether this would ever be a real issue in practice, I don't know.

I think for now, I might just reword the comments and variable names to make it clearer, and leave it up to external source authors how they want to organize their namespaces.

ElPiloto: Yeah I suggested that just to simplify the logic here.  But since you've already got it done and this way is more flexible, I agree it makes sense to leave it like this.

MunifTanjim: Instead of depending on the filesytem structure, can we define the sources as tables?

Each source should have a specific structure:

```lua
{
  commands = {},
  components = {},
  setup = function() end,
}
```

and use that source as:

```lua
 require("neo-tree").setup({
        sources = {
            "filesystem",
            "buffers",
            "git_status",
            example_source
        }
}
```

that way, `neo-tree` wouldn't have to worry about `require`-ing stuffs itself.

Also, eventually convert the internal sources to that structure too.

MunifTanjim: Or maybe an option to register sources and then use strings in setup call:

```
require("neo-tree").register_source("example", example_source)

require("neo-tree").setup({
        sources = {
            "filesystem",
            "buffers",
            "git_status",
            "example"
        },
})
```

cseickel: > Instead of depending on the filesytem structure, can we define the sources as tables?

I could see having the source's main module just include references to the correct `commands` and `components` modules. This would add a small amount of flexibility for the source and would not be a meaningful amount of work to implement. 

Does it really change anything though? Wouldn't the author have them in separate files anyhow, and wouldn't it be better if all sources were forced to have a consistent file structure?


> Or maybe an option to register sources and then use strings in setup call

That was my first thought, but it's ultimately not needed. I also think it's better for the end users if we just interop by convention and alleviate the need for them to remember to add those extra `register_source` calls. If anything, I would prefer your first suggestion.

cseickel: > This feels like a completely arbitrary restriction.

You know what, I suppose you are right. It doesn't really matter how the files may be organized.

I'll make it so that the `commands` and `components` module may be keys on your main module. It will fall back to the old require by expected path convention. This new format will eventually become the default and the old require code will be removed.

cseickel: As I was making this change I remembered why I had initially coded it this way. It was to avoid problems with circular requires. It's not necessarily a problem for everyone, but due to that I'll leave it as optional to either follow the file structure or instead use the right keys in the module.

The built-in sources will stick with the current file path conventions.

PR #408, opened 2022-07-10
feat: disable signcolumn by default

danilshvalov:
Hi! I think it would be better if `signcolumn` was disabled by default.

<details>
<summary>Before</summary>

before
</details>

<details>
<summary>After</summary>

after
</details>

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -705,6 +705,7 @@ create_window = function(state)
     },
     win_options = {
         colorcolumn = "",
+        signcolumn = "no",
     }
   }
 
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -703,9 +703,9 @@ create_window = function(state)
       undolevels = -1,
     },
     win_options = {
-        colorcolumn = "",
-        signcolumn = "no",
-    }
+      colorcolumn = "",
+      signcolumn = "no",
+    },
   }
 
   local win


cseickel:
I'm not sure. Either way a person can override whatever the default is, but is it better to explicitly set it to off by default, or do nothing and assume the user's global default is what they expect?

I personally want the sign column on in all of my windows, because I think it looks wrong without some kind of margin.

We get a few people asking how to disable it, but not that many. The question is: 

1. Are there a lot of people that want to change it, but they quietly have figured it out on their own?
2. Or do most people actually like it the way it is?

Another thing, your screen shot has an extra space on the left hand side that I don't have. I wonder why that is? Here is what mine looks like with the sign column on:
image

and with the sign column off:
image

MunifTanjim:
> but is it better to explicitly set it to off by default, or do nothing and assume the user's global default is what they expect?

I think if the signcolumn is not being used, it should be hidden by default. It shouldn't just be used for padding. There is already an option for padding in `indent` component.

cseickel:
> But the default value is `1`.

Well that does change things. I wonder why no one else has ever said anything about the extra padding? I really need to clean up that config of mine, I use it for so many test cases it ends up full of junk...

> I think if the signcolumn is not being used, it should be hidden by default. It shouldn't just be used for padding. There is already an option for padding in indent component.

OK, there's one vote to remove it.

cseickel (approved):

akinsho:
Only commenting on here since I think you were wondering what users in general thought, 
> Are there a lot of people that want to change it, but they quietly have figured it out on their own?

I was very much in this camp, the spacing was duplicated for me, I guess due to the default padding + the signcolumn, so I had it disabled locally. Glad it's now the default

PR #417, opened 2022-07-11
test: use docker

danilshvalov:
Possible testing implementation provided by docker. Refers to #413

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: test
 test:
 	docker build -t neo-tree .
-	docker run --rm -it neo-tree make test-run
+	docker run --rm neo-tree make test-run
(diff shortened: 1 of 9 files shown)


cseickel:
While we no longer need it to deal with the changes in #413, it is nice to have a minimal docker container as an optional tool. I think I'll also add a version for a stable release at some point.

Makefile
```diff
@@ -1,11 +1,16 @@
 .PHONY: test
 test:
+	docker build -t neo-tree .
```
cseickel, line 3: Let's leave the originally `test` action alone and add the docker version as a new one called `test-docker`

cseickel (approved):

PR #420, opened 2022-07-12
docs(README): fix incorrect window picker tag

akinsho:
The tag in the window picker plugin is prefixed with `v` if you don't have it, packer won't install the correct tags

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ use {
       {
         -- only needed if you want to use the commands with "_with_window_picker" suffix
         's1n7ax/nvim-window-picker',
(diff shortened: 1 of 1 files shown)


cseickel (approved):

cseickel:
Thanks for the correction! 

In the future, please target the `main` branch for PRs. In this case I was able to change it for you without any conflicts.

akinsho:
Ah, sorry about that, wasn't really looking at the branches, was just a fly by update, noticed the inconsistency in my config since I copy pasted it 👍🏿

Issue #425, opened 2022-07-13
[Feature request] Show alternate sources as tabs at the top of the tree UI


akinsho:
Firstly, this feature is kind of pure nice to have territory, as a user can easily do what they want without it. I do think though that it will improve discoverability. With that said, onto the actual feature.

I happened upon a screenshot of Intellij's new editor fleet and noticed that it offered separate tabs to view the git tree/source as well as history. It made me think that neo-tree offers multiple sources, but they are only accessible via the command line. You have to reopen neotree with that source specified in order to see it. Maybe there is a more ergonomic way I'm not aware of. Also, you need to remember in the moment that a source exists. I think it would be really nice if there was a visual indicator like a tab maybe using winbar or something to show multiple tabs and a mapping +/- click function to swap between them.

<img width="328" alt="image" src="https://user-images.githubusercontent.com/22454918/178691208-f7a320ec-4779-45be-a104-9f1e6c4fc67c.png">

This is sort of a "this could be cool" not sure how much work it will be on your side but essentially just floating the idea to see what you think.

akinsho:
These things are probably old news to ex GUI editor users, but I've only ever used vim/nvim, so I'm always pretty intrigued by some of the discoverability that GUI editors tend to provide with things like this

cseickel:
I think a Winbar would be a great way to implement this, but that is a very personal feature so maybe we could just provide a function that a user can call within their winbar if they use it.

Something like:

```lua
if vim.bo.filetype == "neo-tree" then
    vim.wo.winbar = "%!v:lua.require'neo-tree'.winbar_source_picker()"
end
```

I don't know if that is the correct syntax, but you get the idea.

cseickel:
Also, if someone happens to use bufferline, and that bufferline provides an offset for the file tree plugin, the bufferline plugin could probably provide the feature as well.

akinsho:
@cseickel I imagined it would be gated behind a config option, so people can enable/disable as they please. Re. providing the function, I think it will be more user-friendly to take a config option as a boolean or a list of sources. Whereas a function puts quite of the onus on a new user to do a lot of wiring up. Ultimately any mechanism that could work would be great, but I've notice a trend towards huge lumps of lua in configs (which I'm also very guilty of in my plugins) I think a function as an option should be a power tool for people looking to go wild

akinsho:
[...]

PR #424, opened 2022-07-13
feat: accept negative padding for `default_component_configs.indent.padding`

pysan3:
With the following config, the outcome still has an extra padding on the left by default as you can see on line `lua/neo-tree` in the picture.

I would like to delete this extra space by setting `default_component_configs.indent.padding` to a negative value to work around this feature.

```lua
require("neo-tree").setup({
  hide_root_node = true,
  default_component_configs = {
    indent = {
      padding = 0,
      -- padding = -2, -- This will work with this PR
    },
  },
})
```

### Current Behavior
(The white line is the edge of the shell window)

- `default_component_configs.indent.padding = 0`
image

- `default_component_configs.indent.padding = -2`
image

### Expected Behavior
- `default_component_configs.indent.padding = -2`
image

Again, thanks for the awesome plugin,
@pysan3

PR #427, opened 2022-07-13
feat: add a clickable source switcher in winbar (related to #425)

pysan3:
Hi, I think this feature (#425) is awesome so I kinda looked into the code to see if this is possible.

Thanks to `nui.nvim` it seems to be quite easy to set `winbar` for Neotree buffers.

This is a very rough and nasty implementation (**not intended to be merged**) and I would like to hear everybody's opinion about this.

#### Default [config here](https://github.com/pysan3/neo-tree.nvim/blob/151f1b5c9d835dd6e36fa54fe27afcb89466ddac/lua/neo-tree/defaults.lua#L41)

neo-tree_winbar_default

### Other Tweaking Params Exposed

```lua
require("neo-tree").setup({
  source_switcher = {
    enable = true,
    tabs_default = { icon = "裡", hl = "NeoTreeTitleBar", hl_active = "NeoTreeModified" },
    tabs = { -- could be table or string (=icon), missing keys fall back to `tabs_default`
      filesystem = { icon = "", text = "Files", hl = "ErrorMsg" },
      -- buffers = "", -- no need to specify all sources (will use default then)
      git_status = "", -- no icon
    },
    -- not implemented but maybe nice to have configs
    justify = "center", -- start, end, center, between, around (idea from bootstrap https://getbootstrap.com/docs/4.0/utilities/flex/#justify-content)
  },
})
```

neo-tree_winbar

Again, this is not production ready and just a proof of concept.
I am looking forward to any feedbacks ;)
@pysan3 

## Future TODOs / Please Help Me I'm not familiar with these stuffs X(

- [ ] Maybe add keybinds to switch
- [ ] Better and specific `hl`s


cseickel:
Wow, I can't believe how quickly you put this together! You're definitely on the right track and I'll merge this when it's done. I'll have some comments later today.

akinsho:
@pysan3 wow this looks great and as was mentioned come together super quickly 👍🏿 . A few things I'd note/suggest, just visually not code, would be to augment the extensibility a bit. I'd allow users to set the separator as well, and I'd also suggest highlight groups being able to distinguish a tab and the background of the winbar separately. I think you could essentially achieve a very similar UI to the screenshot I shared with that. i.e. you could set the entirety of a tab to say black and the background to grey, so tabs if they don't fill the space are visually distinct.

Don't think you need to try and make it follow the screenshot by default, as long as the groundwork allows for that. In the screenshot, you shared the highlight of the "tab" doesn't extend the full width of it, which is likely a minor thing to tweak

akinsho:
Re. keybinds up to @cseickel, but in my mind the feature is (as an obstinate non-mouse user) not very helpful/complete without any key bindings 😅 or the options for some.

pysan3:
@akinsho Just to be clear, please tell me if I'm understanding you correctly.

You mean tab by the color of the text and background as the color behind the text?
Or does it mean like ` <tab> | <tab> . <remaining space as bg>`?

Also, with the current implementation, if you click the `<remaining space>` it will fall back to clicking the right-most tab. Could this feature be confusing maybe?
Actually, to be exact, the area you can click is ` <tab> |` (until the bar). I can easily change this behavior, so what do you think is the most suitable?

@cseickel Thanks!! btw, kinda related to #424 , but currently I am adding `string.rep(" ", padding + 2)` in front of winbar, but this will vary with `hide_root_node` and other settings.

So, could you create a global api that calculates somewhat the _reasonable padding of the root node_? It will be very nice for the UI.

akinsho:
@pysan3 I mean
```vim
%#TabHl# Files %#TabSeparator#|%#TabActiveHl# Git %#TabSeparator#|%#TabBackground#..................
```
The dots indicate the remaining space. i.e. You should be able to set `TabHl` to something like `background = black, foreground = grey` and active tab to `background = black, foreground = red`, and winbar background should be `background = grey`.

Tbh I *personally* think it's weird that clicking space beyond the visual indication of a thing should trigger it i.e. clicking the empty space triggering the last item is weird. Imagine this were your browser and you only two tabs open and a bunch of empty space after the last tab. It would be weird if you then clicked the empty space after the last tab and it opened the last tab IMO.

cseickel:
I would agree with all of @akinsho's points. I would want the default look to be pretty much exactly what the original screenshot showed, with custom `NeoTreeSourceBar*` highlights that might be linked to common tabbar highlights by default.

As far as keybinds, I'm not sure if there needs to be window specific keybinds or of the expectation is that users have global keybinds for each source (as I do). I could see adding optional commands that allow people to bind a key to change to a particular source. I would definitely add `next_source` / `prev_source` commands that would follow the same order as what is visually displayed. I'm not sure if they should be mapped by default and if so, what to map them too.

One last thing I would point out is that I think this could work equally as well in the statusbar, so I might make it a little more generic and call the feature `source_selector` with the choice of showing it in either the statusbar or the winbar being part of the config.

cseickel:
> @cseickel Thanks! btw, kinda related to #424 , but currently I am adding `string.rep(" ", padding + 2)` in front of winbar, but this will vary with `hide_root_node` and other settings.
> 
> So, could you create a global api that calculates somewhat the _reasonable padding of the root node_?

Actually, it's not a given to me that the tab text should always align with the root node of the tree. I would make the padding around each label a config option, then a person can play with that padding if they want it to align with the nodes in the tree.

pysan3:
Thank you all for the feedbacks!

> `%#TabHl# Files %#TabSeparator#|%#TabActiveHl# Git %#TabSeparator#|%#TabBackground#..................`

@akinsho This is very nice. And I totally agree with you about the clicking interface. I will be taking that route.

> custom `NeoTreeSourceBar*` highlights that might be linked to common tabbar highlights by default

My bad, I hardly have any knowledge about the highlightings, so may I hand it to you @cseickel adding the `NeoTreeSourceBar*` names later? I will be sticking with random highlights for now.

> I'm not sure if there needs to be window specific keybinds

Actually, in my usecase, I don't regularly use `buffers` and `git_status` and only have a global mapping to `filesystem`. Yet I do want a way to access these features, so a keybind to switch will be a very handy way.
I was thinking of something like `H: prev_source, L: next_source` as of the vim motion, moving left and right would be nice (but capitalized). (BTW, I usually map them to `:bprev` and `:bnext`)

> make it a little more generic and call the feature `source_selector`

Nice! I didn't notice that but it should be very easy to do. Thanks for pointing out.

> I would make the padding around the labels a config option

Got it b

cseickel:
> My bad, I hardly have any knowledge about the highlightings, so may I hand it to you @cseickel adding the `NeoTreeSourceBar*` names later? I will be sticking with random highlights for now.

That's fine, I don't mind adding this in a follow up commit.



> I was thinking of something like `H: prev_source, L: next_source` as of the vim motion, moving left and right would be nice (but capitalized). (BTW, I usually map them to `:bprev` and `:bnext`)

I also have existing H and L mappings, I'd bet many people do because they are prime real estate on the keyboard. What about `<` and `>` as the defaults? I like reusing editing commands because they are unlikely to be mapped to anything useful already in the context of the neo-tree buffer.

I can add the source switching commands in another PR if you'd like.

pysan3:
> That's fine, I don't mind adding this in a follow up commit.

Thanks

> `<` and `>` as the defaults

I actually like that, let's go with this.

pysan3:
@cseickel , @akinsho and others who came across this thread.

I am not yet finished with the implementation but prior to that, this is the current defaults I am thinking to be configurable.
Can I get your options? It seems to be quite difficult to change the code if I go too far.

```lua
  source_selector = {
    winbar = true, -- toggle to show selector on winbar
    statusbar = true, -- toggle to show selector on statusbar
    tabs_default = {
      icon = "裡",
      icon_place = "left",
      hl = "NeoTreeTitleBar",
      hl_active = "NeoTreeModified",
      content_layout = "center", -- only with `tabs_layout` = "equal", "focus"
      --                start  : |/ 裡 bufname     \/...
      --                end    : |/     裡 bufname \/...
      --                center : |/   裡 bufname   \/...
    },
    hl_backgound = "NeoTreeCursorLine",
    tabs = { -- could be table or string
      filesystem = { icon = "", text = "Files" },
      buffers = "",
      git_status = "",
    },
    tabs_layout = "expand", -- start, end, center, equal, focus
    --             start  : |/  a  \/  b  \/  c  \            |
    --             end    : |            /  a  \/  b  \/  c  \|
    --             center : |      /  a  \/  b  \/  c  \      |
    --             equal  : |/    a    \/    b    \/    c    \|
    --             active : |/  focused tab    \/  b  \/  c  \|
    hide_text_to_fit = false, -- toggle to show only icons in each tab when not enough width
    separator = {
      icon = "|", -- can be string or table, see bellow
      -- icon = { left = "/", right = "\\", override = nil },     -- |/  a  \/  b  \/  c  \...
      -- icon = { left = "/", right = "\\", override = "right" }, -- |/  a  \  b  \  c  \...
      -- icon = { left = "/", right = "\\", override = "left" },  -- |/  a  /  b  /  c  /...
      -- icon = "|",                                              -- |  a  |  b  |  c  |...
      hl = "NeoTreeNormalNC",
    },
    separator_active = nil, -- set separators around the active tab. nil falls back to `source_selector.separator`
  },
```

cseickel:
@pysan3 That's great that you've gone all out on the configuration. This seems to be pretty well thought out.

One comment is that I don't think you need any of the icon configuration unless you plan on supporting different highlight groups for the icon vs text. I think that's probably unnecessary, so you can just specify the tab labels and be done with it. The user can add the icon to that string if they want one, and put it in whatever position they want. I think this also removes the need for a `tabs_default`:


```lua
  source_selector = {
    winbar = true, -- toggle to show selector on winbar
    statusbar = true, -- toggle to show selector on statusbar
    hl = "NeoTreeTitleBar",
    hl_active = "NeoTreeModified",
    hl_backgound = "NeoTreeCursorLine",
    content_layout = "center", -- only with `tabs_layout` = "equal", "focus"
    --                start  : |/ 裡 bufname     \/...
    --                end    : |/     裡 bufname \/...
    --                center : |/   裡 bufname   \/...
    padding = 0, -- can be int or table
    -- padding = { left = 2, right = 0 },
    tab_labels = {
      filesystem = " Files",
      buffers = " Buffers",
      git_status = " Git",
    },
    tabs_layout = "expand", -- start, end, center, equal, focus
    --             start  : |/  a  \/  b  \/  c  \            |
    --             end    : |            /  a  \/  b  \/  c  \|
    --             center : |      /  a  \/  b  \/  c  \      |
    --             equal  : |/    a    \/    b    \/    c    \|
    --             active : |/  focused tab    \/  b  \/  c  \|
    hide_text_to_fit = false, -- toggle to show only icons in each tab when not enough width
    separator = {
      icon = "|", -- can be string or table, see bellow
      -- icon = { left = "/", right = "\\", override = nil },     -- |/  a  \/  b  \/  c  \...
      -- icon = { left = "/", right = "\\", override = "right" }, -- |/  a  \  b  \  c  \...
      -- icon = { left = "/", right = "\\", override = "left" },  -- |/  a  /  b  /  c  /...
      -- icon = "|",                                              -- |  a  |  b  |  c  |...
      hl = "NeoTreeNormalNC",
    },
    separator_active = nil, -- set separators around the active tab. nil falls back to `source_selector.separator`
  },
```


I might also group all the highlights together into one `highlights` table, or make them all a flat list right next to each other with a `highlight_whatever` naming convention.

pysan3:
That's a good point. Do you think its better to change the keynames?

- `hl` -> `hl_tab`
- `hl_active` -> `hl_tab_active`
- `separator.hl` -> `hl_sep`
- `separator_active.hl` -> `hl_sep_active`

cseickel:
> That's a good point. Do you think its better to change the keynames?
> 
> * `hl` -> `hl_tab`
> * `hl_active` -> `hl_tab_active`
> * `separator.hl` -> `hl_sep`
> * `separator_active.hl` -> `hl_sep_active`

Yes, although I would lean toward writing out the word "separator" in full, just for clarity.

pysan3:
Hi all,

I am almost finished with the implementation. Here is the configs and output. And please go to https://github.com/pysan3/neo-tree.nvim/tree/clickable_source_switcher to see the changes.

```lua
  source_selector = {
    winbar = true, -- toggle to show selector on winbar
    statusline = false, -- toggle to show selector on statusline
    tab_labels = { -- falls back to source_name if nil
      filesystem = " Files",
      buffers = " Buffers",
      git_status = " Git",
    },
    content_layout = "center", -- only with `tabs_layout` = "equal", "focus"
    --                start  : |/ 裡 bufname     \/...
    --                end    : |/     裡 bufname \/...
    --                center : |/   裡 bufname   \/...
    tabs_layout = "equal", -- start, end, center, equal, focus
    --             start  : |/  a  \/  b  \/  c  \            |
    --             end    : |            /  a  \/  b  \/  c  \|
    --             center : |      /  a  \/  b  \/  c  \      |
    --             equal  : |/    a    \/    b    \/    c    \|
    --             active : |/  focused tab    \/  b  \/  c  \|
    text_trunc_to_fit = true, -- toggle to truncate text in each tab when not enough width
    padding = 0, -- can be int or table
    -- padding = { left = 2, right = 0 },
    separator = "|", -- can be string or table, see bellow
    -- separator = { left = "/", right = "\\", override = nil },     -- |/  a  \/  b  \/  c  \...
    -- separator = { left = "/", right = "\\", override = "right" }, -- |/  a  \  b  \  c  \...
    -- separator = { left = "/", right = "\\", override = "left" },  -- |/  a  /  b  /  c  /...
    -- separator = { left = "/", right = "\\", override = "active" },-- |/  a  / b:active \  c  \...
    -- separator = "|",                                              -- ||  a  |  b  |  c  |...
    separator_active = nil, -- set separators around the active tab. nil falls back to `source_selector.separator`
    show_separator_on_edge = false,
    --                       true  : |/    a    \/    b    \/    c    \|
    --                       false : |     a    \/    b    \/    c     |
    highlight_tab = "NeoTreeModified",
    highlight_tab_active = "NeoTreeGitDeleted",
    highlight_backgound = "NeoTreeCursorLine",
    highlight_separator = "NeoTreeNormalNC",
    highlight_separator_active = "NeoTreeGitDeleted",
  },
```

### Output with Defaults

neo-tree_winbar

### Output with Truncated Texts

*properly renders even when not enough space
neo-tree_winbar_trunc

## TODOs

- [ ] Better highlightings. Change these lines.
```lua
    highlight_tab = "NeoTreeModified",
    highlight_tab_active = "NeoTreeGitDeleted",
    highlight_backgound = "NeoTreeCursorLine",
    highlight_separator = "NeoTreeNormalNC",
    highlight_separator_active = "NeoTreeGitDeleted",
```
- [ ] Further testings. @cseickel @akinsho Could you test if this works in your environment? (Please set `window.width` and `window.popup.size.width` with fixed integer. See bellow.)
```lua
-- packer
use({
  "pysan3/neo-tree.nvim",
  branch = "clickable_source_switcher",
})
```

## Known Bug

- [ ] @cseickel I need the width of the window to render the `winbar` / `statusline`. There are no problems if user's config sets width in `window.width` and `window.popup.size.width` with int, but it fails with things like `"100%"`. I could not find a way to get the window width in integer before calling `nui.nvim` classes. Is there a way to fix this problem? Please look [here](https://github.com/pysan3/neo-tree.nvim/blob/e95da4d3102a63d8dd7bb258826d24fb18f0f63b/lua/neo-tree/ui/renderer.lua#L713) and [here](https://github.com/pysan3/neo-tree.nvim/blob/e95da4d3102a63d8dd7bb258826d24fb18f0f63b/lua/neo-tree/ui/renderer.lua#L737).

lua/neo-tree/defaults.lua
```diff
@@ -69,7 +69,7 @@ local config = {
     show_separator_on_edge = false,
     --                       true  : |/    a    \/    b    \/    c    \|
     --                       false : |     a    \/    b    \/    c     |
-    highlight_tab = "NeoTreeGitDeleted",
+    highlight_tab = "NeoTreeModified",
     highlight_tab_active = "NeoTreeGitDeleted",
     highlight_backgound = "NeoTreeCursorLine",
```
akinsho, line 74:
```suggestion
    highlight_background = "NeoTreeCursorLine",
```

lua/neo-tree/defaults.lua
```diff
@@ -38,6 +38,43 @@ local config = {
...
+    tabs_layout = "equal", -- start, end, center, equal, focus
+    --             start  : |/  a  \/  b  \/  c  \            |
+    --             end    : |            /  a  \/  b  \/  c  \|
+    --             center : |      /  a  \/  b  \/  c  \      |
+    --             equal  : |/    a    \/    b    \/    c    \|
+    --             active : |/  focused tab    \/  b  \/  c  \|
+    text_trunc_to_fit = true, -- toggle to truncate text in each tab when not enough width
+    padding = 0, -- can be int or table
+    -- padding = { left = 2, right = 0 },
+    separator = "|", -- can be string or table, see bellow
```
akinsho, line 62:
```suggestion
    separator = "│", -- can be string or table, see below
```
I think this is a nicer character to use here as it actually spans the full cell height, but up to you

akinsho:
A couple of notes from testing. I don't seem to be able to set the `background` color. I thought this was due to the typo in the highlight name, but using `highlight_background` and nothing seems to change?

When clicking the tabs, they seem to *replace* my current buffer in the other window rather than replace the neotree window contents? Not sure if this is due to some setting I have re. hijacking or something, but I'd expect regardless of other settings that this would replace the current tree pane, not change my buffer to a new tree window

lua/neo-tree/defaults.lua
pysan3: Thanks! Actually I tried to find that char but couldn't find it

akinsho:
I also think that by default the separators shouldn't have a bright color of any kind they should be a bit like the `WinSeparator` or `NonText` and then a user who wants something shiny can change that but by default they should all be muted. I think really the only thing that should have a clear distinguishing colour is the currently selected tab.

I think if you have separator coloured separators and toned down inactive tabs e.g. using something close to a comment highlight which is usually muted compared to regular text and a contrasting background it will end up looking similar to the screenshot from the original issue

pysan3:
@akinsho Thanks for testing!!

> A couple of notes from testing. I don't seem to be able to set the background color. I thought this was due to the typo in the highlight name, but using highlight_background and nothing seems to change?

My bad, I totally forgot about that setting. Should be fixed with the latest commit.

> When clicking the tabs, they seem to replace my current buffer in the other window rather than replace the neotree window contents

Did you move the focus to a different window, and then click the winbar? I thought that clicking the winbar automatically change the focused window to neotree but maybe not. I will look into that.

> Colors

Yes, I think so too. Currently I am just using _insane_ defaults to make it obvious if something is going wrong, especially for colorings cuz I make a lot of mistakes. Will be fixed when this is stable enough.

akinsho:
That makes sense re. winbar buffer switch issue, you are right, I was clicking the winbar with my focus in my current window. I think that's bound to happen, so it should probably as you say switch to the neotree window first. Maybe use `nvim_win_call` or something.

akinsho:
I'm still unable to get the background highlight working using
```
      highlight_tab = 'NeoTreeTab',
      highlight_separator = 'NeoTreeSeparator',
      highlight_separator_active = 'NeoTreeSeparator',
      highlight_tab_active = 'NeoTreeActiveTab',
      highlight_background = 'NeoTreeTabBackground',
```
The others all work (note these are custom highlights I set up, and I've verified they work), only the background is still not being set.

akinsho:
> That makes sense re. winbar buffer switch issue, you are right, I was clicking the winbar with my focus in my current window. I think that's bound to happen, so it should probably as you say switch to the neotree window first. Maybe use nvim_win_call or something.

Actually scratch that it's not just about my window focus, I find this behaviour to happen sometimes even if I'm inside the neotree buffer.

pysan3:
Is `tabs_layout = "start"` what you want?

cseickel:
> ## Known Bug
> * [ ]  @cseickel I need the width of the window to render the `winbar` / `statusline`. There are no problems if user's config sets width in `window.width` and `window.popup.size.width` with int, but it fails with things like `"100%"`. I could not find a way to get the window width in integer before calling `nui.nvim` classes. Is there a way to fix this problem? Please look [here](https://github.com/pysan3/neo-tree.nvim/blob/e95da4d3102a63d8dd7bb258826d24fb18f0f63b/lua/neo-tree/ui/renderer.lua#L713) and [here](https://github.com/pysan3/neo-tree.nvim/blob/e95da4d3102a63d8dd7bb258826d24fb18f0f63b/lua/neo-tree/ui/renderer.lua#L737).

@pysan3 Actually, I have a better solution for you. Do not set the winbar/statusline to the static text output by your function, just set it to the function itself. Then you can get the actual width within that function and it will update itself automatically. I can give you an example later if you haven't seen that trick before.

akinsho:
Nice @pysan3 that it's so configurable, one thing though is that this removes all padding from the tab which a user can fudge themselves, but I think it would be nice to be able to say each tab is 10 chars or whatever and truncate the tab, and then they all look uniform.

Locally using start and manually adding on space to each tabs content I was able to do

image

Which looks nice, but the irregularity of the tab sizes irks me (since I'm a finicky person 😅) I think it'd be better if they could be uniform but otherwise behave how start behaves.

Also and this one might be a step too far, but I just thought I'd try and underline to improve the delineation for the focused tab and noticed it covers the entire tab which makes sense since it's the highlight for the tab, but there's no way to target just the text with a highlight. I mean, I could possibly hack a ` #highlightgroup#Files# ` in there but, not ideal. In any case, I agree this might be too many things at this point 🤷🏿

pysan3:
> Do not set the winbar/statusline to the static text output by your function, just set it to the function itself. Then you can get the actual width within that function and it will update itself automatically

Shoot, I was too late. I just got it fixed with the latest commit. I could just call the function _after_ the window was created. But your solution seems to be more practical. I will look into that.

> When clicking the tabs, they seem to replace my current buffer in the other window rather than replace the neotree window contents

@akinsho This problem should be fixed with the latest commit. Would you mind checking it?

akinsho:
@pysan3 just tested it and that works well re. winbar clicking 🚀

pysan3:
@akinsho 

> I think it'd be better if they could be uniform but otherwise behave how start behaves.

Nice point. Maybe implementing `tab_width_min`, `tab_width_max` can handle that? If yes, I am going to add that feature.

> there's no way to target just the text with a highlight

With the current implementation, I think the easiest way would be to add `highlight_tab_background` which would define the colors of the spaces between the separator and the source_name. And people may define a highlight **without** the underlines for that.
Do you think this way is feasible?

cseickel:
> Tbh will wait to see what @cseickel says, but maybe the extra highlight it's too much I think with tabline stuff you've got to draw the line otherwise you'll find yourself in a deep rabbit hole, I guess the most intuitive solution to my mind would be having the behaviour as is but nest another hl group which by default mirrors the parent e.g.
> 
> `%#HL_PARENT# %#HL_NAME#Files%#HL_PARENT# #`. It's kind of busy, but works. A lot of this is because nvim's DSL for status/tablines is kind of simplistic when it comes to highlighting, so there are hoops to jump through

Honestly @akinsho, I think it's fair to say you are the expert on creating configurable tabs in a plugin and I would defer to you. I really appreciate your participation in this review. I think the structure of highlight groups you outlined looks good to me.

I also like the idea of having min and max widths for the tabs.

As far as setting the winbar/statusline option as a function call, here is a good example from the [nvim-navic](https://github.com/SmiteshP/nvim-navic) plugin:

```lua
vim.o.statusline = "%{%v:lua.require'nvim-navic'.get_location()%}"
--  OR
vim.o.winbar = "%{%v:lua.require'nvim-navic'.get_location()%}"
```

akinsho:
Thanks for the speedy addition, looks great now

image

With a min tab width of 12 on my setup

Noticed the following when setting the max though 👇🏿 

image

Think maybe since it ran out of space things are lopsided, tbh this is probably a misconfiguration error with the max width behaviour since I didn't fully read how it should work. Guess it is an absolute stop and doesn't do truncation, which is fine. Tbh since I'm able to achieve an ideal layout this is good enough for me.

One thing to note which is a niche edge case is that if you reload the neotree setup which I do whenever any file in my dotfiles is loaded i.e. reloads all my plugins, the winbar is re-applied but to the current window. I think maybe however it is set it should be local to the window e.g. `vim.wo[neotree.window].winhighlight = value` so even on a reload it doesn't try to change other windows.

lua/neo-tree/ui/selector.lua
```diff
@@ -0,0 +1,266 @@
...
+end
+
+M.get_selector_tab_info = function(source_name, source_index, is_active, separator)
+  local config = require("neo-tree").config
+  local separator_config = utils.resolve_config_option(config, "source_selector", nil)
+  if separator_config == nil then
+    log.warn("Cannot find source_selector config. `create_selector` abort.")
+    return {}
+  end
+  local get_strlen = vim.fn.strdisplaywidth
```
akinsho, line 42: the api actually provides an alternative that doesn't require a viml round trip i.e. `api.nvim_strwidth`

pysan3: Thanks!

miversen33:
A completely useless comment, but god that looks good.

pysan3:
TBH, I think the current state is what I can improve with my limited knowledge.

I can't thank you two enough of all your help and many contributions.

One last thing I think I have to do is to brush up the default highlights. Honestly, I really love @akinsho's setup and would be more than happy to set that to the defaults.
Is that ok @cseickel ? And do you mind providing me the details @akinsho ?

Or maybe a better solution would be to create a temporary branch in neo-tree's repository where others can make PRs such as adding keybinds and highlight settings. Which way would be more suitable @cseickel ?

akinsho:
Something like
```lua
local api = vim.api
local normal = api.nvim_get_hl_by_name("NeoTreeNormal", true)
local pmenu = api.nvim_get_hl_by_name("Pmenu", true)
local comment = api.nvim_get_hl_by_name("Comment", true)
api.nvim_set_hl("NeoTreeTabBackground", { link = 'NeoTreeNormal' })
api.nvim_set_hl("NeoTreeTab", { bg = pmenu.background, fg = comment.foreground })
api.nvim_set_hl("NeoTreeSeparator", { link = 'Pmenu' })
api.nvim_set_hl("NeoTreeActiveTab", { bg = normal.background, normal.foreground, bold = true })
```
haven't tested this fully

cseickel:
> One last thing I think I have to do is to brush up the default highlights. Honestly, I really love @akinsho's setup and would be more than happy to set that to the defaults.
> Is that ok @cseickel ?

Absolutely, I think his screenshot looks great and would love for that to be the default.



> Finding good default colours I've found is really hard to predict since people might have transparent backgrounds, they might have completely hand crafted their theme and missed out a bunch of necessary highlights etc.
> 
> What I do in my plugins (and config) is take the normal background colour and darken it enough to create a contrast but that will move this plugin into a rabbit hole I think.

I actually do the same sort of thing in the `highlights` module, where I calculate certain highlight groups as variations of existing groups such as `Normal`. If that's to far down the rabbit hole for you @pysan3 I can put those finishing touches on it.

cseickel:
> Or maybe a better solution would be to create a temporary branch in neo-tree's repository where others can make PRs such as adding keybinds and highlight settings. Which way would be more suitable @cseickel ?

I think this PR/branch is enough for interested users to work with. 

As far as keybinds, I will start off with just `<` and `>` for prev/next source, and provide the "switch to specific source" commands but not bind them by default.

One last thing, once all this is settled we will need instructions in the `help` file.

pysan3:
> If that's to far down the rabbit hole for you I can put those finishing touches on it.

Yes please... I'm the ones who's never touched highlights and just googled `Best ColorSchemes for Neovim` lol

> I think this PR/branch is enough for interested users to work with.

I mean, since this is currently placed in my repo (fork), only I can make changes to the branch (or make PRs), but I was a bit hesitant about my **insane** defaults being merged to the `main` branch, so I would like to have a place to ask others to configure the highlights.

> we will need instructions in the `help` file.

Roger that. Bear with me I haven't got much spare time for a couple of days, but I'll do that ASAP.
Additionally, I will be happy if this feature could be added to the README.

cseickel:
> I mean, since this is currently placed in my repo (fork), only I can make changes to the branch (or make PRs), but I was a bit hesitant about my **insane** defaults being merged to the `main` branch, so I would like to have a place to ask others to configure the highlights.

`main` is for experimentation, and it's fully opt in anyhow, so don't worry about that. I won't release until I feel it is done.



> > we will need instructions in the `help` file.
> 
> Roger that. Bear with me I haven't got much spare time for a couple of days, but I'll do that ASAP.

No problem. I understand that this is volunteer work and probably more than you expected when starting this PR. It's done when it's done. THANK YOU for doing this work!



> Additionally, I will be happy if this feature could be added to the README.

I plan on doing a "soft" release without much fanfare at first, then I will highlight it in the README after I feel it's mature and tested.

pysan3:
Ok, I will tag this PR as `ready for review` and would appreciate if you could merge this.

Again, thanks for all your help. I wouldn't be able to come this far without everybody's help.
I can't wait for this feature to be polished and be added to the mainstream branch ;)))

I'll open another PR when I'm ready with the `help` page.

lua/neo-tree/ui/selector.lua
```diff
@@ -0,0 +1,266 @@
...
+    res = res .. text_with_hl(left_sep, sep_hl)
+  end
+  res = res .. text_with_hl(text, tab_hl)
+  if right_sep ~= nil then
+    res = res .. text_with_hl(right_sep, sep_hl)
+  end
+  return res
+end
+
+M.create_selector = function(state, width)
```
cseickel, line 114: The components should just get the actual window width from vim.api.get_window_width instead of being passed this value, that way it can update dynamically as the window is resized. I can make this change in subsequent commits.

lua/neo-tree/defaults.lua
```diff
@@ -38,6 +38,45 @@ local config = {
   sort_function = nil , -- uses a custom function for sorting files and directories in the tree
   use_popups_for_input = true, -- If false, inputs will use vim.ui.input() instead of custom floats.
   use_default_mappings = true,
+  source_selector = {
+    winbar = true, -- toggle to show selector on winbar
```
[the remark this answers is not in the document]
pysan3: Changed!

lua/neo-tree/defaults.lua
```diff
@@ -38,6 +38,45 @@ local config = {
...
+    --                start  : |/ 裡 bufname     \/...
+    --                end    : |/     裡 bufname \/...
+    --                center : |/   裡 bufname   \/...
+    tabs_layout = "equal", -- start, end, center, equal, focus
+    --             start  : |/  a  \/  b  \/  c  \            |
+    --             end    : |            /  a  \/  b  \/  c  \|
+    --             center : |      /  a  \/  b  \/  c  \      |
+    --             equal  : |/    a    \/    b    \/    c    \|
+    --             active : |/  focused tab    \/  b  \/  c  \|
+    text_trunc_to_fit = true, -- toggle to truncate text in each tab when not enough width
```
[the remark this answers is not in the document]
pysan3: This is one of the easiest things to expose to the user, since I have to only change one line of the code. I thought more configuration is nicer.
A very niche case is that with `:set ambiwidth=double` (which I have never seen even a single person do this setting btw) the truncation sometimes go wrong with multibyte fonts.

image
image

lua/neo-tree/ui/selector.lua
pysan3: I added several functions that calculates the window width with the api.
Commit: https://github.com/pysan3/neo-tree.nvim/commit/56dfe424b02eefb132fff266fb510c39d8671707

Please add the resizing features since it was too complicated for me to understand that part.

One thing I noticed that with `:set laststatus=3` (global statusline), the output might be a bit awkward since it loads with the width of the current window in mind.

- split
image
- floating window
image

pysan3: Besides, I am not intending for `create_selector` to be used outside and just exported for people who might want to change the small details.
Instead I have added several wrapping functions that are meant to be used from outside. Here: 
https://github.com/pysan3/neo-tree.nvim/blob/56dfe424b02eefb132fff266fb510c39d8671707/lua/neo-tree/ui/selector.lua#L217
- `append_source_selector `
- `set_source_selector `
- `auto_set_source_selector `
- `return_source_selector `

lua/neo-tree/defaults.lua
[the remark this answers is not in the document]
akinsho: I have hit this in my bufferline plugin re correctly truncating multibyte characters so you could borrow the function that does that if this would allow you to truncate without needing this option. Only skimmed the comment so this might not be that helpful

lua/neo-tree/ui/selector.lua
cseickel: > Please add the resizing features since it was too complicated for me to understand that part.

You are almost there with `M.return_source_selector`, I'll just need to have it lookup it's own state so it can be called without arguments.

> One thing I noticed that with :set laststatus=3 (global statusline), the output might be a bit awkward since it loads with the width of the current window in mind.

The statusline option was not really meant for global statusline users. I figure each person is either a winbar/global statusline user and they would use the winbar option, or they are a statusline per window user and they will use the statusline option here.

pysan3: > I'll just need to have it lookup it's own state so it can be called without arguments.

I am not sure if we could do that. Currently `create_selector` relies on these elements, which I don't think is resolvable due to the following reasons.

- `state.name`: to see which tab is active, by comparing it with the source_name. For this one, I had no clue how to dodge it.
- `state.winid`: maybe possible to get current winid with lua api and so, but I didn't test it if it could function with `:Neotree action=show`. And also [this issue](https://github.com/nvim-neo-tree/neo-tree.nvim/pull/427#issuecomment-1187239881) ([solution](https://github.com/nvim-neo-tree/neo-tree.nvim/pull/427#issuecomment-1187255975)). At least, it didn't work with a simple `vim.api.nvim_win_call` since the rendering process hops around the current window and passed it to a fallback to nui to come back to the previous focused window? I think.

> I figure each person is either a winbar/global statusline user and they would use the winbar option, or they are a statusline per window user and they will use the statusline option here.

Yes, I think so too.

lua/neo-tree/defaults.lua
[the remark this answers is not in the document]
pysan3: Oh, somehow I didn't notice that you were the author of `bufferline.nvim`! Thanks a lot for the awesome plugin, I heavily depend on it everyday and keep up with your amazing work!
Yes, I do think there is a solution to it, but I didn't have the time and energy to go into this very niche edge-case. (and I got a bit lazy cuz neo-tree itself also has this problem.)

It would be so nice of you if you could fix this problem after this is merged into the `main` branch.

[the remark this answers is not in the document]
akinsho: The function is
```lua

-- truncate a string based on number of display columns/cells it occupies
-- so that multibyte characters are not broken up mid-character
---@param str string
---@param col_limit number
---@return string
local function truncate_by_cell(str, col_limit)
  if str and str:len() == api.nvim_strwidth(str) then return fn.strcharpart(str, 0, col_limit) end
  local short = fn.strcharpart(str, 0, col_limit)
  if api.nvim_strwidth(short) > col_limit then
    while api.nvim_strwidth(short) > col_limit do
      short = fn.strcharpart(short, 0, fn.strchars(short) - 1)
    end
  end
  return short
end
````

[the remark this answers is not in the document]
akinsho: Can definitely be improved haven't looked at this in while tbh but as the name suggests the idea is truncation by cell. There's also since been the addition of `vim.str_utfindex` but really not sure if that can help.

lua/neo-tree/ui/selector.lua
cseickel: > > I'll just need to have it lookup it's own state so it can be called without arguments.
> 
> I am not sure if we could do that

Don't worry, I know how to do that 😉 .

I'm going to merge this and denote it as experimental for now. I'll add in some of the changes I am looking for over the next few weeks before removing the "experimental" note and highlighting it on the README.

If anyone else wants to work on this, just let me know what you are doing so we don't duplicate the work.

cseickel (approved):

PR #436, opened 2022-07-16
fix:render indent_marker for hidden_message

smoka7:


cseickel:
This doesn't actually render an indent marker for hidden messages, it just moves the message to the top of the folder. That doesn't look right to me and it doesn't seem to do what you intended.

smoka7:
Sorry, poor choice of word on my side. Currently with having  
```lua
require('neo-tree').setup({message = {
      { "indent", with_markers = true },
      { "name", highlight = "NeoTreeMessage" },
    }})
```
It renders like this.
2022-07-16T17:45:21,669053742+04:30
I came to two solutions: Moving the message to the top or changing `is_last_child` value and option one seemed easier to me.

cseickel:
That's strange, it doesn't render that way for me at all. Here is how it looks for me in `main`:

image

Do you think it could be and "indent-line" or "indent-blankline" plugin that running in the window and creating that line?

cseickel:
Oh, I see. You must have copied your config here wrong because it should have been within a `renderer` key. When I fixed that I saw the behavior you are talking about. What I see is because the default value is to disable markers for the message.

I guess the tricky part here is to make sure that it renders correctly for people that want to include indent markers on the "hidden items" message, and also render correctly for people that like the default value of not including it in the indent markers.

smoka7:
with  [this](https://github.com/nvim-neo-tree/neo-tree.nvim/pull/436/commits/fa6968d94aa551feeab1dd5e0afcf449eba792ee) commit having same value for both `with_markers` renders 
```lua
require("neo-tree").setup({
	renderers = {
		message = {
			{ "indent", with_markers = true },
			{ "name", highlight = "NeoTreeMessage" },
		},
	},
	default_component_configs = {
		indent = {
			indent_size = 2,
			with_markers = true,
			indent_marker = "│",
			last_indent_marker = "└",
		},
	},
})
```
with_markers 2022-07-17T12:19:23,879170522+04:30
but diffrent values is problematic. Is it necessary two have 2 different config for it?

cseickel:
> but diffrent values is problematic. Is it necessary two have 2 different config for it?

No you don't need to specify that config twice _if you want the default values_. The problem is that the global default is true, and the message default is false. So if you want to disable all markers, that only needs to be specified in the `default_component_config` because it is already off for `messages`. If you want to turn it on for everything, you only need to change the `message` renderer because it is already on for everything else.

The current code is still putting the messages at the top for me, which is undesirable. I think that it belongs at the bottom.

Just to be explicit, there are three situations to cover cleanly:

- [ ] User has no indent markers
- [ ] User has indent markers for all but the message (the default)
- [ ] User enables indent markers for all **including** the `(hidden items)` message

smoka7:
I moved the message to the bottom.
for now just this isn't covered
> * [ ]  User has indent markers for all but the message (the default)

Is there a way that i can get the `default_component_configs.indent.with_markers` value in [here](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/471e11f9bd906a2f87081944535389c697110d23/lua/neo-tree/sources/common/components.lua#L323,L324)?

cseickel:
> Is there a way that i can get the `default_component_configs.indent.with_markers` value in [here](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/471e11f9bd906a2f87081944535389c697110d23/lua/neo-tree/sources/common/components.lua#L323,L324)?

Yes, it is part of the `state` object: `state.renderers.message[0]` will give you the indent component for messages. That makes a lot of assumptions though. There is no guarantee that there is a `message` renderer or that the first item is an indent component, but you get the idea.

smoka7:
No, they're not same. `default_component_configs.indent.with_markers` gets merged with renderes in init. 
I can't find a way to handle the default state.
Thanks for putting the time.

cseickel:
> No, they're not same. `default_component_configs.indent.with_markers` gets merged with renderes in init.

That's exactly what you need, I don't see why you would want to know what the intermediate configs are, it's the end result that matters.

> I can't find a way to handle the default state.
Thanks for putting the time.

I see what needs to be done and I can take care of it later if you've stopped working on this.

Issue #439, opened 2022-07-17
Is a PR for notification / dialog options welcome?


tobealive:
I'm glad I found your plugin. Thanks so much for all the work. It's really next level 🚀

_No. 1:_  I would like to make notifications configurable. Currently, I just have this pretty dirty modified on my local machine and maybe redoing it clean is worth a PR. I thought of **render style**, **timeout length** and **notification position**.

Render style and timeout should be pretty simple and straightforward as these are just args we can pass when calling notify.
To set the position I would write a small notification lib which allows choosing the notification position.

Where does this idea come from? Basically from using a tree mainly on the right side and impairments with that. (Why right? It won't shift the text when opening the tree. Also having a lot of photoshop hours on my back from my early days as a media designer I really got accustomed to having my document structure and layers on the right side, it feels more out of the way). A "minimal" rendered notification would not overlap that much of the file tree. A notification positioned on the bottom would stay even more out of the way.

---

_No 2.:_  Another thing when the tree is positioned on the right is anchoring of dialogs. E.g., when triggering renaming or adding a file. Currently, **parts of the dialogs are outside the ui** when not using an overly wide tree. This should be a simple fix to implement. Probably just changing window anchors if the tree is opened on the right. Also, it is probably more crucial as the first one.

If such options are welcome and someone wants to take one of these please feel free to join on this contrib.

cseickel:
> I'm glad I found your plugin. Thanks so much for all the work. It's really next level 🚀

Thanks!

> What I would like to make notifications configurable. Currently, I just have this pretty dirty modified on my local machine and maybe redoing it clean is worth a PR. I thought of render style, timeout length and notification position.
>
> Render style and timeout should be pretty simple and straightforward as these are just args we can pass when calling notify.
To set the position I would write a small notification lib which allows choosing the notification position.

I remember thinking that I will need to add some config options for this eventually when I add `vim.notify` support. I'd be happy to merge a PR for this. I would just say that for any new options, always make sure it is opt-in and does not affect users that are not interested in this feature. Also, it's probably better to discuss the proposed config structure first in a draft PR before writing the actual code.

-----

> Another thing when the tree is positioned on the right is anchoring of dialogs. E.g., when triggering renaming or adding a file. Currently, parts of the dialogs are outside the ui when not using an overly wide tree. 

That's true that I never use it on the right hand side so I'm sure there are some oversights there. It would be great to have a right side user submit some quality of life fixes for that situation.

[...]

PR #440, opened 2022-07-17
fix: dialogs outside ui when tree pos is right

tobealive:
Fixes No. 2 of #439 

There are always chances for improvement. When you see one here, I would love if you let me know.

### DIFF
--- a/lua/neo-tree/ui/popups.lua
+++ b/lua/neo-tree/ui/popups.lua
@@ -20,11 +20,10 @@ M.popup_options = function(title, min_width, override_options)
   local popup_border_style = nt.config.popup_border_style
   local popup_border_text = NuiText(" " .. title .. " ", highlights.FLOAT_TITLE)
   local col = 0
-  local offset = (vim.api.nvim_win_get_width(0) - width) - 2
-  local is_right_window = vim.api.nvim_win_get_position(0)[2] + vim.api.nvim_win_get_width(0)
-      == vim.o.columns
-  if is_right_window and offset < 0 then
-    col = offset
+  -- fix popup position when using multigrid
+  local popup_last_col = vim.api.nvim_win_get_position(0)[2] + width + 2
+  if popup_last_col >= vim.o.columns then
+    col = vim.o.columns - popup_last_col
   end
   local popup_options = {
     relative = "cursor",
--- a/lua/neo-tree/ui/popups.lua
+++ b/lua/neo-tree/ui/popups.lua
@@ -20,9 +20,8 @@ M.popup_options = function(title, min_width, override_options)
   local popup_border_style = nt.config.popup_border_style
   local popup_border_text = NuiText(" " .. title .. " ", highlights.FLOAT_TITLE)
   local col = 0
-  local tree_pos = vim.api.nvim_buf_get_var(0, "neo_tree_position")
   local offset = (vim.api.nvim_win_get_width(0) - width) - 2
-  if tree_pos == "right" and offset < 0 then
+  if vim.api.nvim_win_get_number(0) ~= 1 and offset < 0 then
     col = offset
   end
   local popup_options = {
--- a/lua/neo-tree/ui/popups.lua
+++ b/lua/neo-tree/ui/popups.lua
@@ -21,7 +21,9 @@ M.popup_options = function(title, min_width, override_options)
   local popup_border_text = NuiText(" " .. title .. " ", highlights.FLOAT_TITLE)
   local col = 0
   local offset = (vim.api.nvim_win_get_width(0) - width) - 2
-  if vim.api.nvim_win_get_number(0) ~= 1 and offset < 0 then
+  local is_right_window = vim.api.nvim_win_get_position(0)[2] + vim.api.nvim_win_get_width(0)
+      == vim.o.columns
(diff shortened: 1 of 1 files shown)


tobealive:
Uff, those are some heavy little errors. My bad that I didn't test with floating windows. Nice to have those tests.

We need to get the position in another way then `vim.api.nvim_buf_get_var(0, "neo_tree_position")`.
I tried it with  `state` prior to this approach, but it was unsuccessful. Probably we should paddle back to state.

lua/neo-tree/ui/popups.lua
```diff
@@ -19,11 +19,16 @@ M.popup_options = function(title, min_width, override_options)
   local nt = require("neo-tree")
   local popup_border_style = nt.config.popup_border_style
   local popup_border_text = NuiText(" " .. title .. " ", highlights.FLOAT_TITLE)
+  local col = 0
+  local offset = (vim.api.nvim_win_get_width(0) - width) - 2
+  if vim.api.nvim_win_get_number(0) ~= 1 and offset < 0 then
```
cseickel, line 24: I don't think this is always going to work out the way you want it to. A window could still be on the left and not be window number 1.

cseickel:
You know, I just tested it and I found that nvim already avoids having a floating window go off the screen:

image

AM I misunderstanding what you are trying to do, or does that built-in feature just not work in all situations?

lua/neo-tree/ui/popups.lua
tobealive: I've tried some quick test to force this and check the docs `(windows are numbered from top-left to bottom-right)`. So I guessed it could be a save way.

Do you have a suggestion how to solve this? I mistakenly though pcalling the position would fix the error with floating windows so I removed it from the comment above.

tobealive:
> You know, I just tested it and I found that nvim already avoids having a floating window go off the screen:
> 
> image
> 
> AM I misunderstanding what you are trying to do, or does that built-in feature just not work in all situations?

You are so right. I got so accustomed to neovide in the last days that I didn't opend a proper CLI. It's actually related to the client I was using. And something that should be fixed on the clients side.
Screenshot_20220717_223401

<details>
<summary>Using the fix</summary>

Screenshot_20220717_223726
</details>

cseickel:
> edit: out of curiosity i checked on other clients using multigrid and they are suffering from the same issue. So if you want to implement support for those, maybe this is still a thing to consider 👍.

Certainly, if it's actually an issue for multiple clients then lets go ahead and add it. I just thought I might have been misunderstanding something when I couldn't recreate the issue. I'll review the code now.

lua/neo-tree/ui/popups.lua
cseickel: If you look at this screenshot, you'll see that Neo-tree is actually window 2 even though it is left aligned.
image

I think the correct solution is to just calculate the starting column for the window as an absolute position within the neovim editor, and see if the width of the popup will end up beyond the max columns (`vim.o.columns`). If so, then move the position of the popup whatever amount is needed to keep it within the editor window. This way it will always work regardless of the position of the parent window.

tobealive:
Okay great. I've seen top and bottom position were introduced. Probably those positions need to be excluded as well.
The worst thing that can happen currently as far as I can see, is to lose the left border of the popup dialog as the text input is moved over it to the left.

cseickel:
> In case of doing an opt-in on the right window instead of an opt-out of the left - still using numbers though - something like this came to my mind.
> 
> ```lua
> local windows = vim.api.nvim_list_wins()
> local window_nr = vim.api.nvim_win_get_number
> local is_most_right_window = window_nr(0) == window_nr(windows[#windows])
> ```
> 
> But the first one is slimmer and as far as I can tell does the job just well.

Sorry, but I'm going to be stubborn on this one. I won't accept any solution that depends on using either the window number or the "position" that the window was originally opened with. Both methods are fragile and will be hot spots for bugs.

I think that the correct way to do this is to get the coordinates of the window within the editor and figure out if the starting column of the popup + width of the popup will be greater than `vim.o.columns`. If it is, adjust accordingly. This will work universally for any window configuration.

tobealive:
> Sorry, but I'm going to be stubborn on this one. I won't accept any solution that depends on using either the window number or the "position" that the window was originally opened with. Both methods are fragile and will be hot spots for bugs.

Hey no problem.

But you know that window numbers swap dynamically when windows are moved.
The IDs of course are kept but the window, which is closet to the top left will always be Nr. 1.

cseickel:
> But that doesn't sound like you consider that window numbers swap dynamically when windows are moved.The IDs of course are kept but the window which is closet to the top left will always be Nr. 1.

I am very familiar with the window number and id system. My point is that for every assumption you can make about where a window is based on it's number, I can find a window layout where that is not true. If there is a solution based off of window numbers that is actually correct for all cases, it would be 10x the amount of code than just doing it the way I suggested, which is actually simple and to the point.

cseickel (approved):
Now that looks perfect. I am assuming that a negative `col` value works here. I can't fully test it because I don't use a client that would have this problem.

tobealive:
Yep negative col works here :+1:  😊

:shipit:

PR #449, opened 2022-07-21
source selector refinement

cseickel:
This is related to #427 #425 

I'll use this PR for further refinement of the source selector feature. Here is the current default look:

image

Next up is to:

- [x] make sure that it resizes along with the window resize
- [x] fix glitches with clicking when there are multiple trees open.
- [x] add "context", showing the current parent node when it scrolls off visible window

EDIT: These two will be handled separately. I think I'd like to refactor, move more config into the sources, and allow for other functionality besides just source selectors in the winbar.
- [ ] handle external sources
- [ ] add color to icons?

### DIFF
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -48,6 +48,7 @@ local config = {
   source_selector = {
     winbar = false, -- toggle to show selector on winbar
     statusline = false, -- toggle to show selector on statusline
+    show_scrolled_off_parent_node = true,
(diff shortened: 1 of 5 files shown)


cseickel:
Winbar with `show_scrolled_off_node_parent_node = true` after scrolling down:
image

This is to provide context and replaces the selector buttons after scrolling down. The next step is to rename things to reflect that this functionality is not purely about source selectors anymore.

Also, the button labels belong in the source configs. This will be more appropriate for external sources.

PR #468, opened 2022-08-04
Hide by pattern path

johnf:
Add support to hide files by comparing to glob of full path

### DIFF
--- a/lua/neo-tree/sources/common/file-items.lua
+++ b/lua/neo-tree/sources/common/file-items.lua
@@ -123,7 +123,9 @@ function create_item(context, path, _type)
     end
     if f.hide_by_pattern then
       for _, p in ipairs(f.hide_by_pattern) do
-        if string.find(name, p) then
+        local separator_pattern = utils.is_windows and '\\' or '/'
+        local match = string.find(p, separator_pattern) and path or name
+        if string.find(match, p) then
           item.filtered_by = item.filtered_by or {}
           item.filtered_by.pattern = true
         end
--- a/lua/neo-tree/sources/common/file-items.lua
+++ b/lua/neo-tree/sources/common/file-items.lua
@@ -123,7 +123,7 @@ function create_item(context, path, _type)
     end
     if f.hide_by_pattern then
       for _, p in ipairs(f.hide_by_pattern) do
-        local separator_pattern = utils.is_windows and '\\' or '/'
+        local separator_pattern = utils.is_windows and "\\" or "/"
         local match = string.find(p, separator_pattern) and path or name
         if string.find(match, p) then
           item.filtered_by = item.filtered_by or {}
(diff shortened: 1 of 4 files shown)


lua/neo-tree/sources/common/file-items.lua
```diff
@@ -119,7 +119,8 @@ function create_item(context, path, _type)
     end
     if f.hide_by_pattern then
       for _, p in ipairs(f.hide_by_pattern) do
-        if string.find(name, p) then
+        local match = string.find(p, '%/') and path or name
```
cseickel, line 122: The only thing I would change is to use `utils.path_separator` so that this supports windows as well.

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/a922c3aff32f3a9498466dd1aa99e14df360fbc5/lua/neo-tree/utils.lua#L513

cseickel (changes requested):

lua/neo-tree/sources/common/file-items.lua
johnf: I'll need to spend some more time on this over the weekend.
My quick look the glob function is swallowing up the separators if you pass 'C:\path\file.txt' so will need to work out how best to solve for that

lua/neo-tree/sources/common/file-items.lua
```diff
@@ -119,7 +119,8 @@ function create_item(context, path, _type)
     end
     if f.hide_by_pattern then
       for _, p in ipairs(f.hide_by_pattern) do
-        if string.find(name, p) then
+        local match = string.find(p, '%' .. utils.path_separator) and path or name
```
cseickel, line 122: I think the Windows issue can be solved with this:

```suggestion
        local separator_pattern = utils.is_windows and '\\' or '/'
        local match = string.find(p, separator_pattern) and path or name
```

johnf:
@cseickel I've made the changes as suggested.
This works fine on Unix systems
In theory it will work fine on Windows systems but I'm not sure the glob library actually works on windows systems
When I was printing out the glob patterns it looks like it eats up the backslashed

cseickel:
> In theory it will work fine on Windows systems but I'm not sure the glob library actually works on windows systems
> When I was printing out the glob patterns it looks like it eats up the backslashed

I think it should work if they use double backslash to escape it. Windows users are used to that.

lua/neo-tree/sources/common/file-items.lua
```diff
@@ -123,7 +123,9 @@ function create_item(context, path, _type)
     end
     if f.hide_by_pattern then
       for _, p in ipairs(f.hide_by_pattern) do
-        if string.find(name, p) then
+        local separator_pattern = utils.is_windows and '\\' or '/'
```
cseickel, line 126: There's just one silly stylua check issue:
```suggestion
        local separator_pattern = utils.is_windows and "\\" or "/"
```

cseickel (approved):
This looks good, thanks!

PR #473, opened 2022-08-08
feat: added basic UNIX trash support

### LINKED ISSUE #202: Support trash instead of delete
Thanks for the awesome plugin!

Adding support for `trash` instead of `rm` would be awesome!
Nvim-Tree does this with the `D` shortcut.

RaafatTurki:
This closes #202, I've intentionally kept it separate from and similar to the normal deletion operation for now.

- [ ] visual mode support
- [ ] documentation
- [ ] windows support
- [ ] configurable trash commands (per OS)

Questions:
- I've remapped `fuzzy_finder_directory` from `D` to `F` and mapped `trash` to `D`, I'll revert that back later to avoid breaking changes however what do you suggest `trash` to be mapped to?
- I've defined `trash` and `trash_visual` yet `D` doesn't work in visual mode, am I missing something?
- I'm not that versed in windows commands and don't have a windows machine rn, I understand `Recycle <path>` is the trash command there but does it work on `folders` as well?


lua/neo-tree/defaults.lua
```diff
@@ -293,6 +293,7 @@ local config = {
       },
       ["A"] = "add_directory", -- also accepts the config.show_path option.
       ["d"] = "delete",
+      ["D"] = "trash",
```
cseickel, line 296: I would just not map this by default.

lua/neo-tree/sources/common/commands.lua
```diff
@@ -342,6 +342,25 @@ M.delete_visual = function(state, selected_nodes, callback)
...
+  local node = tree:get_node()
+  if node.type == "message" then
+    return
+  end
+  fs_actions.trash_node(node.path, callback)
+end
+
+M.trash_visual = function(state, selected_nodes, callback)
+  local paths_to_delete = {}
+  for _, node_to_delete in pairs(selected_nodes) do
```
cseickel, line 356: I think the only problem here is that it should be **i**pairs, it is a list and not a key/value map.

```suggestion
  for _, node_to_delete in ipairs(selected_nodes) do
```

cseickel (commented):
Hi @RaafatTurki, thanks for working on this. To answer your questions:

>I've remapped fuzzy_finder_directory from D to F and mapped trash to D, I'll revert that back later to avoid breaking changes however what do you suggest trash to be mapped to?

Not everything is mapped by default. I would make this an available function (or config argument) that users can map themselves if they want to use it.

> I've defined trash and trash_visual yet D doesn't work in visual mode, am I missing something?

See `ipairs` comment.

> I'm not that versed in windows commands and don't have a windows machine rn, I understand Recycle <path> is the trash command there but does it work on folders as well?

I don't know, I can check tomorrow.

> A lot of logic is shared with delete and delete_visual, It would make more sense to merge them and add a use_trash arg, How does this sound?

That sounds like a great idea.

lua/neo-tree/sources/filesystem/lib/fs_actions.lua
```diff
@@ -482,6 +482,113 @@ M.delete_nodes = function(paths_to_delete, callback)
...
+      return
+    end
+
+    if _type == "directory" then
+      local success = false
+      if utils.is_windows then
+        -- TODO: implement trashing for windows
+        log.warn("Trashing is not supported on windows")
+      else
+        local result = vim.fn.system({ "gio", "trash", path })
```
MunifTanjim, line 536: Trash command should be configurable. `gio` is probably only available on systems with GTK. MacOS has `trash` command. Windows probably has someting.

I would like to use this wrapper script with `neo-tree`: https://github.com/MunifTanjim/dotfiles/blob/f5c1ae9be297199a1188df9218ba2f921bf4af26/private_dot_config/lf/executable_trash

RaafatTurki: Actually `gio` is available on every system with `glib2` which is almost every UNIX system.

IIRC MacOS doesn't come with `trash` and needs to be installed.

Yes windows does come with `Recycle`

If you want to use an external trashing utility (such as your script) you could put it somewhere on your `$PATH` and pass its name into the trash config this PR would provide.

cseickel: I wonder, does there need to be a separation between "delete" and "trash"? Maybe it can be a single `delete` command that has a config option for what command is executed. Lets call that config option `command`. If that command is `nil`, we stick with the current OS specific logic. If it is a list, we append the path to that list and send it to `vim.fn.system` as-is.

RaafatTurki: An interesting approach.
Speaking of which how do I read configs?
Could you point a place in the codebase that's doing something based on configs.

cseickel: Check out the `add` command:

```lua
      ["a"] = {
        "add",
        -- some commands may take optional config options, see `:h neo-tree-mappings` for details
        config = {
          show_path = "none" -- "none", "relative", "absolute"
        }
      },
```
That config object will be added to the `state` object that is passed to the command function, you can see it being used here:
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/a922c3aff32f3a9498466dd1aa99e14df360fbc5/lua/neo-tree/sources/common/commands.lua#L31

Issue #393, opened 2022-06-28
RAM usage increases on each save if the window is open


shaksiper:
I have been noticing this weird thing that each save action increases the RAM usage of nvim process.
I did some manual trouble shooting (I had to because there doesn't seem to be trivial way to monitor RAM usage within vim/neovim), and the result seems to be the neo-tree.

Steps to reproduce:
start neovim in the directory `nvim .`
open couple of files.
`:w` 10-20 times.

nvim starts with around **50MB RAM** usage, then after this process it becomes **100MB**, then stays that way. Memory is only deallocated when the neo-tree is closed (although I didn't thoroughly try this, if it always frees the memory when the window/panel is closed). It only increases without limit until the nvim prcess is killed by the OS.

Same behavior with regardless of (use_libuv_file_watcher = true) option.

Also I am using Linux with btrfs. I don't have a chance to try it on another file system.

cseickel:
I usually start with eliminating diagnostics and git updates with any memory or performance issues. You can disable those with these settings:

```lua
require("neo-tree").setup({
  enable_diagnostics = false,
  enable_git_status = false,
})
```

In addition to testing without those turned on, can you also tell me:

- What version of neovim?
- What version of neo-tree?
- What files are you working on?
- Have you tried repeating the steps in a different project?

shaksiper:
I will do that the first chance I get and report back

kamilez:
@cseickel I am facing the same problem and see that @shaksiper didn't respond yet so it would be nice to have it fixed since in large projects it seems to grow much faster than in directory with one file. My versions are:
- NVIM v0.8.0-dev-nightly-5-g1f1863ed5,
- latest (2.32),
- mostly go but I think it doesn't matter - I've tried in my project but also in empty directory and only one file within - it seems to take more saves in second case but still with ~100 writes it grows at least 2x,

miversen33:
Be sure to investigate your LSPs as well. I've noticed sumneko is a huge memory hog/leak, but it took me a while to establish that

cseickel:
@kamilez Did you try disabling git and diagnostics? They are the first things I look at for any performance or memory issues:

```lua
require("neo-tree").setup({
  enable_diagnostics = false,
  enable_git_status = false,
})
```

The hard part here is finding the particular thing that causes the memory leak. It doesn't happen for me so I can't do much about it until I find a way to recreate the issue.

kamilez:
[...]

PR #475, opened 2022-08-08
refactor: update fs_scan to avoid memory leaks, part of #393

cseickel:


cseickel:
Note: The `nvim_buf_clear_namespace()` call that was in an earlier version is being moved upstream to Nui in https://github.com/MunifTanjim/nui.nvim/pull/176.

Issue #134, opened 2022-02-08
File Preview


bennypowers:
I would love to have file preview while the cursor is on a file, in both the sidebar and float windows

nyngwang:
I would still recommend [fzf-lua](https://github.com/ibhagwan/fzf-lua) first. But it could be a good idea to implement it inside neo-tree itself.

bennypowers:
As an initial straw proposal, I suggest that placing the cursor on a file should open a new buffer which previews the contents of that file. if a buffer for that file is already open, switch to it. when the cursor moves, if that file didn't already have a buffer, close it

That would more-or-less map with atom/v****e behaviour

nyngwang:
> Telescope also has fuzzy file preview
> 
> but fuzzy finding a file by name and browsing a file tree are _two separate cognitive tasks_.

You can set the `cwd` option in fzf-lua to preview files under the current folders and/or any folders with respect to `cwd`. But yeah, people are telescoping nowadays. I just personally hate it :) It's too heavy and energy-consuming.

nyngwang:
Just previewing files without the ability to **filtering** them, seems to be just a fancy eye-catching function. I used to try many, all uninstalled.

bennypowers:
@nyngwang I'm really glad you found a tool that you enjoy for fuzzy finding. I'm also equally happy for you that you've discovered you don't really need a file browser.

This feature request has *nothing to do* with fuzzy finding, and this repo is primarily a file browser, so while I appreciate that you want me to use your fuzzy finding tool instead of this package, perhaps we could continue that discussion in a *separate forum*.

cseickel:
> As an initial straw proposal, I suggest that placing the cursor on a file should open a new buffer which previews the contents of that file. if a buffer for that file is already open, switch to it. when the cursor moves, if that file didn't already have a buffer, close it
> 
> That would more-or-less map with atom/v****e behaviour

I didn't realize vscode has this feature. Can you provide a link to a demo or documentation page?

bennypowers:
https://github.dev/nvim-neo-tree/neo-tree.nvim

single-click (NOT double-click) any file in the file explorer

bennypowers:
I'd think cursor-over maps to single click and <enter> maps to double click

Doesn't need to be exactly like the M$ editor, just brought it as an example of a common ui. A floating window would likely also be ok, but the advantage to a full 'temp' buffer is that user can see more columns in her buffer when previewing

[...]

PR #482, opened 2022-08-13
feat: add preview command and position handling

mrbjarksen:
This PR adds commands to preview nodes which correspond to files or file positions. It also allows the standard open commands to open files at positions given by position nodes.

A position node is defined as a node with the field `extra.position`, which is an array-like table of the form `{ <line>, <column> }`, where the line and column numbers are 0-based (it might be more sensible to change this to (1, 0)-based indexing). If the field `extra.end_position` is present, previewing the node will highlight the range between `position` and `end_positions`.

The code is adapted from mrbjarksen/neo-tree-diagnostics.nvim#1. It is likely that this feature will need some tweaking and testing before it can be merged. I have added the branch [native-open-and-preview](https://github.com/mrbjarksen/neo-tree-diagnostics.nvim/tree/native-open-and-preview) to [neo-tree-diagnostics.nvim](https://github.com/mrbjarksen/neo-tree-diagnostics.nvim) to test this feature using position nodes.


cseickel:
I played around with it and I found that I intuitively wanted it to work a different way. I am picturing a "preview mode", where once activated, it previews whatever my cursor is focused on and follows that focus, until the "preview mode" is toggle off. 

When the preview goes off, I'm not sure if it should stick on the last file or go back to the file that was open beforehand.

I would also prefer if there were a single mapping that toggled it off and on.

I think there also needs to be some visual indication that preview mode is enabled.

One way to accomplish the above is if it previews in a floating window where the title bar has some text to indicate that it is a preview window. Then when it is closed it is clear that it should just disappear and you will be left with the original windows underneath it. Then hitting enter, or any other command that normally opens a file will close that preview window and open the file as normal.

What do you think?

mrbjarksen:
I agree that it would be great to have the preview follow the focused node, and that is how I envisioned it at first. At the time, I had some issues with `CursorMoved` events (switching windows caused errors). It might be time to revisit this. The suggestions in #134 to use `vim.loop` sound promising, but I'll admit the API has gone over my head when I've tried to learn to use it.

In my mind, the preview should ideally be a sort of answer to '_What happens if I open this?_', showing exactly which window will be used and the view within said window if you choose to press enter, while changing nothing about the state of the user space (e.g. listed buffers, cursor positions, etc.) while it is active. Note that the preview commands in their current state do not entirely accomplish this.

Using a floating window might be more sensible, being less complex and only losing the ability to see which window will be used. I'm also not sure how to show visual indication using the method above. Another option, although less ideal, is using the built-in preview window.

I'd imagine the behavior of a preview is very opinionated and it may be worth implementing multiple options to choose from. I don't know if that is worth the extra maintenance though.

I'll look into making the preview follow the focus for now.

cseickel:
> I agree that it would be great to have the preview follow the focused node, and that is how I envisioned it at first. At the time, I had some issues with `CursorMoved` events (switching windows caused errors).

Ideally showing the preview should not have to focus that preview window, so the focus should stay in the Neo-tree window. AT least that's how I picture it.

> I'd imagine the behavior of a preview is very opinionated and it may be worth implementing multiple options to choose from. I don't know if that is worth the extra maintenance though.

I think in this case it is worthwhile because I'd imagine the preference will be pretty evenly split. If you want, I can add the floating window code (my preference) while you focus on the current "where it would normally open" code.

> I'm also not sure how to show visual indication using the method above.

I was thinking that we could just add a window local variable that acts as a flag, then it could be added to someone's customized statusline or winbar. Something like 'vim.w.neo_tree_preview = 1'. Not everyone digs that deep into customizing their statusline/winbar, but for those of us that do then this would be enough.

mrbjarksen:
> Ideally showing the preview should not have to focus that preview window, so the focus should stay in the Neo-tree window. AT least that's how I picture it.

I agree. What I meant was that manually switching windows while the preview was active caused issues. Sorry if I wasn't clear.

> I think in this case it is worthwhile because I'd imagine the preference will be pretty evenly split. If you want, I can add the floating window code (my preference) while you focus on the current "where it would normally open" code.

That sounds good. It shouldn't be too much trouble to hook into my existing preview module, opening a new floating window instead of using `findWindow`. I'll leave it to you if you want to go that route or implement it separately.

> I was thinking that we could just add a window local variable that acts as a flag, then it could be added to someone's customized statusline or winbar. Something like 'vim.w.neo_tree_preview = 1'. Not everyone digs that deep into customizing their statusline/winbar, but for those of us that do then this would be enough.

That's a good idea, I hadn't thought of that. It's probably the best way to go.

mrbjarksen:
I just added a `toggle_preview` command. It was simpler than I thought and I'm not having the same issues I had before. I've only tested it a small amount though. Let me know if you find anything.

lua/neo-tree/sources/common/commands.lua
```diff
@@ -343,6 +346,53 @@ M.delete_visual = function(state, selected_nodes, callback)
   fs_actions.delete_nodes(paths_to_delete, callback)
 end
 
+M.preview = function(state)
+  local node = state.tree:get_node()
```
cseickel, line 350: Since this is being used as the event handler, I would short circuit here if the current window is not the same as `state.winid` to avoid unnecessary processing or unexpected behavior.

lua/neo-tree/sources/common/commands.lua
```diff
@@ -343,6 +346,53 @@ M.delete_visual = function(state, selected_nodes, callback)
...
+M.toggle_preview = function(state)
+  local preview_event = {
+    event = events.VIM_CURSOR_MOVED,
+    handler = utils.wrap(M.preview, state),
+    id = "preview-event",
+  }
+  if state.preview and state.preview.active then
+    state.preview:revert()
+    manager.unsubscribe(state.name, preview_event)
+  else
```
cseickel, line 391: What if the Neo-tree window is closed or the source is changed? I think it should probably end the preview mode. I think this could be handled by subscribing to `NeoTreeBufferLeave`, which would fire when the Neo-tree buffer loses focus for any reason.

cseickel (commented):
Overall this works pretty well.

lua/neo-tree/sources/common/commands.lua
mrbjarksen: Good note. I think its best to add the check to the event handler itself since this can also be used as a regular command.

lua/neo-tree/sources/common/commands.lua
mrbjarksen: I was hesitant to add it, but you're probably right. My thinking was that it would not be good for sources which benefit from activating the preview by default (such as diagnostics). How do you think it would be best to handle that? An event handler in the source's setup function should be enough, right?

cseickel: > An event handler in the source's setup function should be enough, right?

Yes. I think if you want to activate it automatically when showing a certain source, then you can subscribe to `NeoTreeBufferEnter` or `NeoTreeWindowAfterOpen` in the source's setup function, depending on what you want. I would not want to automatically enter preview mode though, so I would make that functionality optional.

mrbjarksen:
I've implemented the fixes you mentioned. It was more trouble than I was expecting as I ran into a mess of autocommands, mostly triggered by jumping windows in `get_appropriate_window`, which took a while to debug. It seems to be fixed now.

One thing I noticed was that entering a neo-tree buffer from another neo-tree buffer does not trigger `NeoTreeBufferLeave`, since the event handler for `BufEnter` returns early if the entered buffer belongs to neo-tree. I'm not sure if this is intentional or not, so I haven't touched it, but it means that the preview mode will not be turned off when switching sources. This could be fixed by adding an event handler for `NeoTreeBufferEnter` as well, if you don't want to change this.

cseickel:
> One thing I noticed was that entering a neo-tree buffer from another neo-tree buffer does not trigger `NeoTreeBufferLeave`, since the event handler for `BufEnter` returns early if the entered buffer belongs to neo-tree. I'm not sure if this is intentional or not, so I haven't touched it,

No, that's not intentional. I just never considered that situation before. It should trigger a `NeoTreeBufferLeave` followed by a `NeoTreeBufferEnter`. I'll put that in as a bug.

lua/neo-tree/sources/common/preview.lua
```diff
@@ -0,0 +1,209 @@
...
+    return
+  end
+  vim.api.nvim_win_set_cursor(self.winid, { (pos[1] or 0) + 1, pos[2] or 0 })
+  vim.api.nvim_win_call(self.winid, function()
+    vim.cmd("normal! zz")
+  end)
+end
+
+---Highlight the previewed range
+function Preview:highlight()
```
cseickel, line 167: What is this code supposed to do? I don't notice any special highlights when I use the preview, but the highlight group does have settings which are different from Normal.

cseickel: It doesn't seem to work for me. It also does not jump to the correct line for the focused diagnostic, is it supposed to do that?

image

mrbjarksen: Ah, you're probably just not using the `native-open-and-preview` branch I set up to test this. The preview and the open commands look to `node.extra.position` and `node.extra.end_position` to determine what to highlight and where to open. The main branch uses `node.extra.diag_struct` for its own open commands and does not set these fields.

cseickel: Ahhh, OK. I think this is good to merge now.

cseickel (approved):

lua/neo-tree/sources/common/preview.lua
mrbjarksen: Alright, sounds good. I might revisit the highlight logic at a later date, but for now it's fine.

barreiroleo:
Greatest feature!
When there's no open buffer (such as starting with the alpha-nvim greeting), disabling preview causes some errors:
Other than that, it works great.
```
E5108: Error executing lua: ...rt/neo-tree.nvim/lua/neo-tree/sources/common/preview.lua:150: Vim(append):Error executing lua callback: ..
.are/nvim/site/pack/packer/start/alpha-nvim/lua/alpha.lua:550: Cursor position outside buffer
stack traceback:
        [C]: in function 'nvim_win_set_cursor'
        ...are/nvim/site/pack/packer/start/alpha-nvim/lua/alpha.lua:550: in function 'draw'
        ...are/nvim/site/pack/packer/start/alpha-nvim/lua/alpha.lua:566: in function 'redraw'
        ...are/nvim/site/pack/packer/start/alpha-nvim/lua/alpha.lua:473: in function <...are/nvim/site/pack/packer/start/alpha-nvim/lua/a
lpha.lua:473>
        [C]: in function 'nvim_win_set_buf'
        ...rt/neo-tree.nvim/lua/neo-tree/sources/common/preview.lua:150: in function 'setBuffer'
        ...rt/neo-tree.nvim/lua/neo-tree/sources/common/preview.lua:73: in function 'revert'
        ...t/neo-tree.nvim/lua/neo-tree/sources/common/commands.lua:400: in function 'func'
        .../packer/start/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:695: in function <.../packer/start/neo-tree.nvim/lua/neo-tree/ui/rend
erer.lua:693>
stack traceback:
        [C]: in function 'nvim_win_set_buf'
        ...rt/neo-tree.nvim/lua/neo-tree/sources/common/preview.lua:150: in function 'setBuffer'
        ...rt/neo-tree.nvim/lua/neo-tree/sources/common/preview.lua:73: in function 'revert'
        ...t/neo-tree.nvim/lua/neo-tree/sources/common/commands.lua:400: in function 'func'
        .../packer/start/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:695: in function <.../packer/start/neo-tree.nvim/lua/neo-tree/ui/rend
erer.lua:693>
```

cseickel:
😱 OH NO! A stack trace that includes two different plugins! 😱 

Seriously though, I'm going to open a new issue for this.

barreiroleo:
Maybe a check before reverting the buffer is all, if there isn't a previous buffer open a new one... Just thinking out loud, haven't read the code yet.

PR #494, opened 2022-08-23
feat: added drop and tab drop commands

rkoval:
Adds exposure to `:drop` and `:tab drop` for ways to open up files in neo-tree


cseickel:
Thanks for contributing! I changed your target to the `main` branch. Please target `main` for any future contributions as well.

One thing that needs to be added is entries in the help file for these two new commands, somewhere around here:
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/a7d6f05e57487326fd70b24195c3b7a86a88b156/doc/neo-tree.txt#L184

Also, we usually add all commands to the appropriate `mappings` section in `defaults.lua` and the example in the `README`, even if they are commented out and not mapped by default.

rkoval:
Got it! Potentially fixed in [64c40bb](https://github.com/nvim-neo-tree/neo-tree.nvim/pull/494/commits/64c40bbe2304885cd47cea61c358f46873f0b58a). How does that look?

doc/neo-tree.txt
```diff
@@ -198,6 +198,13 @@ s             = open_vsplit: Same as open, but opens in a vertical split.
 
 t             = open_tabnew: Same as open, but opens in a new tab.
 
+                  open_drop: Same as open, but opens with the `:drop` command.
```
cseickel, line 201: Since it is a built-in help topic, you can actually just wrap it in `|` and it will become a hyperlink to that help topic:
```suggestion
                  open_drop: Same as open, but opens with the |:drop| command. 
```

doc/neo-tree.txt
```diff
@@ -198,6 +198,13 @@ s             = open_vsplit: Same as open, but opens in a vertical split.
 
 t             = open_tabnew: Same as open, but opens in a new tab.
 
+                  open_drop: Same as open, but opens with the `:drop` command. 
+                             (run `:help drop` for more info)
```
cseickel, line 202: Which would make this line unnecessary.

```suggestion
```

doc/neo-tree.txt
```diff
@@ -198,6 +198,13 @@ s             = open_vsplit: Same as open, but opens in a vertical split.
...
+
+              open_tab_drop: Same as open, but opens in a new tab with the 
+                             `:tab drop` command. (run `:help drop` for more
```
cseickel, line 205: Same as above, this can be changed to
```suggestion
                             `:tab drop` command. (see |:drop| for more
```

cseickel (changes requested):

doc/neo-tree.txt
rkoval: Ahaa! I am clearly a novice! Good call: https://github.com/rkoval/neo-tree.nvim/commit/82cd313328ebfaf42409611a739d1ba5f8467173

doc/neo-tree.txt
rkoval: https://github.com/rkoval/neo-tree.nvim/commit/82cd313328ebfaf42409611a739d1ba5f8467173

doc/neo-tree.txt
rkoval: https://github.com/rkoval/neo-tree.nvim/commit/82cd313328ebfaf42409611a739d1ba5f8467173

cseickel (approved):
Looks good, thanks!

PS: Pay no attention to the failing test. It's not from your code and it seems to be a flaky test rather than a real problem.

PR #502, opened 2022-08-29
feat(preview): add `use_float` option for preview mode

cseickel:


cseickel:
I really wanted to do a little more, including adding tests, but I am just not going to have the time for a while so I'll just merge it now so I can stop holding up releases and other small tweaks and fixes.

axieax:
Love this feature. Just wanted to raise a bug where the calling window (e.g. side = right) doesn't get restored on revert_preview (become side = left).

teddywaweru:
a bug I've noted whle running the version that comes re-installed n AstroNvim is that if I press `Enter` while in Preview mode, the floating pane does not hide on pressing `P`. I have to force restart the console entirely.

PR #558, opened 2022-10-07
fix(highlight): tab separator respects transparent background

emxxjnm:
Before:
WeChatdc2aac08516f9ccb67117192343ed4f3

After:
WeChat6242ff23b6c30e9d108b3b40ed1aba68


cseickel:
Thanks for the contribution, but this particular change does not belong in the code base. The defaults that exist suit most people, but are configurable. If you want to change the colors of highlight groups, just do it in your own config. This should do the same thing, although you may have to run it after calling setup on neo-tree:

```
hi NeoTreeTabActive guifg=NONE guibg=NONE
hi NeoTreeTabInactive guifg=NONE guibg=NONE
```

PR #568, opened 2022-10-15
fix: Clickable filesystem source winbar component closes neo-tree immediately

### LINKED ISSUE #565: Clickable filesystem source winbar component closes neo-tree immediately
Clicking the filesystem source in my neo-tree winbar seems to instantly close the neo-tree buffer. Other components seem to not have this issue when clicked.
Note: navigating between sources with `<` and `>` still works as intended.

Config: https://github.com/axieax/dotconfig/blob/4870bba67f90f068f9e01dae547c49d1c0cbadb2/nvim/lua/axie/plugins/neotree.lua

pysan3:
Hi, I think this is enough to fix #565
I'm still investigating why `action = 'show'` fails on a floating window.

#### Some Notes
- `prev_source, next_source` (i.e. `<`, `>`) uses `action = 'focus'` instead
- Works without fix for `config.window.position = 'left'` and all other options except `'float'`
- Switching to other sources has no problem even with `'float'`

I thought that `action = 'show'` will cache or whatever and speed up the switch, but looking at the code, I don't think there are any downsides by doing it this way. (And this is how `<`, `>` was implemented in the first place, which did not cause any problem)

### DIFF
--- a/lua/neo-tree/ui/selector.lua
+++ b/lua/neo-tree/ui/selector.lua
@@ -384,11 +384,10 @@ _G.___neotree_selector_click = function(id, _, _, _)
     log.warn("state not found for window ", winid, "; ignoring click")
     return
   end
-  vim.api.nvim_set_current_win(winid)
   require("neo-tree.command").execute({
     source = sources[source_index],
     position = state.current_position,
-    action = "show",
+    action = "focus",
   })
 end
 


cseickel:
The only difference between `show` and `focus` is that `show` will return focus to the original window after neo-tree opens, because opening neo-tree always forces the focus to the new neo-tree window. Also, if the requested window is already open, then `show` will do nothing while `focus` will make that window active.

All of this can have different behavior for `position=float` because floats will automatically close when they lose focus. `show` should never be used for a float.

cseickel (approved):
This seems to work perfectly. Thanks!

PR #573, opened 2022-10-18
hide diagnostics and git_status when a directory is expanded.

haoyun:
I think this is a matter of taste and I am not sure if this possible via configuration. Hence I create this PR.

Suppose we have
```
dir1
|- dir2
   |- dir3
      |- dir4
         |- file1
         |- file2
         |- file3
         |- file4
```
and there is an error in `file4`, currently, when the directories are full expanded, E (or some Icon) is shown on all directories, which I think might be a bit distractive.
```
dir1                 E 
|- dir2              E
   |- dir3           E
      |- dir4        E
         |- file1
         |- file2
         |- file3
         |- file4    E
```
With the pull request, the icon is shon on folders only if its collapsed, i.e., only on the deepest level node.
```
dir1
|- dir2
   |- dir3           E
```
```
dir1
|- dir2 
   |- dir3
      |- dir4        E
```
```
dir1
|- dir2
   |- dir3
      |- dir4
         |- file1
         |- file2
         |- file3
         |- file4    E
```

Images in README seem to have the latter behavior but on my machine it's the former case. Did I misconfigure something?


cseickel:
> Images in README seem to have the latter behavior but on my machine it's the former case. Did I misconfigure something?

No, it's just that the default config is to only show errors at the folder level, and there just did not happen to be errors to show in those screenshots.


This is such a simple and elegant solution to the problem, I can't believe it did not occur to me! I do think it is a good idea, but I'm trying to decide if this is a breaking change or not. It might be a good idea to add a config option to enable this behavior, and I'll make it the default in v3.0.

lua/neo-tree/sources/common/components.lua
```diff
@@ -77,6 +77,9 @@ end
 M.diagnostics = function(config, node, state)
   local diag = state.diagnostics_lookup or {}
   local diag_state = diag[node:get_id()]
+  if node:is_expanded() then
```
cseickel, line 80: This should probably be:
```suggestion
  if node.type == "directory" and node:is_expanded() then
```

Technically there are nested files and diagnostic messages (in the external diagnostics source) that can be expanded but should still show the diagnostics markers.

nyngwang:
@haoyun I was looking for the option to do the same thing and realized it didn't exist. Looking forward to the merging of the PR, since I don't need folder-level diagnostics hints when folders got expanded.

cseickel:
I think this is perfect. Most people would probably prefer the new behavior, but if they don't there is a simple fix.

Thanks!

cseickel (dismissed):

cseickel:
@haoyun You''l just need to satisfy the stylua check before it can be merged: https://github.com/nvim-neo-tree/neo-tree.nvim/actions/runs/3291356600/jobs/5446832713

It's hard to see, but the issue is just that the indentation of those lines is wrong, you need to remove two spaces.

haoyun:
Thanks. The extra spaces were added by the language server `sumneko/lua-language-server`. I've removed them.

cseickel (approved):

Issue #576, opened 2022-10-22
Closing a window and opening a file moves Neo-tree to the wrong place


abdullatif-k:
Hello, love the plugin. Have a small issue. Neo-tree should stay on the left, however the following happens:

Scenario (A)

1.Open an empty buffer, then open Neo-tree with 
`:Neotree position=left toggle=true action=focus<CR>`
2. At this point only Neo-tree on the left, and an empty buffer on the right are on the screen.
3. Close the empty window with `:q`
4. Open another file in Neo-tree
5. Notice how Neo-tree now opens on the right.

Scenario (B)

I have an autocmd which opens Neo-tree whenever I open nvim on a directory. I don't depend on Neo-tree's own hijacking behavior for 2 reasons, (1) I've disabled netrw (2) I'm lazy loading Neo-tree on the `Neotree` command. This is the autocmd:
```vimscript
autocmd BufEnter * if isdirectory(expand("%")) | enew | Neotree
```
The same problem as Scenario (A) exists, however it also happens when you close the buffer. In scenario (A) closing the buffer doesn't close the window, but in in (B) it does. So you end up with the same issue.

cseickel:
You are correct in that it does not handle this situation correctly. When it is the only window, it will simply issue a `vsplit` command to open a new window. For me, this opens to the right regardless of whether Neotree was opened as `position=left` or `position=right`. That is because I have these settings in my init.vim:

```
set splitbelow
set splitright
```

If you set those, it will always open new windows to the right as expected.

It should ultimately respect your desired position regardless of any other settings. I'll mark this as a bug and fix it when I can.

abdullatif-k:
> You are correct in that it does not handle this situation correctly. When it is the only window, it will simply issue a `vsplit` command to open a new window. For me, this opens to the right regardless of whether Neotree was opened as `position=left` or `position=right`. That is because I have this setting in my init.vim:
> 
> ```
> set splitright
> ```
> 
> If you set that, it will always open new windows to the right as expected.
> 
> It should ultimately respect your desired position regardless of any other settings. I'll mark this as a bug and fix it when I can.

I think this PR will suffice as fix. 
https://github.com/nvim-neo-tree/neo-tree.nvim/pull/577

PR #577, opened 2022-10-22
Fix new window orientation when Neo-tree is only window

abdullatif-k:
This fixes https://github.com/nvim-neo-tree/neo-tree.nvim/issues/576 for me.


cseickel (approved):

cseickel:
Thanks!

Issue #547, opened 2022-09-28
Feature Request: Auto Width or Full Name Under Cursor


jjiangweilan:
Hi thanks for the awesome work! I just switched from nvim-tree. And I think nvim-tree has one feature I use but I did't find here is the auto width of the explore.
I have projects that use really deep folder hierarchy. The files are easily out of the explorer region which makes it hard to navigate and read. I have read the docs and only see "fit_content" for container which I think itn't what I want here.

nyngwang:
> [...] The files are easily out of the explorer region which makes it hard to navigate and read. [...]

I also encountered this and found it annoying. Maybe showing a one-word height horizontal floating window for showing the full name could be a cool idea for implementing this.

pysan3:
Hi, just wanted to announce that we now have an auto width expand option added! (details: #580 and #629)

It has been ported to `v2.x` branch and seems to be stable enough.

This feature is accessible via two ways.

1. Adding the following option to the configuration. This makes neotree always expand to fit the longest line.
```lua
require("neo-tree").setup({
  window = {
    auto_expand_width = true, -- default: false
  },
})
```

2. Alternatively, there is a keybind `e` that toggles the above option.
  IMO the best way to go is to keep `window.auto_expand_width = false` and toggle the feature whenever needed with this keybind.

I really wanted to add the floating window feature as well, but unfortunately I haven't got any knowledge on floating window management so I would really appreciate if someone could implement this feature as well :)

nyngwang:
@pysan3 am using.

btw, you forgot to mention in 2.(while you're recommending it) that the toggle is named `toggle_auto_expand_width`. (since I have `use_default_mappings = false,`).

<img src="https://user-images.githubusercontent.com/24765272/205458554-7fe17dcd-f3d4-41a3-abae-96dbf70f94c6.png" width="400">

nhat-vo:
Closing this as solved.

PR #580, opened 2022-10-24
feat: Auto Width or Full Name Under Cursor (related to #547)

pysan3:
Hi, I love the extensibility of this plugin.

I think this feature (#547) is quite useful, especially when you want to peek a filename or dealing with very deep file structures.

I tried to implement this feature with the following configs and I would like to hear your opinions.

### 1. Added option `window.auto_expand_width`

Default: `false`

Usually, long file names and deep hierarchy folders (by the indents) are truncated and goes underneath modified symbols etc. However, by setting this to `true`, the filename is not truncated and instead, the window width is expanded (ignoring `window.width` option) to show the whole content.

|Default|`window.auto_expand_width = true`|
|-|-|
|image|image|

What it differs from `width = "fit_content"` is that the window width expands even after launching neotree, meaning that when a new folder with a very longfile name is expanded, neotree window adjusts its width to fit the new filename.

### 2. Added command `toggle_auto_expand_width` (with keybind)

Although I added the above feature, I actually like the fading effect on normal circumstances.
So I created a keybind that toggles `window.auto_expand_width`.

In my usecase (and IMO recommended to everybody) is to keep `window.auto_expand_width` option to `false` and toggle the feature whenever needed.

Currently it is binded to `e`. In the gif bellow, I am pressing `e` several times after opening neotree.

## Know Issues

There are some issues with the current implementation. However, I don't think this is a big problem because, first of all, not so many people are going to use (or even notice) this feature, and second, it hardly effects the usability.

- Window width of the lines above the `very-long-file-name.txt` is not updated
  - This is pretty much impossible to fix because the length of lines are calculated on the fly.
  - Not a big problem since it will get fixed by `renderer.redraw(state)`, i.e. doing any of the followings
    - reopen neotree
    - open another directory or make a new file or do something that will fire an update
- The default keybind `e` conflicts with `example.window.example_command`?
  - I have never used this keybind and have no idea what it is for.
  - I would like to hear @cseickel and others' ideas about a better keybinding.
- Currently disabled for `window.position = "float"`
  - With the floating window, it looks awkward after resizing the window.
  - Therefore, I currently disable this feature on floating window.
  - Floating windows tend to have a wider width by default and I find myself rarely overflowing the width in the first place. (Fig: tested on floating windows; The right side of the window is kinda broken.)
Screenshot_20221024_170959

I hope this helps,
pysan3

### DIFF
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -287,6 +287,7 @@ local config = {
     position = "left", -- left, right, top, bottom, float, current
     width = 40, -- applies to left and right positions
     height = 15, -- applies to top and bottom positions
+    auto_expand_width = false, -- expand the window when file exceeds the window width. does not work with position = "float"
     popup = { -- settings that apply to float position only
       size = {
         height = "80%",
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -481,7 +481,7 @@ local config = {
     window = {
       mappings = {
         ["<cr>"] = "toggle_node",
-        ["e"] = "example_command",
+        ["<C-e>"] = "example_command",
         ["d"] = "show_debug_info",
       },
     },
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -338,6 +338,7 @@ local config = {
       ["p"] = "paste_from_clipboard",
       ["c"] = "copy", -- takes text input for destination, also accepts the config.show_path option
       ["m"] = "move", -- takes text input for destination, also accepts the config.show_path option
+      ["e"] = "toggle_auto_expand_width",
       ["q"] = "close_window",
       ["?"] = "show_help",
       ["<"] = "prev_source",
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -146,7 +146,13 @@ M.toggle_auto_expand_width = function(state)
   end
   state.window.auto_expand_width = state.window.auto_expand_width == false
   if not state.window.auto_expand_width then
-    vim.api.nvim_win_set_width(0, state.window.width)
+    if (state.window.last_user_width or state.window.width) >= vim.api.nvim_win_get_width(0) then
+      state.window.last_user_width = state.window.width
+    end
+    vim.api.nvim_win_set_width(0, state.window.last_user_width)
+    state.win_width = state.window.last_user_width
+    state.longest_width_exact = 0
+    log.trace(string.format("Collapse auto_expand_width."))
   end
   renderer.redraw(state)
 end
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -140,6 +140,17 @@ M.close_window = function(state)
   renderer.close(state)
 end
 
+M.toggle_auto_expand_width = function(state)
+  if state.window.position == "float" then
+    return
+  end
+  state.window.auto_expand_width = state.window.auto_expand_width == false
+  if not state.window.auto_expand_width then
+    vim.api.nvim_win_set_width(0, state.window.width)
+  end
+  renderer.redraw(state)
+end
+
 local copy_node_to_clipboard = function(state, node)
   state.clipboard = state.clipboard or {}
   local existing = state.clipboard[node.id]
--- a/lua/neo-tree/sources/common/container.lua
+++ b/lua/neo-tree/sources/common/container.lua
@@ -187,7 +187,8 @@ local merge_content = function(context)
   -- * Repeat until all layers have been merged.
   -- * Join the left and right tables together and return.
   --
-  local remaining_width = context.container_width
+  local huge_number = 2000
+  local remaining_width = context.auto_expand_width and huge_number or context.container_width
   local left, right = {}, {}
   local left_width, right_width = 0, 0
 
@@ -253,6 +254,9 @@ local merge_content = function(context)
     end
   end
 
+  if context.auto_expand_width then
+    remaining_width = context.container_width + remaining_width - huge_number
+  end
   if remaining_width > 0 and #right > 0 then
     table.insert(left, { text = string.rep(" ", remaining_width) })
   end
@@ -271,6 +275,7 @@ M.render = function(config, node, state, available_width)
     left_padding = config.left_padding,
     right_padding = config.right_padding,
     enable_character_fade = config.enable_character_fade,
+    auto_expand_width = state.window.auto_expand_width and state.window.position ~= "float",
   }
 
   render_content(config, node, state, context)
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -15,7 +15,7 @@ local M = { resize_timer_interval = 50 }
 local ESC_KEY = vim.api.nvim_replace_termcodes("<ESC>", true, false, true)
 local default_popup_size = { width = 60, height = "80%" }
 local floating_windows = {}
-local draw, create_window, create_tree
+local draw, create_window, create_tree, render_tree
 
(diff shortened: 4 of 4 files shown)


cseickel:
> Hi, I love the extensibility of this plugin.

Thanks!



> Window width of the lines above the `very-long-file-name.txt` is not updated
> * This is pretty much impossible to fix because the length of lines are calculated on the fly.
> * Not a big problem since it will get fixed by `renderer.redraw(state)`,

You can just call `renderer.redraw()` a second time after you have determined and set the correct width.



> * The default keybind `e` conflicts with `example.window.example_command`?

Don't worry about that. That's just an example for devs to see how sources work. You can just move that to another unused binding.

-----

I was originally picturing a floating window popup that shows the full name of a truncated item on hover. Have you thought about that? It is a standard way of handling this situation in other apps.

If we do stick with auto expand, I would keep track of the width prior to the auto expand instead of just using the default value from the settings. I know I will usually adjust the width manually after opening to fit different projects and screen sizes.

pysan3:
> You can just call `renderer.redraw()` a second time after you have determined and set the correct width.

I tried to do this but as far as I had not mistaken something, it goes into an infinite recursive function call and results to an error.

> Don't worry about that. That's just an example for devs to see how sources work. You can just move that to another unused binding.

[1f32cda](https://github.com/nvim-neo-tree/neo-tree.nvim/pull/580/commits/1f32cda04099c0215be887633881d81803eb1b0f) : Changed to `<C-e>`

> I was originally picturing a floating window popup that shows the full name of a truncated item on hover. Have you thought about that? It is a standard way of handling this situation in other apps.

- First of all, I thought it was easier to implement it this way rather than managing a floating popup for each cursor move.
- I didn't know other apps do it with popups (do you mean other plugins like nvimtree? or IDEs like VSCode?), but I think expanding the viewer makes more sense because I see many people manually dragging the window edge and expand it to see the whole name.
- Actually I just wanted to implement this for me, and since I don't think the two features conflict, we might be able to implement both ways. 

> I would keep track of the width prior to the auto expand instead of just using the default value from the settings.

I once thought about this but currently it is difficult.
With the current implementation, the window is not expanded at once (based on the longest line), but it is expanded each time there is not enough width.
So, for example if the file structure is as below and the initial width is like 5, the window width is increased for each row since the line gets longer. Since the width changes on each line, the previous width on `bar.txt` is what was set on `foo/` and not 5.
```txt
hoge/
  fuga/
    foo/
      bar.txt
```

cseickel:
> > You can just call `renderer.redraw()` a second time after you have determined and set the correct width.
> 
> I tried to do this but as far as I had not mistaken something, it goes into an infinite recursive function call and results to an error.

You would just have to keep track of it in the `state` variable. Something like `state.render_pass = "first"` on the initial pass to get the ideal size. Then at the end of render if `state.render_pass == "first"`, change it to `state.render_pass = "second"` and do one more render to get the right aligned icons correct.  

Another option is that you can take advantage of `state.longest_node`, which I estimate for the purpose of right aligning symbols when opened in `"current"` position. That is a rough estimate right now but it only requires one pass and is known before render. It's currently used here here:
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/35e9f9b53d05bfec90314865787cd920a8b468aa/lua/neo-tree/ui/renderer.lua#L331-L334

But you can use it earlier to just change the window size before this point when auto width is on.



> > I was originally picturing a floating window popup that shows the full name of a truncated item on hover. Have you thought about that? It is a standard way of handling this situation in other apps.
> 
> * First of all, I thought it was easier to implement it this way rather than managing a floating popup for each cursor move.

It's definitely easier the way you are doing it.



> * I didn't know other apps do it with popups (do you mean other plugins like nvimtree? or IDEs like VSCode?),

I mean gui apps like VSCode.



> * Actually I just wanted to implement this for me, and since I don't think the two features conflict, we might be able to implement both ways.

Fair enough. In that case I would make sure the option is renamed and the value is a string rather than a boolean so it can be a three state option to expand the width of the window, show a float, or do nothing.




> > I would keep track of the width prior to the auto expand instead of just using the default value from the settings.
> 
> I once thought about this but currently it is difficult. With the current implementation, the window is not expanded at once (based on the longest line), but it is expanded each time there is not enough width.

I feel like there must be a way to logically distinguish between a user change and your changes, but maybe it's not worth the effort to do so. Don't worry about it.

pysan3:
Sorry for the late update but I revised the code quite a lot and I think I was able to fix most of the bugs and issues.

One warning is that you should call `render_tree(state)` instead of `state.tree:render()` due to the fact that it needs to check the width __after__ the rendering the whole tree.
It cannot be done only with `state.tree:render()` because in that case, Nuitree calls `prepare_node` (`render.lua`) for each row and that's the only place to decide the width. I could not find kinda `POST_RENDER` hook that can be attached to Nuitree.

cseickel:
> One warning is that you should call `render_tree(state)` instead of `state.tree:render()` due to the fact that it needs to check the width **after** the rendering the whole tree.

No problem, that was a nice solution to the problem. There is one more place to replace `tree:render()` though:
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/683d0278b76e7f129866f1a9a59aa570ee369313/lua/neo-tree/ui/renderer.lua#L503

I tested it out and I think it works well. The only other left to do is to add some documentation about the new feature to the help file and add the mapping in the README's Quickstart section.

pysan3:
I know that line and I think there's no need to call `render_tree` in that situation.

Possible change in width only happens when opening a new window with `auto_expand_width == true` in the first place, or toggling between true and false with keybind (`e`).

Therefore, I couldn't think of any situation of needing to change width with that `tree:render()`.
Am I right?

But in any cases, I think it's not difficult to change to `expand_to_node(state, node)` because this function is only referenced once.

> The only other left to do is to add some documentation about the new feature to the help file and add the mapping in the README's Quickstart section.

May I hand it to you to write the docs and README because I was not sure enough where in the document I should write about this.
I think the content should be enough with what is written as a comment in `defaults.lua`.
```lua
config = {
  window = {
    auto_expand_width = false, -- expand the window when file exceeds the window width. does not work with position = "float"
  }
}
```

cseickel:
> May I hand it to you to write the docs and README because I was not sure enough where in the document I should write about this.

Sure. Thanks for contributing!

cseickel (approved):

Kerwood:
Nice job.. This is an awesome feature. Thanks :)

dudicoco:
I've tried using this feature but the expanded window still cuts off the end of the file name:
Screenshot 2024-01-11 at 11 12 48

Should I open a new issue?

cseickel:
@dudicoco it looks like it is working fine to me, assuming "ten" is the end of the filename. You may just want to turn off character fading if you don't like that look.

You can disable character fading by adding this to your config:

```lua
        default_component_configs = {
          container = {
            enable_character_fade = false
          },
```

dudicoco:
> @dudicoco it looks like it is working fine to me, assuming "ten" is the end of the filename. You may just want to turn off character fading if you don't like that look.
> 
> You can disable character fading by adding this to your config:
> 
> ```lua
>         default_component_configs = {
>           container = {
>             enable_character_fade = false
>           },
> ```

@cseickel the file name ends with `.txt`, that part is cut off.
`enable_character_fade = false` did not fix the issue.

dudicoco:
@cseickel i've found out the cause for this - I have a config to add relative line numbers to the neo-tree buffer:
```
      require("neo-tree").setup({
        event_handlers = {
          {
            event = "neo_tree_buffer_enter",
            handler = function(arg)
              vim.cmd [[
                setlocal relativenumber
              ]]
            end,
          }
        },
      }
    )
```

Commenting out this block fixes the issue.
How can we have relative line numbers work properly together with auto width?

cseickel:
> How can we have relative line numbers work properly together with auto width?

It sounds like a bug where the needed with is not accounting for the width of the line numbers. Open a bug report, or better yet, find the bug and submit a PR.

Issue #308, opened 2022-04-18
Update file handler/buffer after move/rename


gegoune:
Currently renaming `a.lua` to `b.lua` leaves me with open buffer named `a.lua` (saving which recreates `a.lua` on filesystem) and no `b.lua` gets opened.

Expected behaviour is such that a.lua buffer gets removed and b.lua is opened in its place (not with the same buffer number, but 'instead').

_Originally posted by @gegoune in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/47#issuecomment-1085097404_

nyngwang:
There is also a related error message shown upon that the neo-tree got closed and the cursor went back to the no-longer-exist buffer(i.e. `a.lua` in the example). While it is not harmful but definitely related so I provided this info here.

gegoune:
Without checking, afair, it's regular (neo)vim's messages saying that file for that buffer no longer exists. It's not an issue in itself as much as it is product of not updating/removing that buffer by neo-tree (which this ticket is about).

cseickel:
This is now working in `main` and will be in the next release.

nyngwang:
@cseickel: This causes an error:

```
E5108: Error executing lua: /Users/ningwang/.config/nvim/init.lua:35: Vim(write):E13: File exists (add ! to override)
stack traceback:
        [C]: in function 'cmd'
        /Users/ningwang/.config/nvim/init.lua:35: in function </Users/ningwang/.config/nvim/init.lua:35>
Press ENTER or type command to continue
```

nyngwang:
:( waiting for the fix.

cseickel:
I'll fix it, but to be clear it's still writable and not a fake buffer at all. You just have to add the `!` to your write the command the first time you write it.

```
:w!
```

cseickel:
So the E13 error issue is resolved, and I also realized that previously there was no handling for renaming buffers in a directory that was renamed, that was fixed too.

I did find a new bug related to cut/pasting files. I'm not sure if it's related to this or not but I'm investigating that right now.

Gelio:
[...]

PR #591, opened 2022-11-05
fix: crash when renaming files that have sync-formatting autocmd

Gelio:
When renaming or moving files/directories, neo-tree used to change the name of renamed buffers while retaining the same buffer numbers. It did not delete the old buffers and open new ones for editing.

This ended up crashing Neovim in the following scenario:

* a `BufWritePre` autocmd was created to call `vim.lsp.buf.format({ async = false })`
* there was an LSP client registered for the buffer
* the buffer was opened and then put in the background (e.g. by `:Neotree position=current`)
* the file was renamed/moved in neo-tree

The operation that caused the Neovim crash was the `silent! write!` that neo-tree executes on the buffer right after it is renamed. During that write, synchronous formatting is attempted in the `BufWritePre` autocmd.

My assumption is that the LSP server does not get a notification about the file rename and sends some invalid formatting information back to Neovim. Neovim cannot interpret it and crashes.

[Crash backtrace: bt.txt](https://github.com/nvim-neo-tree/neo-tree.nvim/files/9943883/bt.txt)

https://user-images.githubusercontent.com/889383/200121515-db2c2cc9-fa0a-42bd-96c7-dffa50312cf1.mp4

The crash does not happen if the renamed buffer is currently visible, rather than being hidden (e.g. when the renamed buffer is in one window and neo-tree is opened in a vertical split). The crash also does not happen if I don't run `vim.lsp.buf.format` on `BufWritePre`. Neither of these is a great workaround, as I don't want to remember to keep files open or remove my formatting autocmd.

The approach to handling file renamed used in this commit is the same as used in the [`vim.lsp.util.rename`](https://github.com/neovim/neovim/blob/234b8c5f3d57294dda06dbc6c1760e5983bd2c19/runtime/lua/vim/lsp/util.lua#L751-L775). Instead of reusing the buffers in Neovim before and after rename, neo-tree deletes old buffers for the file paths before the rename, and opens new buffers for files after the rename. If old buffers were shown in some Neovim window, the code opens the new buffers in these windows.

https://user-images.githubusercontent.com/889383/200121526-0e1b2a63-b263-479f-b6b3-f18a90236445.mp4

The code that handles modified buffers during rename calls `silent! write!` on the new buffers. This no longer crashes Neovim. The downside is that LSP clients are not attached at the moment `write!` is called, which means the files after the rename will not be formatted after neo-tree saves the modified content. Overall, this seems like a lesser evil compared to Neovim crashing on each rename.

All in all, the functionality seems to be the same before and after my fix, with the difference that I no longer get Neovim crashes when renaming files that have a formatting autocmd.

Related to #308

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/fs_actions.lua
+++ b/lua/neo-tree/sources/filesystem/lib/fs_actions.lua
@@ -39,6 +39,18 @@ local function clear_buffer(path)
   end
 end
 
+---Opens new_buf in each window that has old_buf currently open.
+---Useful during file rename.
+---@param old_buf number
+---@param new_buf number
+local function replace_buffer_in_windows(old_buf, new_buf)
+  for _, win in ipairs(vim.api.nvim_list_wins()) do
+    if vim.api.nvim_win_get_buf(win) == old_buf then
+      vim.api.nvim_win_set_buf(win, new_buf)
+    end
+  end
+end
+
 local function rename_buffer(old_path, new_path)
   local force_save = function()
     vim.cmd("silent! write!")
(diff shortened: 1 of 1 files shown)


cseickel (approved):
Awesome work, thanks!

Gelio:
Thanks for merging my PRs!

What is the schedule of merging `main` into `v2.x`? I'm wondering if I should switch to `main` temporarily to have the fixes or wati a short time to see them on the stable branch

cseickel:
I just released. My normal rule is to let new changes sit in main for at least a day to see if there are any bugs reported.

Gelio:
I see, thanks for quick action!

PR #621, opened 2022-11-25
fix(git): make `mark_ignored` async

musjj:
This PR introduces an alternative async interface for `git.mark_ignored`. By providing a callback to the function, it will run the git operations via `luv.spawn` instead of `vim.fn.systemlist`.
In the async interface, the files to be checked are fed through the stdin, instead of through the CLI arguments. This is because certain collection of filenames causes `luv` to fail to spawn the git process (maybe because of the length or certain unicode characters?). If there are any performance issues caused by the pipe method, I'll try to investigate further on it.

### DIFF
--- a/lua/neo-tree/git/ignored.lua
+++ b/lua/neo-tree/git/ignored.lua
@@ -1,3 +1,5 @@
+local Job = require("plenary.job")
+
 local utils = require("neo-tree.utils")
 local log = require("neo-tree.log")
 local git_utils = require("neo-tree.git.utils")
@@ -38,10 +40,11 @@ local get_root_for_item = function(item)
   return root
 end
 
-M.mark_ignored = function(state, items)
+M.mark_ignored = function(state, items, callback)
   local folders = {}
   log.trace("================================================================================")
   log.trace("IGNORED: mark_ignore BEGIN...")
+
   for _, item in ipairs(items) do
     local folder = utils.split_path(item.path)
     if folder then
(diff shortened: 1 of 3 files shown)


musjj:
Just realized that some tests are failing, not sure what's causing this 🤔

musjj:
@cseickel Sorry for bugging you again, but any idea why this is happening?
Apparently during the test, directories aren't being folded correctly when `async_git_status` is set to `true`, but I'm not experiencing this issue on my own setup.
I'm guessing that there's some kind of race condition because it's asynchronous?

musjj:
Alright, I figured out what's happening, it is some kind of race condition.
The error messages completely confused me, because neo-tree's test utilities mixed up the `{expected}` and `{actual}` arguments: https://github.com/nvim-neo-tree/neo-tree.nvim/blob/7c6903b05b13c5d4c3882c896a59e6101cb51ea7/tests/utils/init.lua#L94, which makes for a very confusing message.
For some reason, there's a bit of a delay between the sending the keys, and the folder tree actually expanding when `git_status_async` is `true`. I'm still not sure about the cause, but putting in an arbitrary delay in the test seem to fix the issue.

cseickel:
> there's a bit of a delay between sending the keys, and the folder tree actually expanding when `git_status_async` is `true`. I'm still not sure about the real culprit causing that delay, but putting in an `vim.wait` after sending the keys seem to fix the issue.

That seems perfectly reasonable, if git status is async and the testing code is not, you woudl certainly expect the testing code to execute first.

----

I think this PR looks good, but I want to take the time to really read it carefully before merging and I haven't gotten that time yet. I'll do that tonight or this weekend.

lua/neo-tree/git/ignored.lua
```diff
@@ -83,26 +74,77 @@ M.mark_ignored = function(state, items)
...
+          local result
+          if code ~= 0 then
+            log.debug("Failed to load ignored files for", state.path, ":", self:stderr_result())
+            result = {}
+          else
+            result = self:result()
+          end
+          vim.list_extend(all_results, process_result(result))
+          progress = progress + 1
+          if progress == #jobs then
```
cseickel, line 122: I'm always on the hunt for unexpected memory leaks. In this case, the garbage collector _should_ clean this up, but just in case, we don't really need to keep a reference to these jobs at all. The number of jobs is equal to the number of folders, so you can just compare that instead.
```suggestion
          if progress == #folders then
```

Then you can just go ahead and start the job as soon as it is defined without storing any reference to it at all. i.e. 
```lua
Job:new({ 
  ...
}):start()
```

cseickel (commented):
Looks good, I just have one nit-pick comment.

lua/neo-tree/git/ignored.lua
cseickel: Ah yes, you are absolutely correct. I was thinking that folders was a list for some reason.

cseickel (approved):
This is good to go. Thanks!

PR #625, opened 2022-11-26
feat: add `on_rename_file` hook

sidlatau:
This will enable to change of imports of the renamed file before the actual file rename.


cseickel:
@sidlatau Thanks for submitting the PR. One thing I would change is that we already have an event system that would suit this situation, and I wouldn't want to add a new `hooks` configuration.

Here is an example with something similar:

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/b68ebd1de117c6ad577016242b62797fb6b7ad69/lua/neo-tree/utils.lua#L474-L479

And an example of that event being consumed: https://github.com/nvim-neo-tree/neo-tree.nvim/wiki/Recipes#custom-window-chooser-for-file-open-commands

sidlatau:
Thanks for the provided example. But the problem is, that in my case I need an event handler to call async code and wait until a callback is called - only then I need to return control to `neo-tree` to continue with file rename/move. If I add code in event_handler, it does not wait until callback is called, `apply_workspace_edit` will be called after control is returned to `neo-tree`:
```lua
  event_handlers = {
    {
      event = "before_file_rename",
      handler = function(data)
        local params = ...
	lsp_client.request("workspace/willRenameFiles", params, function(err, result)
	  vim.lsp.util.apply_workspace_edit(result, dartls_client.offset_encoding)
	end)
      end,
    },
```
Do you have any ideas on how to make it work in blocking way using events system?

cseickel:
> Do you have any ideas on how to make it work in blocking way using events system?

You don't actually need to do it in a blocking way. One of the keys passed in the `data` object can be a callback to actually perform the rename. Then you can return `{ handled = true }` to indicate that the event is being handled elsewhere and to not continue on to do the rename in the function that called the event.

The `FILE_OPEN_REQUESTED` example uses the `{ handled = true }` technique.

sidlatau:
> The `FILE_OPEN_REQUESTED` example uses the `{ handled = true }` technique.

Thanks for the explanation, It finally clicked to me, how to use the events system without blocking, nice solution :) Just had an issue with events results - they are actually not returned from the `fire_event_internal` function.  Not sure about that `queue` - now it will return the last event handler result - is that ok?

cseickel:
> Just had an issue with events results - they are actually not returned from the `fire_event_internal` function. Not sure about that `queue` - now it will return the last event handler result - is that ok?

I don't think that is true, maybe there was some other mistake that gave you that impression? 

If you look at the actual code change you made, the only logical difference is that if there are multiple event handlers for the same event, then it will run all of them instead of stopping at the first one that reports `{ handled = true }`. That change is definitely incorrect. In the case where there is a single handler registered, there should not be any difference in behavior at all.

sidlatau:
I removed changes in `queue.lua`, I will investigate what is going on with event handler results separately. Are the remaining changes OK?

sidlatau:
> Sorry for the delay @sidlatau, these changes look like they will work perfectly. If they are not working for you, are you returning this from your handler?
> 
> ```lua
>     return { handled = true }
> ```

Yes, I am returning `{handled = true}` - I registered separate issue for this #631 where I asked more specific question about implementation.

Are there any actions from me needed to move forward this PR?

cseickel:
@sidlatau If you rebase this on `main`, the event handler should work as expected. Give it a try and let me know if this PR does what you need it to now.

sidlatau:
> @sidlatau If you rebase this on `main`, the event handler should work as expected. Give it a try and let me know if this PR does what you need it to now.

Rebased the PR on `main`. @cseickel I confirm, that it works now.

cseickel (approved):

Issue #624, opened 2022-11-26
`auto_expand_width` gets caught in infinite loop and hangs neovim


cseickel:
I've noticed that with `window.auto_expand_width` enabled, sometimes neovim will freeze when I open neo-tree as a sidebar. I njust tailed the logs when this happened and saw this scrolling by:
![image](https://user-images.githubusercontent.com/5160605/204097694-8ecd44bd-7796-431d-8645-5c720bb8d7b8.png)

Related to #580 
@pysan3 Can you take a look at this?

pysan3:
Hi, I don't have a pc with me right now so I can't test it right now but it seems that the estimated length of the expanded width is too big that it cannot fit the nvim window. 

When there is not enough width for the expanded width, it tries to expand more indefinitely falling into an infinite loop which is causing the nvim to freeze. 

I see that the estimated width is over 10,000 but did you purposely put a very very long file name?
Or maybe expanding a recursive directory symlink might explain this problem?

Could you give me a screenshot shot of the sidepanel when setting it off but with a quite big window.width in the first place?

pysan3:
Oh, the estimated width increments one by one. 

This might be a bug in estimating the required width of the resulting panel. 

Could you send me your whole neo-tree configuration because I didn't face this issue with my setup?

cseickel:
> I see that the estimated width is over 10,000 but did you purposely put a very very long file name?
> Or maybe expanding a recursive directory symlink might explain this problem?

This is just a standard set of files, nothing particularly long and it shouldn't be opening folders recursively.  Here is my current config: https://github.com/cseickel/dotfiles/blob/e4189bff222470ad07eeb9c947b4cbc4848b6a86/config/nvim/lua/plugins/config/neo-tree.lua#L2-L294

pysan3:
Thanks.

One thing I'm concerned is that the width required is not aligned with the strwidth of the content, and I had to add `1` in order to align with what is shown in the window.
[Here](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/e00391097df82d246c75117cef4d7add864d4eaf/lua/neo-tree/ui/renderer.lua#L346)

Can you guess where this `1` is coming from, and is this actually calculating the width of the window accurately?

pysan3:
> When there is not enough width for the expanded width, it tries to expand more indefinitely falling into an infinite loop which is causing the nvim to freeze.

[...]

PR #627, opened 2022-11-26
fix: auto_expand_width gets caught in infinite loop and hangs neovim (#624)

pysan3:
Should fix https://github.com/nvim-neo-tree/neo-tree.nvim/issues/624.

For more info and discussion, please read the above issue.

### DIFF
--- a/lua/neo-tree/sources/common/container.lua
+++ b/lua/neo-tree/sources/common/container.lua
@@ -187,7 +187,7 @@ local merge_content = function(context)
   -- * Repeat until all layers have been merged.
   -- * Join the left and right tables together and return.
   --
-  local huge_number = 2000
+  local huge_number = vim.o.columns
   local remaining_width = context.auto_expand_width and huge_number or context.container_width
   local left, right = {}, {}
   local left_width, right_width = 0, 0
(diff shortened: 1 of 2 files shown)


cseickel (approved):
Works great for me!

cseickel:
Actually, this is still not quite right. I just got into a situation where I could not scroll vertically in a neo-tree window because every movement caused a re-render and put me back to the top of the window.

I disabled the feature in my own config for now. I think this can use some more work.

pysan3:
@cseickel Thanks for testting.

Could you give me a minimal way to reproduce your issue?

I really cannot have the same behavior with my setup and I have no issues whatsoever.

And could you test what happens when `auto_expand_width = false` and toggle it on with keybind after opening neo-tree.

The rendering should only happen on what was already implementing before adding this feature or when the rendering text length exceeds the width of the window.
I could not think of a reason why this feature interferes with scrolling.

Thanks in advance.

cseickel:
@pysan3 I think this is the core issue: https://github.com/nvim-neo-tree/neo-tree.nvim/blob/4c3d9d55591ffad68518a77447f2ed141913685f/lua/neo-tree/ui/renderer.lua#L946

That creates an infinite loop of renders whenever the auto_expand code is actually triggered.

I ended up rewriting the logic to be more direct and ensure there is only one extra render. I'm also making it something that can be reused for position = `current` and `float`. I'm not quite done with that but the code is in #629.

pysan3:
@cseickel Thank you so so much!!

Your rewrite works like a charm!

PR #637, opened 2022-12-05
feat(git): async git interface for `get_repository_root`

musjj:
This adds a callback argument to `git.utils.get_repository_root`, which allows the function to be called asynchronously.
Callers are also adjusted so that it respects the `async_git_status` option.

Also, I've been wondering, is it really necessary to have synchronous support at all? As long as the asynchronous version is implemented correctly, I can't imagine a user that would prefer to use the synchronous version. 

Once the async paths are battle-tested and its rough edges polished, what do you think about dropping non-async support in the next major version? I feel that it would simplify the codebase.

### DIFF
--- a/lua/neo-tree/git/status.lua
+++ b/lua/neo-tree/git/status.lua
@@ -201,129 +201,136 @@ local function parse_lines_batch(context, job_complete_callback)
 end
 
 M.status_async = function(path, base, opts)
-  local git_root = git_utils.get_repository_root(path)
-  if utils.truthy(git_root) then
-    log.trace("git.status.status_async called")
-  else
-    log.trace("status_async: not a git folder: ", path)
-    return false
-  end
-
-  local event_id = "git_status_" .. git_root
-  local context = {
-    git_root = git_root,
-    git_status = {},
-    exclude_directories = false,
-    lines = {},
-    lines_parsed = 0,
-    batch_size = opts.batch_size or 1000,
-    batch_delay = opts.batch_delay or 10,
-    max_lines = opts.max_lines or 100000,
-  }
-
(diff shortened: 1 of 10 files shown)


cseickel:
> Also, I've been wondering, is it really necessary to have synchronous support at all? As long as the asynchronous version is implemented correctly, I can't imagine a user that would prefer to use the synchronous version.

Believe it or not, some people do prefer sync over async. The reason is that async does effectively make the whole thing slower. Also, some people notice and are really bothered by the flicker that async introduces. What we settled on is that by default, Neo-tree will use sync methods when opening for the first time for that immediate response, then switch to async for any background refreshes.

I think git status is the only thing that is always async by default now, but the git ignored lookup needs to happen before render so that definitely needs both the sync and async code paths.

musjj:
> the flicker that async introduces

I've experienced this too, but it turns out that the flicker is caused by the non-async git calls. With the previous PR and this one, the flicker is now completely gone (at least with my config). As the overall speed, I'm not really sure since measuring it seems tricky.
But I guess you're right, non-async options should be kept for now.

Btw, I'm not so sure what's going on with stylua-check, I've ran the formatter in my editor, so I'm not sure what it's complaining about.

cseickel:
> Btw, I'm not so sure what's going on with stylua-check, I've ran the formatter in my editor, so I'm not sure what it's complaining about.

The problem is the stylua 14.x is incompatible with the behavior of 13.x. When the switch happened, I stuck with 13.x for a while while distributions were being updated with the new version. I think everyone is updated to 14.x now, but this project still needs to be cut over. Just one of a million little things I don't have time to do.

https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/422

Issue #670, opened 2022-12-28
Distinct filter prompt names


tobealive:
Currently, all filter prompts have the same title `"Enter Filter Pattern:"`

Having an unambiguous visual feedback of what command the plugin received and what is going to happen is a good design pattern. Considering this, further improvements could be made with this specific issue.

It could be achieved via renaming the current prompts, e.g.,:
- `filter_on_submit`: "Search..."
- `fuzzy_finder` "Filter..."
- `fuzzy_finder_directory` "Filter Directories..."

An alternative would be to keep the current title and add an icon at the beginning of the title that makes it distinguishable (e.g., a magnifying glass or return symbol for filter_on_submit, a filter icon for fuzzy_finder, a directory icon for fuzzy_finder_directory). This could also be made configurable. 

If it's of any help and time saving for you if such textual changes are submitted as a PR, please let me know.

cseickel:
> If it's of any help and time saving for you if such textual changes are submitted as a PR, please let me know.

PRs are always desirable! Most of my time recently is spent on PR review and bug fixes, I am leaving enhancements to others for now.

tobealive:
@cseickel thanks for the maintenance.

I'll prepare something and submit a PR.
If you have a preference for one of the suggestions, please let me know.

cseickel:
Sounds good. I would prefer changing the text of the prompt. You never know if someone will have a nerd font installed so I wouldn't want to be dependent on that.

lougreenwood:
Thankyou for fixing this @tobealive!

I'm running an older version of neotree and was really confused why pressing `return` on a `/` filter was opening the file, searched around and found this issue which describes the problem perfectly. Looking forward to when my nvim bistro (`astronvim`) updates so I can get the proper labels too 😄

PR #671, opened 2022-12-28
feat: use distinct filter prompt titles

tobealive:
#670 

The titles are inspired by modern file mangers like dolphin. Which use "Filter" for live filtering. "Search" for seraching on submit.

Screenshot_20221228_231549
Screenshot_20221228_231628

Please let me know if any changes are desired or feel free to submit as co-auth.


cseickel (approved):
Looks good. Thanks!

tobealive:
Thanks for merging!

What I see now is that I probably would have named the variable `popup_title` if I hadn't been working with messages prior to this pr, and if the day hadn't already ended for my brain. But such semantic bean counting can be done at any time.

Issue #672, opened 2022-12-29
Navigating the Preview Window


tobealive:
As a user, I would like to be able to scroll in a preview window when the contents of a buffer displayed in the preview exceed the height of the preview window.

This could be achieved by extending the command list with scroll-commands (those could behave similar to what Telescope does by default with `<C-d>` `<C-u>`) or by adding the ability to move into the preview window (similar to what many "hover" actions do when double pressing the hover key).

cseickel:
Personally, I would prefer to move the cursor into the preview window so I could use the full range of movement and search commands. 

From the neo-tree side it would be simple to add a command that would focus the preview window, which can be mapped to whatever the user wants. I'm not sure if the window might close on you the way it is written now, there may be a small amount of work there to prevent that.

Once you are in the preview window, it would be nice to add just two extra mappings:

- One to close the window and end preview mode
- One to move the cursor back to neo-tree

I'm not sure if there should be extra config options for the preview window mappings or if it would be better to just read the mappings on the Neo-tree window and reuse whatever is mapped to `close_window` and `focus_preview`.

There's also nothing stopping us from also adding scroll commands that work without focusing the window. We can do either one or both.

ajitid:
> There's also nothing stopping us from also adding scroll commands that work without focusing the window. We can do either one or both.

I think it'd be better to only provide controls for scroll and not go for full-fledged ones. Otherwise, we'd need to wrestle with variables to ensure the preview window remains read-only. One can always open the file to do search and other operations, and this also aligns with the previewers that [fzf-lua](https://github.com/ibhagwan/fzf-lua) and [vim-bqf](https://github.com/kevinhwang91/nvim-bqf) provide.

tobealive:
@ajitid sounds reasonable, if I can grab a free moment later today, I'm starting with the implementation. Otherwise, I can schedule a few minutes every day in the coming week :+1:

tobealive:
In the end, I started with the approach preferred by @cseickel. There are two reasons for this:

- I can certainly see myself moving into the preview window and yank some content, as I often do this with lsp hover dialogs.
- Time constraints, it was easier to implement a focus command.
- On second thought, I don't think the preview window needs to be read-only.

We can extend navigation in the preview window incrementally. A command to scroll in the preview while the tree is focused can still be implemented.

dudicoco:
[...]

PR #675, opened 2023-01-02
feat: add focus_preview command

tobealive:
#672 

>Once you are in the preview window, it would be nice to add just two extra mappings:
>
>- One to close the window and end preview mode
>- One to move the cursor back to neo-tree

I didn't implement such additional mappings. 

- Moving the cursor back to the neo-tree: I don't think an additional mapping is necessary as this can be done with the present `:Neotree` commands and with vims window movement keys (`<C-w>h` `<C-w>` and the user-defined equivalents to it) etc..

- Close preview: I was not 100% sure about this one. But I also left it out for now, since the window can be closed with the standard commands for closing windows (i.e. `:q`). Users can also assign buffer local commands to close popup windows. But I can certainly see a use case for implementing this to make it specific to the preview. Please let me know what you think about how it could be implemented (Should it be another separate command to close the focused preview window?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,7 @@ use {
             ["<cr>"] = "open",
             ["<esc>"] = "revert_preview",
             ["P"] = { "toggle_preview", config = { use_float = true } },
+            ["gp"] = "focus_preview"
             ["S"] = "open_split",
             ["s"] = "open_vsplit",
             -- ["S"] = "split_with_window_picker",
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ use {
             ["<cr>"] = "open",
             ["<esc>"] = "revert_preview",
             ["P"] = { "toggle_preview", config = { use_float = true } },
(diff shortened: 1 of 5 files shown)


cseickel:
I tried it out and it works fine. The only question I'll raise is if `gp` is the best choice for the default mapping. The problem is that it conflicts with `git_push` in the git status source, and some people (including me) apply those git mappings as global mappings for all sources.

I personally mapped the `focus_preview` command to `l` in my config which works great for me because I actually use `h` `j` `k` `l` for window navigation. I'm probably unusual in that, but `l` certainly has a certain logic to it. What do you think? If not `l` then I would choose some other mapping that is not used in any sources.

tobealive:
For me, "l" is perfectly fine as the default. - Sending this from my phone. I'll push the change when I get to my workplace tomorrow 🙂👍

cseickel (approved):

PR #687, opened 2023-01-10
fix(queue): don't go to next node if it has been removed

mrbjarksen:
If an event handler should be removed (by way of `once = true` or the handler returning `true`), the event queue erroneously skips handling the next node and jumps two nodes ahead. If this is the last event in the queue, this also results in an error (attempt to index a nil value).


cseickel (approved):
Looks good, thanks!

cseickel:
That must have been a difficult bug to track down which led you to this fix... Sorry about that!

PR #721, opened 2023-01-28
feat(fuzzy): add a proper fuzzy search algorithm

pysan3:
Hi, it's me again :)

Note: this PR is in draft due to lack of docs. But I first wanted to ask for feedbacks. This PR is (~~if no bug~~) done code wise. 

## `fuzzy_finder` is actually not fuzzy searching.

Although we have a `fuzzy_finder` command, mapped to `/` in filesystem source by default, my understanding of fuzzy is not what is currently implemented.

Suppose we want to match the path `lua/neo-tree/sources/filesystem/commands.lua`, the current behavior basically does a regex search with spaces substituted with `.*`, which does not match:

- `commands filesystem`
  - when the order is reversed
- `flsys cmds`
  - when the names are abbreviated

In my opinion (and due to the popularity of [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim)), I think most of neovim users think that the term `fuzzy` corresponds to the algorithm used in programs such as [`fzf`](https://github.com/junegunn/fzf), [`fzy`](https://github.com/jhawthorn/fzy) which successfully matches the examples above.

## Feature Request

TLDR; workaround [this](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8238865e1d9c61f1a260c290653f2c419503e0a9/lua/neo-tree/sources/filesystem/lib/filter_external.lua#L113)

The code already tries to use the `fzf` command line tool [here](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8238865e1d9c61f1a260c290653f2c419503e0a9/lua/neo-tree/sources/filesystem/lib/filter_external.lua#L113) but it fails to run since fzf requires an interactive shell to work properly, as mentioned in [this issue](https://github.com/nvim-lua/plenary.nvim/issues/14).
One solution would be to launch fzf inside an interactive shell but that would be a massive overhead for each fuzzy search (per keypress!).

Therefore, I implemented the [fzy algorithm](https://github.com/jhawthorn/fzy/blob/master/ALGORITHM.md) with lua (mainly in [this file](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/81f5e1984caefe54903ca1b168a0431325ae76b7/lua/neo-tree/sources/filesystem/lib/filter_fzy.lua): BIG CREDITS TO [fzy-lua](https://github.com/swarn/fzy-lua)) which can be used to filter and sort the files based on the search query.

## Behavior

https://user-images.githubusercontent.com/41065736/215294405-4a6e6742-ebee-492c-b017-c39a26cfc82a.mp4

- **NOTE**: ` - 0.00pts` after the filenames is the score of each path and is ONLY shown for demonstration for this PR. This PR does not contain the code to show the points. In order to mimic the video, please add the following code.

```diff
Modified lua/neo-tree/ui/renderer.lua
@@ -228,6 +228,11 @@ create_nodes = function(source_items, state, level)
     }
     local indent = (state.renderers[item.type] or {}).indent_size or 4
 
+    -- TODO: Only for debug. Remove
+    if state.fuzzy_finder_mode ~= nil then
+      nodeData.name = string.format([[%s - %spts]], nodeData.name, state.fzy_sort_result_scores[nodeData.path])
+    end
+
     local node_children = nil
     if item.children ~= nil then
       node_children = create_nodes(item.children, state, level + 1)
```

## Discussions

1. I have taken the fzy algorithm code mostly from [fzy-lua](https://github.com/swarn/fzy-lua). [Their license](https://github.com/swarn/fzy-lua/blob/main/LICENSE) is MIT and I have added credit on top of the file [`filter_fzy.lua`](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/81f5e1984caefe54903ca1b168a0431325ae76b7/lua/neo-tree/sources/filesystem/lib/filter_fzy.lua). Is this ok for you @cseickel?
2. I did not expose any _commands_ to use this feature yet. It is now specified through `state.find_command == "fzy-lua"`. I would like to ask your opinion on the command name and whether this should be the default behavior of `fuzzy_finder`.
3. TODO: docs and readme. Therefore, this PR is draft.

## Usage

I would be more than happy if someone could test my implementation. (Especially on Windows. I can only test on Linux...)

If you are using lazy.nvim,

```lua
return {
  "pysan3/neo-tree.nvim",
  branch = "feat-fzy-finder", -- BRANCH NAME
  version = false,
  dependencies = {
    { "MunifTanjim/nui.nvim" },
    { "kyazdani42/nvim-web-devicons" },
  },
  cmd = { "Neotree" },
  config = {
    filesystem = {
      window = {
        mappings = {
          ["/"] = "modified_filter", -- add a keybind
        },
      },
      commands = {
        modified_filter = function(state)
          local filter = require("neo-tree.sources.filesystem.lib.filter")
          state.find_command = "fzy-lua" -- specify the algorithm here
          filter.show_filter(state, true, true) -- same as command "fuzzy_finder"
        end,
      },
    },
  }
}
```

---

I am looking forward to any feedbacks ;)
pysan3


pysan3:
Oh I didn't notice the [state of the plugin](https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/704). 

I would love to help but I unfortunately haven't got the time at least this year. I'd be more than happy to take the role from next year (if that is an option). 

Back to this PR, could you @cseickel tell me how feature requests are considered for now?
I think it would be nice to make this PR the first example dealt by people commented over at the discussion?

I can't wait to see this plugin evolve even more towards the future and I'd love to hear you opinion. 

Best,

cseickel:
> Oh I didn't notice the [state of the plugin](https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/704).

The one thing I am most reserving my available time for is PR review. That's really the priority for me. Even if we get a volunteer to be the new lead, it would take time for me to become comfortable that they will manage the project the way I'd want it to be managed, so I will keep the ship on course for now.

> Back to this PR, could you @cseickel tell me how feature requests are considered for now?

For now, I will continue as before.


## My comments on this PR:

I definitely would love a better fuzzy search, and I especially like the fzy algorithm. The problem we have to overcome is performance. Recursively scanning the filesystem with plenary is just too slow for any reasonably large folder, so you would need a different source to feed into the fzy scoring system. 

Of course, I did try it out and it works remarkably well in a smaller directory. I just worry about how it will scale. I would suggest that the first stage come from the existing fd/find route but with a search pattern that will return more results. It can even go so far as to just search for a regex that would "must contain at least one of each of the non-space characters typed into the box." As a test input, I would use some larger mono-repo. I always go with the linux repo for stress testing this sort of thing.

With regards to licensing, I think it must be best to include the fzy project as a dependency rather than copying the file wholesale. If you did copy the file, I would copy the contents of  the entire license file to the top of the file you are copying as a block comment.

As far as the command, something this significant will definitely have a period of bugs to shake out, and should therefore be opt-in. Maybe it becomes the default in 3.0. I would either make it a separate command, or add a config option to the existing mapping to enable the `fzy` matching method.

miversen33:
> I would suggest that the first stage come from the existing fd/find route but with a search pattern that will return more results

I cant imagine that reaching out over shell to either find or fd would be faster than using the in built c binding to `luv`. I think the biggest hurdle is just going to be generically "Big filesystems suck". If you run `find` from your shell over a large file system, it also sucks lol. `fd` Might be better, I haven't played much with it. But I still don't think it will perform better for simple tree walking than staying in lua/c and avoiding stream processing.

miversen33:
Welp, I threw together a test (you will need [plenary](https://github.com/nvim-lua/plenary.nvim), as well as my [spawn wrapper](https://github.com/miversen33/netman.nvim/blob/b65f54cbd5f2b52c83cfd4509925b993c91160f4/lua/netman/tools/shell.lua) if you want to run it yourself). I have it scanning my `usr` directory which apparently contains about 620k items (not huge, but large enough that the tests don't finish immediately)

Code
```lua
local shell = require("netman.tools.shell")
local luv = require("luv")
local plenary = require("plenary.scandir")
local start_dir = '/usr/'

local paths = { {path=start_dir} }

local count = 0
-- I don't know if defining these outside the
-- while loop actually helps performance, but it shouldn't
-- whereas its possible that continually redefining
-- these in the loop might get memory painful overtime
local head = nil
local head_path = nil
local head_type = nil
local head_stat = nil
local head_fd = nil
local child_info = nil
local test1 = function()
    local duration = luv.hrtime()
    while #paths > 0 do
        count = count + 1
        -- Pops the head off the paths table. At start, this causes the table to be empty
        head = table.remove(paths, 1)
        head_path = head.path
        if not head.type then
            head_stat = luv.fs_stat(head_path)
            if(head_stat) then
                head_type = head_stat.type
            else
                -- Some how we didn't get any stat results for head. Head probably doesn't exist.
                goto continue
            end
        else
            head_type = head.type
        end
        if head_type ~= 'directory' then
            -- Don't need to walk head_type, still should store its info though?
            goto continue
        end
        head_fd = luv.fs_scandir(head_path)
        if(head_fd) then
            while true do
        --         -- Iterate through each item returned by luv.fs_scandir_next until there is nothing left to iterate over
        --         -- add each item to the paths if its a directory only
                child_info = {luv.fs_scandir_next(head_fd)}
                if not child_info or #child_info == 0 then
                    break
                end
                count = count + 1
                if child_info[2] == 'directory' then
                    table.insert(paths, {path=string.format("%s/%s", head_path, child_info[1]), type='directory'})
                end
            end
        end
        ::continue::
    end
    duration = (luv.hrtime() - duration) / 1000000
    return duration
end

local test2 = function()
    local duration = luv.hrtime()
    plenary.scan_dir_async(start_dir, {
        hidden = true
    })
    duration = (luv.hrtime() - duration) / 1000000
    return duration
end

local test3 = function()
    local duration = luv.hrtime()
    shell:new({"/usr/bin/find", "/opt/"})
    duration = (luv.hrtime() - duration) / 1000000
    return duration
end

local test4 = function()
    local duration = luv.hrtime()
    shell:new({"/usr/bin/fd", ".", "/opt/"})
    duration = (luv.hrtime() - duration) / 1000000
    return duration
end

print(string.format("Scanning %s", start_dir))
print("Doing Sync Libuv test")
local test1_time = test1()
print("Doing Async Plenary Test")
local test2_time = test2()
print("Doing Spawn Test with find")
local test3_time = test3()
print("Doing Spawn Test with fd")
local test4_time = test4()
print(string.format("    Test 1 (libuv sync scan) took approximately %s ms to scan %s items", test1_time, count))
print(string.format("    Test 2 (plenary async scan) took approximately %s ms to scan %s items", test2_time, count))
print(string.format("    Test 3 (spawn find scan) took approximately %s ms to scan %s items", test3_time, count))
print(string.format("    Test 4 (spawn fd scan) took approximately %s ms to scan %s items", test4_time, count))

```

Results
```
Scanning /usr/
Doing Sync Libuv test
Doing Async Plenary Test
Doing Spawn Test with find
Doing Spawn Test with fd
    Test 1 (libuv sync scan) took approximately 528.469989 ms to scan 619357 items
    Test 2 (plenary async scan) took approximately 0.158897 ms to scan 619357 items
    Test 3 (spawn find scan) took approximately 0.06071 ms to scan 619357 items
    Test 4 (spawn fd scan) took approximately 0.004345 ms to scan 619357 items
```

So I was incorrect. Running with find and fd (even through a shell) **is** faster. Substantially when comparing `fd` to plenary's async.

cseickel:
Thanks for running those benchmarks @miversen33, it's definitely good to know those results.

> I think preferring a cli tool (such as find or fd) is fine, though we need to also consider windows (which does not have native find). Either including a way to reach into a windows tool that is "comparable" to cli (I don't know windows well enough to know what this is) or simply fall back to plenary if find/fd is not present on the host system.

Windows does have some utilities, and the default external filter code will actually find the best available even on Windows and use it. I think that on windows, the cost of the plenary scan is significantly worse than on linux or mac, so it's even more important to use the `find` or `where` utility on Windows.

miversen33:
Those results didn't make sense to me. It turns out I forgot to run the shell command in the tests (notice the `shell:new` but no `:run` after :upside_down_face: )

So my results above are garbage.
To sanity check this, I removed my spawn wrapper and used the job functions provided by plenary.

Below are the actual results (along with updated code)
```lua
local luv = require("luv")
local plenary = require("plenary.scandir")
local job = require("plenary.job")
local start_dir = '/usr/'

local paths = { {path=start_dir} }

local count = 0
-- I don't know if defining these outside the
-- while loop actually helps performance, but it shouldn't
-- whereas its possible that continually redefining
-- these in the loop might get memory painful overtime
local head = nil
local head_path = nil
local head_type = nil
local head_stat = nil
local head_fd = nil
local child_info = nil
local test1 = function()
    local duration = luv.hrtime()
    while #paths > 0 do
        count = count + 1
        -- Pops the head off the paths table. At start, this causes the table to be empty
        head = table.remove(paths, 1)
        head_path = head.path
        if not head.type then
            head_stat = luv.fs_stat(head_path)
            if(head_stat) then
                head_type = head_stat.type
            else
                -- Some how we didn't get any stat results for head. Head probably doesn't exist.
                goto continue
            end
        else
            head_type = head.type
        end
        if head_type ~= 'directory' then
            -- Don't need to walk head_type, still should store its info though?
            goto continue
        end
        head_fd = luv.fs_scandir(head_path)
        if(head_fd) then
            while true do
        --         -- Iterate through each item returned by luv.fs_scandir_next until there is nothing left to iterate over
        --         -- add each item to the paths if its a directory only
                child_info = {luv.fs_scandir_next(head_fd)}
                if not child_info or #child_info == 0 then
                    break
                end
                count = count + 1
                if child_info[2] == 'directory' then
                    table.insert(paths, {path=string.format("%s/%s", head_path, child_info[1]), type='directory'})
                end
            end
        end
        ::continue::
    end
    duration = (luv.hrtime() - duration) / 1000000
    return duration
end

local test2 = function()
    local duration = luv.hrtime()
    plenary.scan_dir_async(start_dir, {
        hidden = true
    })
    duration = (luv.hrtime() - duration) / 1000000
    return duration
end

local test3 = function()
    local duration = luv.hrtime()
    job:new({
        command = "/usr/bin/find",
        args = {"/opt"}
    }):sync()
    duration = (luv.hrtime() - duration) / 1000000
    return duration
end

local test4 = function()
    local duration = luv.hrtime()
    job:new({
        command = "/usr/bin/fd",
        args = {".", "/opt/"}
    }):sync()
    duration = (luv.hrtime() - duration) / 1000000
    return duration
end

print(string.format("Scanning %s", start_dir))
print("Doing Sync Libuv test")
local test1_time = test1()
print("Doing Async Plenary Test")
local test2_time = test2()
print("Doing Spawn Test with find")
local test3_time = test3()
print("Doing Spawn Test with fd")
local test4_time = test4()
print(string.format("    Test 1 (libuv sync scan) took approximately %s ms to scan %s items", test1_time, count))
print(string.format("    Test 2 (plenary async scan) took approximately %s ms to scan %s items", test2_time, count))
print(string.format("    Test 3 (plenary find scan) took approximately %s ms to scan %s items", test3_time, count))
print(string.format("    Test 4 (plenary fd scan) took approximately %s ms to scan %s items", test4_time, count))
```

```
Scanning /usr/
Doing Sync Libuv test
Doing Async Plenary Test
Doing Spawn Test with find
Doing Spawn Test with fd
    Test 1 (libuv sync scan) took approximately 580.085922 ms to scan 619357 items
    Test 2 (plenary async scan) took approximately 0.012304 ms to scan 619357 items
    Test 3 (plenary find scan) took approximately 118.025346 ms to scan 619357 items
    Test 4 (plenary fd scan) took approximately 96.017473 ms to scan 619357 items
```

The long short, my initial gut reaction was correct, shelling out is far more expensive than simply using plenary's async scan. I just don't know how to read my own code I guess lol, and apparently needed to lay in bed for a while before realizing the error in my test.

Tagging @cseickel

cseickel:
>    Test 1 (libuv sync scan) took approximately 580.085922 ms to scan 619357 items
>    Test 2 (plenary async scan) took approximately 0.012304 ms to scan 619357 items
>    Test 3 (plenary find scan) took approximately 118.025346 ms to scan 619357 items
>    Test 4 (plenary fd scan) took approximately 96.017473 ms to scan 619357 items

I think you made another mistake there. If you thik about it, async should always take a little longer than sync because of the extra overhead. Parallel is faster than either, but I don't think Plenary's async is parallel. Either way there is no way that it would be so much faster.

I think what you are timing here is the time it takes to declare and **start** the async scan. There is no callback to report the time after the async code actually completes.

```lua
local test2 = function()
    local duration = luv.hrtime()
    plenary.scan_dir_async(start_dir, {
        hidden = true
    })
    -- The scan has not actually run yet...
    duration = (luv.hrtime() - duration) / 1000000
    return duration
end
```

As an aside, our async scan is actually parallel in that it fires off an async scan for each sub-folder in the path we are scanning. It is more efficient than plenary in general but till nowhere near fd speeds.

One last point, is that the `fd` is used to **search**, not just list all of the files. It should be returning a filtered list as the results of some search term. It also is able to do so while omitting gitignored items if that's what your settings specify, which is an incredibly expensive operation that plenary is incapable of doing correctly.

pysan3:
Thank you for the explanation and taking your time to review this PR.

> The problem we have to overcome is performance. Recursively scanning the filesystem with plenary is just too slow for any reasonably large folder,

I am also aware of this problem, and I do see issues over on plenary side about the lack of speed.

However, my take is that exposing a _real_ fuzzy finder at least as an option would be better than having none.
And since the size of the repo, as well as the PC spec varies from user to user, I think we can hand it to the user to decide which algorithm should be used with a big caution in the documentation.

> With regards to licensing, I think it would be best to include the fzy project as a dependency rather than copying the file wholesale. If you did copy the file, I would copy the contents of the entire license file to the top of the file you are copying as a block comment.

fzy-lua is not a neovim plugin and does not have the `repo/lua/<repo-name>` structure so I am not sure if plugin managers can handle it correctly. As well as how should I `require` it?
I think packer can handle these via luarock but I wasn't sure lazy had any option.

I took the copy-paste route and have added the whole license as a block comment. Could you check if this is OK?
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/cdc87151ab4d4d8ef6545e6200c22aac58e04c34/lua/neo-tree/sources/filesystem/lib/filter_fzy.lua

---

Thank you so much @miversen33 for the detailed tests but I found some mistakes here and there, listed below. But your code lead me to a very efficient code that can compete against the other options.

My opinion when I was implementing this was that since the search feature is not something that is always running, I thought performance would not really be something to really care.

However, seeing all the analysis (below), I think it is worth rewriting it to use the outputs of shell commands for the case of `fd`. And since there is already a very well written code to invoke the shell commands [here](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8238865e1d9c61f1a260c290653f2c419503e0a9/lua/neo-tree/sources/filesystem/lib/filter_external.lua#L46), I am thinking of refactoring this part a bit so that I can reuse the mechanism for the input of fzy. But lua actually did surprisingly well.

Besides, it seems to be that the telescope guys are taking a [pretty similar approach](https://github.com/nvim-telescope/telescope.nvim/blob/203bf5609137600d73e8ed82703d6b0e320a5f36/lua/telescope/builtin/__files.lua#L215), so ~~although tinkering with lua was quite fun~~ I think this is the right path to take.

---

I am not bright in time scanning but when I printed out duration inside `on_exit`, `plenary.scan_dir_async` seems to take like a huge amount of time. (around `x100` of `find`)
But when I switched to `plenary.scan_dir` (without async), it actually managed to compete against the shell commands pretty well.
Moreover, I tweaked the code written by @miversen33 and was able to perform better than `find`. Couldn't beat `fd` tho cuz it seems to parallel the queue (which is noted in `-j` option in `fd --help`)

Also, the analysis done by @miversen33 was scanning `/opt` for `fd` and `find` which should be a mistake I guess? And the options for each command were not precisely aligned with what was implemented in `test1` like follow link and ignore hidden files.

- `test1`: Changed to only count the number of `"file"`s
- `test2`: ????
- `test3`: `find /usr -type f`
- `test4`: `fd . /usr -t=f -HI --color=never`

Here are the final results. I can assure the programs are doing the same things (except for test2) because the number of files are exactly the same. (I did not get output from `plenary.scan_dir`. I don't know why...)

```txt
Scanning /usr/
Doing Sync Libuv test
Doing Sync Plenary Test
Doing Spawn Test with find
Doing Spawn Test with fd
    Test 1 (libuv sync scan) took approximately 966.847855 ms -- No warmup will slow it down to like 3000 ms
    Test 2 (plenary sync scan) took approximately 1229.072352 ms
    Test 3 (plenary find scan) took approximately 1246.537361 ms
    Test 4 (plenary fd scan) took approximately 836.98088 ms
luv.fs_scandir: done, count: 429128
find: done, count: 429128
fd: done, count: 429128
```

Code:

```lua
local luv = require("luv")
local plenary = require("plenary.scandir")
local job = require("plenary.job")
local start_dir = '/usr/'

local test1_count = 0


local test1 = function()
  local stack_index = 1
  local stack_path = { start_dir }
  local type_hash = {
    file = 1,
    directory = 2,
  }
  local stack_type = { type_hash["directory"] }
  local count = 0
  local duration = luv.hrtime()
  while stack_index > 0 do
    -- array[#array] access is faster. So using stack instead of queue
    local head_path = stack_path[stack_index]
    local head_type = stack_type[stack_index]
    stack_index = stack_index - 1
    if head_type == 1 then -- "file"
      count = count + 1
    elseif head_type == 2 then -- "directory"
      local head_fd = luv.fs_scandir(head_path)
      if (head_fd) then
        while true do
          local name, type = luv.fs_scandir_next(head_fd)
          if not type then
            break
          end
          stack_index = stack_index + 1
          stack_path[stack_index] = string.format([[%s/%s]], head_path, name)
          stack_type[stack_index] = type_hash[type]
        end
      end
    else
      -- symlinks etc are ignored
    end
  end
  duration = (luv.hrtime() - duration) / 1000000
  test1_count = count
  return duration
end

local test2 = function()
  local duration = luv.hrtime()
  plenary.scan_dir(start_dir, {
    hidden = true,
    on_exit = function(result)
      print(string.format([[plenary.scan_dir: done, count: %s]], #result))
    end,
  })
  duration = (luv.hrtime() - duration) / 1000000
  return duration
end

local function count_output_lines(name)
  return function(j, _)
    print(string.format([[%s: done, count: %s]], name, #j:result()))
  end
end

local test3 = function()
  local duration = luv.hrtime()
  job:new({
    command = "/usr/bin/find",
    args = { start_dir, "-type", "f" },
    on_exit = count_output_lines("find"),
  }):sync()
  duration = (luv.hrtime() - duration) / 1000000
  return duration
end

local test4 = function()
  local duration = luv.hrtime()
  job:new({
    command = "/usr/bin/fd",
    args = { ".", start_dir, "-t=f", "-HI", "--color=never" },
    on_exit = count_output_lines("fd"),
  }):sync()
  duration = (luv.hrtime() - duration) / 1000000
  return duration
end

-- warmup
test1()
test2()
test3()
test4()
test1()
test2()
test3()
test4()

print(string.format("Scanning %s", start_dir))
print("Doing Sync Libuv test")
local test1_time = test1()
print("Doing Sync Plenary Test")
local test2_time = test2()
print("Doing Spawn Test with find")
local test3_time = test3()
print("Doing Spawn Test with fd")
local test4_time = test4()
print(string.format("    Test 1 (libuv sync scan) took approximately %s ms", test1_time))
print(string.format("    Test 2 (plenary sync scan) took approximately %s ms", test2_time))
print(string.format("    Test 3 (plenary find scan) took approximately %s ms", test3_time))
print(string.format("    Test 4 (plenary fd scan) took approximately %s ms", test4_time))

print(string.format([[luv.fs_scandir: done, count: %s]], test1_count))

--[[
Scanning /usr/
Doing Sync Libuv test
Doing Sync Plenary Test
Doing Spawn Test with find
Doing Spawn Test with fd
    Test 1 (libuv sync scan) took approximately 966.847855 ms to scan 0 items
    Test 2 (plenary sync scan) took approximately 1229.072352 ms to scan 0 items
    Test 3 (plenary find scan) took approximately 1246.537361 ms to scan 0 items
    Test 4 (plenary fd scan) took approximately 836.98088 ms to scan 0 items
luv.fs_scandir: done, count: 429128
find: done, count: 429128
fd: done, count: 429128
--]]
```

miversen33:
@pysan3 
Ya I was noticing some more issues with my tests this morning but got a bit sidetracked with work. I planned on revisiting them afterwards, but it looks like I dont have to!

My main concern is just ensuring that large scans are thought through thoroughly (boy that's a tongue twister). As @cseickel pointed out, my plenary async scan test is inaccurate (I couldn't get it to cooperate this morning when toying with it a bit either), so I am glad someone was able to properly get that tested out.

It does seem like reaching out to an external binary (such as find/fd/where) is the way to go here and nice find on how telescope handles this! 

Just for clarity sake, could you elaborate on the `warmup` section of your above test? I can't imagine that is something that should be done in "production" code (running your search twice seems counterintuitive from an end user standpoint).

Also, I am not seeing in this commit, a way to cancel the fuzzy find command should the end user want to. Example, I execute a fuzzy find over my entire root directory, looking for anything containing the letter `e`. That would take forever, do we have a mechanic to "cancel" that find right now? I know underlying you can do this with `spawn`, and since we have plenary, I _believe_ plenary jobs have wrapped this "cancel" ability as well.

I do agree that a true fuzzy finder is a fantastic idea :) But I don't know that I like the idea of "let the user figure out if it works for them or not". We should try to ensure that it "works" for your average user, even if it is an optional command. Otherwise it won't get used.

pysan3:
> Just for clarity sake, could you elaborate on the warmup section of your above test?

I was kinda just having fun with lua to see what happens when I push it to its limit. And because SSD and OS have caching mechanisms, it would not measure the time accurately enough without a warmup.

The decrease in time in the lua case comes from (in my opinion) being literally test 1 and not test 4, and due to the jit compiler. I haven't gone much into this part but in my understanding, lua code should execute faster from the second run.

I was not planning to put this in the neo-tree code but just for the sake of fair comparison.

> a way to cancel the fuzzy find command should the end user want to.

I see that. Yes, I believe we have no way to stop the computation except from killing neovim entirely and it hangs entire neovim which is indeed a problem. Neovim devs seem to be working on this but I'm not sure about the progress.

I guess we could add a code to simply error out when a certain time has elapsed (which can be configured) to avoid a sudden hang but we should ask @cseickel if that is a good idea.

Or we can wrap the whole function with `luv.new_thread()` or `luv.new_async()` I guess? Unfortunately I don't have much knowledge here.

miversen33:
With `luv.spawn`, we can do timeout handling as well as cancellation handling without needing to wait on [the async issue to become resolved and mainstream](https://github.com/neovim/neovim/issues/19624#issuecomment-1407146894).

I have an example of how to do this with `luv.spawn` in netman's shell wrapper, found [here](https://github.com/miversen33/netman.nvim/blob/b65f54cbd5f2b52c83cfd4509925b993c91160f4/lua/netman/tools/shell.lua#L298-L329)

I will research this evening, I am almost certain that Plenary jobs have a similar mechanic which would be great as Neo-tree is already using plenary. But if worst comes to worst, it is possible to do this via the low level `luv.spawn`

pysan3:
Thank you very much for your information. I haven't dealt with coroutines and stuffs in neovim so this is very helpful.

> I will research this evening,

Are you planning on extending the `fd` / `find`? If so, it would be very nice if you could also wrap [the old search function](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8238865e1d9c61f1a260c290653f2c419503e0a9/lua/neo-tree/sources/filesystem/lib/filter_external.lua#L46) since I think it does not have a safety mechanism as well.

BTW, did you test my `test1` on your machine? It seems to be (at least on my end) that pure lua implementation performs the fastest when it comes to smaller repos. (I mean even with the whole `$HOME` dir).

If this is the same in your case, I believe the first priority should be on smaller repos, hence not using shell commands but lua with a safety mechanism.

Anyways I would like to ask @cseickel on which route we should take.

miversen33:
Some quick results on my machine with your test @pysan3 
```
Scanning /usr/
Doing Sync Libuv test
Doing Sync Plenary Test
Doing Spawn Test with find
Doing Spawn Test with fd
    Test 1 (libuv sync scan) took approximately 405.157384 ms
    Test 2 (plenary sync scan) took approximately 704.815711 ms
    Test 3 (plenary find scan) took approximately 270.366907 ms
    Test 4 (plenary fd scan) took approximately 395.056319 ms
luv.fs_scandir: done, count: 511415
```

I find it very interesting that find is significantly faster on than anything else on my machine, but it is the slowest on yours. There's quite the discrepancy in results here :/

> Are you planning on extending the fd / find? If so, it would be very nice if you could also wrap [the current search function](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8238865e1d9c61f1a260c290653f2c419503e0a9/lua/neo-tree/sources/filesystem/lib/filter_external.lua#L46) since I think it does not have a safety mechanism as well.

I had not planned on overhauling or extending anything just yet, but rather to research how to cancel a plenary job. It looks like it _should_ be as simple as calling `:stop` on the job handle provided (per [the job doc](https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/job.lua#L194-L200)).

That means the bigger questions that need answered (imo) are

- Do we prefer shelling out or refining the lua route?
- How do we present a "cancel" option to the end user when they execute a command. Is this functionality already in neo-tree and I just missed it?

pysan3:
Thank you @miversen33 for doing the tests.

It's quite interesting that we all have a very different results and I'm not sure what is the correct way to take.

But I think we can abstract the [find_files](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8238865e1d9c61f1a260c290653f2c419503e0a9/lua/neo-tree/sources/filesystem/lib/filter_external.lua#L46) function to utilize all methods and even expose a config option to the user.

>  It looks like it should be as simple as calling `:stop` on the job handle provided

Thank you so much. Yes this seems to be the answer. 

> How do we present a "cancel" option to the end user when they execute a command. Is this functionality already in neo-tree and I just missed it?

I believe on a nui input, user can press `<Esc>` to abort but I don't think there is any mechanism to cancel any command execution but I'm not exactly sure.

In case of plenary.job, I think this would work but this will overwrite the keybind, and I doubt this is a clean solution.

```lua
local Job = require("plenary.job")

local myjob = Job:new({
  command = "find",
  args = { "/usr", "-type", "f" },
  on_stdout = function(err, line)
    vim.pretty_print([[stdout:]], line)
  end,
  on_stderr = function(err, line)
    vim.pretty_print([[stderr:]], line)
  end,
  on_exit = function(code, signal)
    vim.pretty_print([[CODE: exit:]], code)
    vim.pretty_print([[SIGNAL: exit:]], signal)
  end,
})

myjob:start()

vim.keymap.set("n", "<Esc>", function()
  myjob:_stop()
end, {
  buffer = ...,
})
```

> Do we prefer shelling out or refining the lua route?

I am thinking of implementing all of them because either way has its benefits and cons.

- pure lua
  - should be the fastest for small dirs
  - pretty tedious to implement gitignore and other filters
- plenary.scandir
  - supports gitignore and some filters (imcompletely)
  - not so fast
- find, fd, ...
  - has rich filters
  - we can copy from existing code and telescope
  - less performance decrease on larger repos
  - overhead of spawning a shell, especially in smaller repos


But I would also like to hear from @cseickel for his opinion.

cseickel:
> I am thinking of implementing all of them because either way has its benefits and cons.

Actually, I don't think there are actually any benefits to the libuv and plenary options. One big driver of that is that they only seem viable if you don't bother with handling gitignore. Plenary's gitignore is way to naive and just doesn't work for anything but the simplest case. We have implemented it ourselves by calling git directly to get the ignored files per folder. It's really the only reasonable way to get this right in lua.

`fd` and `find` will do this for you, and will do it correctly and insanely fast. If you combine that with the fact that even just recursively listing all files is fast before doing any filtering, it just doesn't make sense to bother with a hand made solution. In reality you should be using fd/find to also do a first stage filter so that the results you need to iterate over and score are reduced as well, which is a further speed up.


> 
> * pure lua
>   
>   * should be the fastest for small dirs
>   * pretty tedious to implement gitignore and other filters

Why would it be fastest for small directories? Besides, who cares what is fastest for small directories? Small directories are never a problem so no need to make any decisions based on what is best for them. You need to optimize for the slow cases only.

> * plenary.scandir
>   
>   * supports gitignore and some filters (imcompletely)
>   * not so fast

Plenary is a great library for a lot of things, just not for scanning a file system. `Plenary.scandir` is banned from this project. It's slow and inaccurate.

> * find, fd, ...
>   
>   * has rich filters
>   * we can copy from existing code and telescope
>   * less performance decrease on larger repos
>   * overhead of spawning a shell, especially in smaller repos

I don't think there is actually any meaningful overhead for calling the external process. You don't actually spawn a shell, at least not in the way that we call it.

I would reuse the existing code in `filter_external.lua` and provide a way to adjust the filter to cast a super wide net that is suitable for feeding to the fzy methods, then inject that fzy code as an extra intermediate step to what is already there.

pysan3:
Thanks for your comment.

Ok, so I will

- refactor `filter_external.lua` to use fd / find for fzy as well
- add docs and help

Tho I am against reusing the current implementation and simply sticking fzy after the fd / find since it needs to spawn the command for every keypress while searching. (To be exact, only the last query is executed when multiple are in the queue.)

The current implementation is built to spawn a new fd process everytime because fd is what is doing the filtering but when we use fzy, we only need the result of `fd . -t f (--hidden) (--no-ignore)` and can reuse it while user continues narrowing down.

And as far as I know, caching result of fd is not doable with the current call stack.

cseickel:
> The current implementation is built to spawn a new fd process everytime because fd is what is doing the filtering but when we use fzy, we only need the result of `fd . -t f (--hidden) (--no-ignore)` and can reuse it while user continues narrowing down.

That's true, now that you mention it. I guess you can store the original listing of files in memory and just keep re-using if that is the most performant way to do it.

pysan3:
With the latest commit, I have implemented the fzy algorithm that uses file list taken from `fd` and other commands.

Besides, I have written the document and added filesystem commands `fuzzy_sorter` and `fuzzy_sorter_directory`, as well as a default keybind `["#"] = "fuzzy_sorter",` as mentioned in the README.

With this, I think the feature is implemented so I will mark this PR as ready for review.

@cseickel, @miversen33 I can't thank you enough for the help!
And I would be happy to hear any feedbacks. (Better command name and better default keybind?)

cseickel:
Awesome. I'll check this out and use it for work today.

lua/neo-tree/sources/common/container.lua
```diff
@@ -270,8 +270,10 @@ local merge_content = function(context)
   vim.list_extend(result, left)
   vim.list_extend(result, right)
   context.merged_content = result
-  log.trace("wanted width: ", wanted_width, " actual width: ", context.container_width)
-  context.wanted_width = math.max(wanted_width, context.wanted_width)
+  if wanted_width > context.container_width then
```
cseickel, line 273: I'm not sure why this change is in there, but I noticed that autowidth is broken in this branch, which also means that the position="current" trees are squashed.

cseickel:
@pysan3 @miversen33 Sorry I haven't provided any updates. I have been using it and I think it works well for my repos. I definitely prefer it to the old fuzzy search. 

I haven't looked at the code close enough to approve it, and I haven't tried to test it with bigger repos so I can't give it my blessing as of yet. I'll try to give it a closer look this weekend.

cseickel:
FYI - I experienced a bug just now where all of my trees were sorted in a random, yet consistent manner. That means that they stuck with their odd sort order across refreshes and changing the working directory, but I couldn't figure out what the logic was. I couldn't get them right again until I restarted nvim. I don't know for sure if this was caused by using the fzy search, but it seems like a likely suspect.

pysan3:
I do indeed overwrite the sorting function at the start of the filtering [here](https://github.com/pysan3/neo-tree.nvim/blob/2f2b87067dfd6b6e31d3f7071e5551067081e936/lua/neo-tree/sources/filesystem/lib/filter.lua#L79) and restore the old function at the end of search [around here](https://github.com/pysan3/neo-tree.nvim/blob/2f2b87067dfd6b6e31d3f7071e5551067081e936/lua/neo-tree/sources/filesystem/init.lua#L191). 

I didn't find any edge cases with this logic but if the sort is aborted, I guess the reset_search is never called hence not restored for some reason?

lua/neo-tree/sources/filesystem/lib/filter_external.lua
```diff
@@ -43,133 +43,297 @@ local get_find_command = function(state)
...
+    local function on_exit(_)
+      log.debug(string.format([[fzy_sort_files: cached_everything: %s, len: %s]], cached_everything,
+        #state.fzy_sort_file_list_cache))
+      if not cached_everything then
+        state.fzy_sort_file_list_cache = {}
+      end
+      opts.on_exit(0)
+    end
+
+    M.filter_files_external("fd", pwd, nil, nil, true,
```
cseickel, line 310:
```suggestion
    M.filter_files_external(nil, pwd, nil, nil, true,
```
You shouldn't force this to `fd` because not everyone has `fd` and it should work the same with any of the other configured programs. `fd` is already preferred if it is installed.

lua/neo-tree/sources/filesystem/lib/filter_external.lua
```diff
@@ -43,133 +43,297 @@ local get_find_command = function(state)
...
-    if glob ~= "*" then
-      if glob:sub(1) ~= "*" then
-        glob = "*" .. glob
-      end
-      if glob:sub(-1) ~= "*" then
-        glob = glob .. "*"
-      end
-    end
+---filter_files_external
+-- Spawns a filter command based on `cmd`
```
cseickel, line 58:
```suggestion
```
I don't think there is any valid reason to force a particular program to be used here. It's an internal implementation detail that the consumer of this function should not need to know about.

Also, everyone has different programs available and we should leave it up to this function to choose the best available option.

lua/neo-tree/sources/filesystem/lib/filter_external.lua
```diff
@@ -43,133 +43,297 @@ local get_find_command = function(state)
...
+---@param glob string | nil If not nil, do glob search. Take precedence on `regex`
+---@param regex string | nil If not nil, do regex search if command supports. if glob ~= nil, ignored
+---@param full_path boolean If true, search agaist the absolute path
+---@param types FileTypes | nil Return only true filetypes. If nil, all are returned.
+---@param ignore { dotfiles: boolean?, gitignore: boolean? } If true, ignored from result. Default: false
+---@param limit? integer | nil Maximim number of results. nil will return everything.
+---@param find_args? string[] | table<string, string[]> Any additional options passed to command if any.
+---@param on_insert? fun(err: string, line: string): any Executed for each line of stdout and stderr.
+---@param on_exit? fun(return_val: table): any Executed at the end.
+M.filter_files_external = function(cmd, path, glob, regex, full_path, types, ignore, limit, find_args, on_insert, on_exit)
```
cseickel, line 70:
```suggestion
M.filter_files_external = function(path, glob, regex, full_path, types, ignore, limit, find_args, on_insert, on_exit)
```
As described above, I think the `cmd` parameter should be removed.

cseickel (changes requested):
I think this is excellent work and it should become the default in 3.0. I do have a few requested changes to avoid edge cases and keep it cross platform.

pysan3:
Thank you very much for your reviews. 

I'll definitely look at them tomorrow and update those points. 
I'll ping you when I'm finished. 

Can't wait for it to be merged ;)

BTW, what do you mean by default? Since we already map it to `#`, do you mean to remap it to `/`? Or should we not map it to anything now and on `v3.0` overwrite the `fuzzy_finder` command?

cseickel:
> BTW, what do you mean by default? Since we already map it to `#`, do you mean to remap it to `/`? Or should we not map it to anything now and on `v3.0` overwrite the `fuzzy_finder` command?

I meant that for now we can add it as an additional command, but in 3.0 we will remove the old fuzzy finder from the default mappings and use this one as the command mapped to `/`. 

Unless we drop `/` as a mapping, which is a different conversation, but the point is that this will be the recommended option.

lua/neo-tree/sources/filesystem/lib/filter_external.lua
zolrath: In most integrations relying on `fd` best practice seems to be checking for `fdfind` before checking for `fd` as all Ubuntu/Debian distributions have a *different* program named `fd` pre-installed. 

By checking for/using `fdfind` if present you avoid the issue.

https://github.com/sharkdp/fd#on-ubuntu

pysan3:
@cseickel Thank you for your reviews and sorry being late.

As I have commented on individual comments, these are the changes I made.

- use `sort_function_override` instead of modifying `config.sort_function` ( 2e5ee31 )
  - `sort_by_score` -> `set_sort_by_score` as well
- use `get_find_command` to decide the proper find command. ( d0919a5 )
  - As I mentioned in other places, I did not want the `filter_files_external` function to require `state` as an argument.
  - Meaning that it is responsible for the code that is calling this function to decide the `find_command` beforehand.
- Search for `fdfind` before `fd`. ( 5a21092 )
  - I believe I received a notification where someone pointed out that it will false detect a different `fd` command on Ubuntu based distros. (I can't find that comment anywhere tho)

cseickel (approved):

cseickel:
Let's merge it! 🎉

PR #741, opened 2023-02-11
fix(git): hide gitignored and apply show git status when non-ascii charact…

taikamurmeli:
Hello!

Non-ascii characters caused two problems with git functionality in the tree view due to git's octal utf-8 encoding.

(1) Hiding gitignored did not work. 
(2) Git status of a file was not displayed

For instance, if a git project was under a directory "~/säde/", the 'ä' in the path would become "\303\244" and no files would be hidden in the tree view based on gitignore.
If a filename contained 'ä', the git status, e.g. unstaged, would not show in the tree view.

This PR applies a fix to both git/status.lua and git/ignored.lua to convert any octal patterns to proper utf-8 through a util function in git/utils.lua.


cseickel (approved):
Looks good, thanks for the fix!

PR #745, opened 2023-02-13
Fix syntax of Packer example in Readme

ornicar:
Just adding a missing comma in the config example.


cseickel:
Thanks, but this has already been fixed in `main` and will go out on the next release.

Issue #740, opened 2023-02-09
Restoring a session containing a neo-tree sidebar breaks neo-tree


dhazel:
Thanks for the great plugin!

Neovim version: 0.8.2
OS: Windows 10

Steps to reproduce:

1. Open neovim
2. Open the neo-tree sidebar, `:Neotree`
3. Save the session, `:mksession mysession.vim`
4. Close neovim
5. Open neovim
6. Restore the session, `:source mysession.vim`
7. Observe that the sidebar is created but neo-tree does not render in the sidebar
8. Attempt to toggle the neotree sidebar, `:Neotree`
9. Observe that neo-tree does not render and an error is thrown in messages:

````
Error executing vim.schedule lua callback: ...vim-data/lazy/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:867: Failed to rename buffer
stack traceback:
	[C]: in function 'nvim_buf_set_name'
	...vim-data/lazy/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:867: in function 'create_window'
	...vim-data/lazy/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:1015: in function 'draw'
	...vim-data/lazy/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:1177: in function 'show_nodes'
	...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:98: in function 'render_context'
	...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:126: in function <...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:125>
````

nhat-vo:
Hi! Indeed, this looks like a bug. The neo-tree buffer was saved along with the session, and neo-tree is unable to overwrite that buffer when it is created again. A workaround would be to close the tree before creating the session. I will have a look to see if there is any way to do this automatically.

nyngwang:
@nhat-vo Does #748 resolve this now?

nhat-vo:
Yes, thanks for reminding me. I thought the issue will automatically be closed on PR merge.

cseickel:
> Yes, thanks for reminding me. I thought the issue will automatically be closed on PR merge.

The issue will get automatically closed on **release** in this repo. It's not a big deal either way if you want to close it on merge to `main` though.

ehaynes99:
[...]

PR #748, opened 2023-02-14
fix: delete neo-tree buffers created by :mksession

nhat-vo:
Hi! This PR fixes the issue in #740. I added a function to clean up the dummy neo-tree buffers the first time `renderer.show()` is called.

This is my first contribution to the project, so let me know if you have any comments 😅!

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -17,6 +17,27 @@ local default_popup_size = { width = 60, height = "80%" }
 local floating_windows = {}
 local draw, create_window, create_tree, render_tree
 
+-- clean up neotree buffers created by :mksession
+local cleaned_up = false
+local clean_neotree_buffers = function()
+  if cleaned_up then
+    return
+  end
+  local clean_buffer_with_prefix = function(prefix)
+    for bufnr = 1, vim.fn.bufnr("$") do
+      local bufname = vim.fn.bufname(bufnr)
+      if string.sub(bufname, 1, #prefix) == prefix then
+        vim.api.nvim_buf_delete(bufnr, { force = true })
+      end
+    end
+  end
+  local sources = { "filesystem", "buffers", "git_status" }
+  for _, source in ipairs(sources) do
+    clean_buffer_with_prefix("neo-tree " .. source)
+  end
+  cleaned_up = true
+end
+
 local resize_monitor_timer = nil
 local start_resize_monitor = function()
   local interval = M.resize_timer_interval or -1
@@ -805,6 +826,7 @@ create_window = function(state)
     tabnr = state.tabnr,
   }
   events.fire_event(events.NEO_TREE_WINDOW_BEFORE_OPEN, event_args)
+  clean_neotree_buffers()
 
   if state.current_position == "float" then
     state.force_float = nil
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -23,17 +23,12 @@ local clean_neotree_buffers = function()
   if cleaned_up then
     return
   end
-  local clean_buffer_with_prefix = function(prefix)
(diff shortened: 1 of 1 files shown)


lua/neo-tree/ui/renderer.lua
```diff
@@ -17,6 +17,27 @@ local default_popup_size = { width = 60, height = "80%" }
...
+-- clean up neotree buffers created by :mksession
+local cleaned_up = false
+local clean_neotree_buffers = function()
+  if cleaned_up then
+    return
+  end
+  local clean_buffer_with_prefix = function(prefix)
+    for bufnr = 1, vim.fn.bufnr("$") do
+      local bufname = vim.fn.bufname(bufnr)
+      if string.sub(bufname, 1, #prefix) == prefix then
```
cseickel, line 29: This could be more precise to avoid mistakes. This code is a good way to test if it is a neo-tree buffer:
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/245cf1e68840defcc75a16297740f6203f5a045d/lua/neo-tree/setup/init.lua#L497-L498

lua/neo-tree/ui/renderer.lua
```diff
@@ -17,6 +17,27 @@ local default_popup_size = { width = 60, height = "80%" }
...
 local draw, create_window, create_tree, render_tree
 
+-- clean up neotree buffers created by :mksession
+local cleaned_up = false
+local clean_neotree_buffers = function()
+  if cleaned_up then
+    return
+  end
+  local clean_buffer_with_prefix = function(prefix)
+    for bufnr = 1, vim.fn.bufnr("$") do
```
cseickel, line 27: I think this may create errors if one of the numbers in the range is not a valid buffer. This code would be a better way to iterate over all buffers:
```
  for _, buf in ipairs(vim.api.nvim_list_bufs()) do
    local buf_name = vim.api.nvim_buf_get_name(buf)
```

lua/neo-tree/ui/renderer.lua
```diff
@@ -17,6 +17,27 @@ local default_popup_size = { width = 60, height = "80%" }
...
+      local bufname = vim.fn.bufname(bufnr)
+      if string.sub(bufname, 1, #prefix) == prefix then
+        vim.api.nvim_buf_delete(bufnr, { force = true })
+      end
+    end
+  end
+  local sources = { "filesystem", "buffers", "git_status" }
```
cseickel, line 34: There could be more sources than this. It's best to just iterate over all buffers and use the pattern in the previous comment instead of looking for these specific source names.

cseickel (changes requested):

lua/neo-tree/ui/renderer.lua
nhat-vo: I saw that the buffer created by neo-tree is named `neo-tree filesystem [1]`, but this does not seem to match the pattern in the init file (which requires 4 digits inside the square bracket). Should I remove those `%d`?

cseickel: Oh yeah, this pattern was specifically for "position=current" windows which have a 4 digit id. We really want a pattern that has any number of digits in the second bracket, which might be this:
```lua
"neo%-tree [^ ]+ %[%d+%]"
```

cseickel (approved):

cseickel:
Looks good, thanks!

mehalter:
@nhat-vo @cseickel this breaks when using `views` in neovim with `mkview` and `loadview`. It seems that Neo-tree now thinks that it is restoring a session, so it closes the neotree sidebar that's open even though I'm not restoring an entire session. This needs to be fixed

mehalter:
Let me know if you want me to make a new issue regarding this @cseickel

nhat-vo:
@mehalter did you try to see if you encounter the issue in `main` branch? I pushed #771 a few days ago, which should fix this, but it hasn't been merged into release.

Issue #744, opened 2023-02-11
Render the window right away, even when async


musjj:
File scanning can take a particularly long time, so having `async_directory_scan = "always"` can introduce a huge delay before the window is finally opened.
A better behavior would be to open the window right away, before scanning the directory.

pysan3:
Partially related: #713.

charbelnicolas:
Hopefully something can be done to fix this issue

pysan3:
Hi, I created a possible solution to this issue at #752.

Besides, I am thinking of mainly discussing about this over on #713 so please follow that issue if you are interested.

cseickel:
I have a new fix for this in #1015. Can someone try it out and let me know how it works? 
@musjj @charbelnicolas

Issue #713, opened 2023-01-24
Changing source from Any source to Filesystem closes and reopens neo-tree instead of switching source


braxtons12:
Title says it all, switching source, either by clicking the source_selector or pressing "<" or ">", from one source (eg git_status), back to filesystem causes neo-tree to close then reopen on the filesystem source, instead of just switching to filesystem. This causes a jarring flash of the main window growing to the left then shrinking back when neo-tree reopens. This all occurs in ~1 second (or less).

pysan3:
I am able to reproduce this but not sure why this is happening.

I will dig into this since it might be related to the source selector.

pysan3:
So, I now found out what the cause is. (commenting mainly for my memo)

For all sources, the logic is that the window for the previous source is closed [here](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8238865e1d9c61f1a260c290653f2c419503e0a9/lua/neo-tree/command/init.lua#L161) and then a new window for the next source is calculated afterwards.

> This causes a jarring flash of the main window.

So this is actually a _feature_ and not a bug now.

The reason why we cannot observe a flash on buffers / git source etc is that the new buffer is calculated synchronously and therefore, all happens **before** any rendering update of neovim.
(Technically, there is no `vim.schedule` or other async functions between the previous delete and the spawning of new tree)

So, @branxtos12, if you really hate seeing the flickering, an urgent fix would be to set these options in config to make everything calculated in sync, just like other sources.
However, I would not recommend this setting because (obviously) it does everything synchronously meaning Neotree would require time to show up and neovim completely hangs during that period (which is but like half second so might be better than having flickering? Up to you).

```lua
config = {
  git_status_async = false,
  filesystem = {
    async_directory_scan = "never",
  },
}
```

I believe we can delay the deletion of previous buffer, hence the window, until the calculation is done and shift the deletion to the same rendering frame of creating the new window to avoid this issue.

I will try to implement a solution and create a PR.

charbelnicolas:
Ohh yes, I've been noticing this behavior recently too and the flicker is a bit annoying

charbelnicolas:
https://user-images.githubusercontent.com/14960414/218556272-809e5ba1-6985-4d37-a663-4bcb62ed46cf.mp4

pysan3:
Hi, I have created a PR aiming to solve this issue at #752.

I would be more than happy if you could test it out with your default configs.
[...]

PR #752, opened 2023-02-15
feat(renderer): Create a dummy window to prevent window flickering

pysan3:
Let's tackle this after the fuzzy sorter is settled, but I created a logic to create a dummy window before the rendering starts for `filesystem` source to prevent a flash of window movement.

For more details of this problem, please read #713 and #744.
Thanks to @charbelnicolas, providing a screen record as well.

For the internal explanation of this issue please read [this comment](https://github.com/nvim-neo-tree/neo-tree.nvim/issues/713#issuecomment-1419304319).

I have exposed a config option as noted below to toggle this behavior as well, which however defaults to false so that it does not change any behavior at all by default.

```lua
require("neo-tree").setup({
  filesystem = {
    show_split_window_immediately = false, -- true creates a window right away, before starting async_directory_scan.
    --                                     -- ONLY WORKS WITH SPLIT WINDOWS. (i.e. does not affect opening in float or current)
    --                                     -- this should prevent flickering of windows on switching sources.

  -- THIS WILL AFFECT THE BEHAVIOR WITH FOLLOWING OPTIONS
    async_directory_scan = "auto" or "always",
  },
  git_status_async = true,
  window = {
    position = "left" or "right" or "top" or "bottom",
    -- basically not "float" nor "current"
  },
})
```

For those who are willing test this out (THANK YOU!!), please load the plugin with these options. (Suppose using lazy)

```lua
return {
  "pysan3/neo-tree.nvim"
  branch = "dummy-window-on-source-switch",
  opts = {
    ... -- your regular configs
    filesystem = {
      show_split_window_immediately = true -- make it to true (defaults to false)
    },
  },
}
```

### DIFF
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -367,6 +367,9 @@ local config = {
         ["]g"] = "next_git_modified",
       }
     },
+    show_split_window_immediately = false, -- true creates a window right away, before starting async_directory_scan.
+    --                                     -- ONLY WORKS WITH SPLIT WINDOWS. (i.e. does not affect opening in float or current)
+    --                                     -- this should prevent flickering of windows on switching sources.
     async_directory_scan = "auto", -- "auto"   means refreshes are async, but it's synchronous when called from the Neotree commands.
                                    -- "always" means directory scans are always async.
                                    -- "never"  means directory scans are never async.
--- a/lua/neo-tree/sources/filesystem/init.lua
+++ b/lua/neo-tree/sources/filesystem/init.lua
@@ -171,12 +171,35 @@ M._navigate_internal = function(state, path, path_to_reveal, callback, async)
   end
 end
 
+---create_dummy_window
+---@param state table: Readonly state. Will not change anything except for `state.current_position` but only if it is nil. See `ui.renderer.create_window(state, is_dummy=true)` for more info.
+---@return integer? winid of created window. nil if failed.
+local function create_dummy_window(state)
+  vim.pretty_print(state.current_position)
+  if state.current_position == "float" or state.current_position == "current" then
+    return nil
+  end
+  if not utils.resolve_config_option(state, "show_split_window_immediately", false) then
+    return nil
+  end
+  local win = require("neo-tree.ui.renderer").create_window(state, true)
+  if win ~= nil and win.winid ~= nil then
+    log.trace(string.format([[created dummy window: %s]], win.winid))
+    vim.api.nvim_set_current_win(win.winid)
+    return win.winid
+  else
+    log.warn(string.format([[failed to create dummy window]]))
+  end
+  return nil
+end
+
 ---Navigate to the given path.
 ---@param path string Path to navigate to. If empty, will navigate to the cwd.
 ---@param path_to_reveal string Node to focus after the items are loaded.
 ---@param callback function Callback to call after the items are loaded.
 M.navigate = function(state, path, path_to_reveal, callback, async)
   log.trace("navigate", path, path_to_reveal, async)
+  state.close_dummy_window = create_dummy_window(state)
   utils.debounce("filesystem_navigate", function()
     M._navigate_internal(state, path, path_to_reveal, callback, async)
   end, utils.debounce_strategy.CALL_FIRST_AND_LAST, 100)
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -15,7 +15,7 @@ local M = { resize_timer_interval = 50 }
 local ESC_KEY = vim.api.nvim_replace_termcodes("<ESC>", true, false, true)
 local default_popup_size = { width = 60, height = "80%" }
 local floating_windows = {}
-local draw, create_window, create_tree, render_tree
+local draw, create_tree, render_tree
 
 local resize_monitor_timer = nil
 local start_resize_monitor = function()
@@ -776,7 +776,11 @@ local set_window_mappings = function(state)
   state.resolved_mappings = resolved_mappings
 end
 
-create_window = function(state)
+---create_window: Create window based on config and state
+---@param state table The state.
+---@param is_dummy? boolean whether to create a dummy window. Does not work with `state.current_position == "current", "float"`. If true, events are not triggered, and `state` is not modified except `state.current_position` to its new position only if value was nil
+---@return table?: NuiPopup | NuiSplit | ni: Try its best to create new window of some kind
+M.create_window = function(state, is_dummy)
   local default_position = utils.resolve_config_option(state, "window.position", "left")
   local relative = utils.resolve_config_option(state, "window.relative", "editor")
   state.current_position = state.current_position or default_position
@@ -812,9 +816,15 @@ create_window = function(state)
     source = state.name,
     tabnr = state.tabnr,
   }
-  events.fire_event(events.NEO_TREE_WINDOW_BEFORE_OPEN, event_args)
+  if not is_dummy then
+    events.fire_event(events.NEO_TREE_WINDOW_BEFORE_OPEN, event_args)
+  end
 
   if state.current_position == "float" then
+    if is_dummy then
+      log.warn("`create_window` does not work with `state.current_position` == float")
+      return nil
+    end
     state.force_float = nil
     -- First get the default options for floating windows.
     local sourceTitle = state.name:gsub("^%l", string.upper)
@@ -853,6 +863,10 @@ create_window = function(state)
     -- why is this necessary?
     vim.api.nvim_set_current_win(win.winid)
   elseif state.current_position == "current" then
+    if is_dummy then
+      log.warn("`create_window` does not work with `state.current_position` == current")
+      return nil
+    end
     local winid = vim.api.nvim_get_current_win()
     local bufnr = vim.fn.bufnr(bufname)
     if bufnr < 1 then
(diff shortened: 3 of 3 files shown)


charbelnicolas:
I tested your version and it works but there is still a flash inside the window:

https://user-images.githubusercontent.com/14960414/219128772-e3542257-979a-4713-a5ac-4f04a86ce071.mp4

charbelnicolas:
And this is how it looks when using:

`config = {
  git_status_async = false,
  filesystem = {
    async_directory_scan = "never"
  }
}`

https://user-images.githubusercontent.com/14960414/219129108-2571c68c-dea0-4841-ac83-cc22312ee5ac.mp4

This is the one that looks the best to me so I'll keep on using this workaround -_-

pysan3:
Thanks for testing @charbelnicolas !

> I tested your version and it works but there is still a flash inside the window:

I was referring to `flickering` as the instant change in window size of the main window.
Meaning that there is a moment where neo-tree retreats and the main window expands to the whole width, but after that neo-tree kicks in again and the window size goes back.

To prevent this issue, this PR is aimed to create a blank window at the position where neo-tree _will open_ its window so that the main window does not change its size.

The flash you are seeing now is because there is the blank window with no content for a split second. (Which is IMO way better than the main window changing its size)

If you don't want to see that blank window, that should be solved in another PR because it is a whole different story.
In my understanding I couldn't think of a way to solve it without a very large rewrite.

pysan3:
On top of that, in order to visualize the difference, I'd suggest to add an obvious delay to the creation of the window by locking the rendering with an `event_handler`

```lua
require("neo-tree").setup({
   event_handlers = {
      {
        event = "before_render",
        handler = function(_)
          vim.loop.sleep(5000) -- 5 seconds
        end,
      },
   },
})
```

musjj:
Thank you for the PR! Tested it out a bit and it seems to work pretty well.
One thing I'm noticing: with the `show` command, the cursor is not returned into the previous window until the filesystem scanning is complete.
You can try this command:
```lua
require("neo-tree.command").execute { action = "show", source = "filesystem", reveal = true }
```
I'm also encountering a weird bug with big and empty neo-tree windows being spawned after saving a file. I still have yet to come up with an exact set of reproduction steps though.

As for @charbelnicolas's issue, is it not possible to delay the clearing of the window before the scanning is complete?

pysan3:
@musjj Thank you for testing!!

> with the show command, the cursor is not returned into the previous window until the filesystem scanning is complete.

Yes I had a bug in the code which I believe I was able to fix with daa7f86 (one hour ago).

Could you update the repo and try it again?

> I'm also encountering a weird bug with big and empty neo-tree windows being spawned after saving a file. I still have yet to come up with an exact set of reproduction steps though.

Ah that is clearly a bug there. So you are keeping the neo-tree opened at the side while editing the file right? I didn't notice this because I don't have it opened while working on files.

This is caused by neo-tree watching for file change, and it rerenders itself when there is a change in status of any of the files. Here it also triggers the blank window mechanism resulting to two windows.

pysan3:
> I'm also encountering a weird bug with big and empty neo-tree windows being spawned after saving a file. I still have yet to come up with an exact set of reproduction steps though.

Should be fixed by bd9fc09 ! Thanks for pointing out @musjj!!

pysan3:
> As for @charbelnicolas's issue, is it not possible to delay the clearing of the window before the scanning is complete?

I'm not sure if I'm correct but I believe the current implementation assumes that the old buffer of the same source is already closed, therefore many values of the `state` are overwritten along the way.
So, at the time of the creation of the next window, I am not sure if `state.winid` is pointing to the old window so that we can `vim.api.nvim_win_close()` (in all edge cases).
In order to be exactly sure, I think we need to clone `state` before the computation, but then we have to carefully _free_ it so that GC can delete it.

In addition, the current code is like below, where deletion of the other buffers happen in `common.execute` to assure that only one neo-tree buffer is opened at the same time.
If we want to delay the deletion, we have to rely on each separate sources to correctly handle deletion for other sources, which would become much harder to maintain cleanly. (And what about custom sources?)

```txt
- common.execute
  - do_show_or_focus
  - close_other_sources <- deletion here!!
  - manager.navigate
    - require("sources.<name of next source>").navigate => e.g. sources.filesystem.navigate
      - ASYNC? _navigate_internal -- and a massive chain of function call nightmare
      - fs_scan.get_items
      - ASYNC? sync_scan
      - ASYNC? job_complete
      - render_content
      - enderer.show_nodes
      - draw
      - M.create_window <- window creation at last
```

Another solution would be to use a `callback` (which does already exists in the current implementation which we pass all the way from `do_show_or_focus`) but that callback is called _after_ `M.create_window` asyncly meaning that the creation happens first, then neovim rerenders (with two neo-tree windows), and then the deletion happens.
So are we going to add another callback executed at a different timing?

---

This is by far what I have managed to figure out.
I don't think it is _impossible_ to do it somehow but I couldn't find an easy and simple way to manage this situation and therefore thought it would be worth creating this PR with a dummy window instead.

pysan3:
@musjj Thank you for your quick response.

I'm sorry to bother you many times but I pushed a new commit adding one more condition on preventing creation of dummy window.
(Noticed `filesystem.navigate` was called from way more places than I initially thought.)

I'm not quite sure about the BufEnter situation tho.
b553f8e could possibly solve it?

musjj:
Here's an example autocmd:
```lua
vim.api.nvim_create_autocmd("BufEnter", {
  once = true,
  callback = function(opts)
    require("neo-tree.command").execute { action = "show", source = "filesystem", reveal = true }
  end,
})
```

You can try making a dummy folder:
```bash
mkdir /tmp/batch && cd /tmp/batch
touch {0..9999}.txt # or any other arbitrary number high enough to visibly slow down the scan
nvim 0.txt
```
On startup, the sidebar is opened and the cursor stays in it until the filescanning is complete.
You can also try removing the autocmd, then open the filetree manually and you won't see this issue.

Another issue I'm seeing: the cursor is moved back to the filetree once the scanning is complete, before returning it again to the previous window. This can be problematic, because if you're currently typing during this process, you can accidentally modify the neo-tree buffer.

cseickel:
> So, at the time of the creation of the next window, I am not sure if `state.winid` is pointing to the old window so that we can `vim.api.nvim_win_close()` (in all edge cases).
> In order to be exactly sure, I think we need to clone `state` before the computation, but then we have to carefully _free_ it so that GC can delete it.

I would not want to do that, it's a core assumption that there is a singular state for each source/location which is mutated in place. If we break from that the whole thing needs to be re-written because you can't break assumptions without creating bugs.

This conversation has convinced me that the core problem is that it was just not architected to handle this situation as smoothly as some everyone would like.

This PR is a good attempt at a patch to handle the one specific issue of jarring window resizes.

If we really want to have a smooth transition from one source to another, we need to come up with a new abstraction for managing windows which is separate from the `state` object which is owned by a specific source. Exactly how that would work I don't know, but that's what is needed to do it right. The final implementation doesn't have to be complicated, but it will take a lot of thought and possibly some experimentation to get it right.

cseickel (dismissed):
@pysan3 This looks good for the original intent. Were you considering refactoring it to accommodate the other requests or do you want to merge this as-is?

pysan3:
Hi, sorry for the late response. 

May I freeze this for about a week due to lack of time in my real life?

I'll try to fix the problems above and ping you when I have an update. 

Sorry for the inconvenience...

pysan3:
Hi everybody, sorry for not updating for so long.

I actually haven't made any updates and am quite wondering whether this should be merged or not.
Has anyone come up with a better solution to this problem?

Anyways, I have updated the branch to the latest main and should be able to be merged.
If @cseickel is OK, I am OK to merge this PR.

cseickel:
Hey @pysan3 Thanks so much for following up and submitting all these fixes. I'm a little backed up right now but I will get to them soon. I might tackle them out of order and just review whichever one looks the simplest first.

nhat-vo:
It's such a bummer to see this being closed. @pysan3 Thanks a lot for the effort. 

I'm wondering if it's possible to directly hijack the current window (if exists) to render the content instead of creating a new one. I think this way might be less of a "hack", but it would also mean a lot of changes to the code.

What do you think? @pysan3 @cseickel

pysan3:
I think it won't be impossible tbh. (tho the hardest part would be to check the backwards compatibility of the new method)

https://github.com/nvim-neo-tree/neo-tree.nvim/pull/752#issuecomment-1433218506

I guess the best way would be to implement `do_show_or_focus_or_hijack` instead of `do_show_or_focus` and change `close_other_sources` to `close_other_sources_but_keep_the_buffer` where it does clear any trace of the previous `state` (literaly the `state` variable used throughout the code).

And the rest should be able to remain the same.

pysan3:
Oh, I'm pretty surprised that you call this approach a hack (tho the old way definitely is). 

But anyways, I won't be working on this anytime soon so I'd be happy to see that re-architect cuz decoupling buffer management and tree rendering (I took it that this is the aim of the new approach?) would definitely make the code more readable.

cseickel:
> Oh, I'm pretty surprised that you call this approach a hack (tho the old way definitely is).

Honestly, the existing code is a hack as it is. I just don't want to add on to it and go deeper on those original mistakes. Sources and the commands never should have had the job of window management.

nhat-vo:
I'm currently working on making the filter and fuzzy search available to the symbols source, so I will have a look at this refactor once I'm done. Thanks for the inputs!

pysan3:
Just to be clear, are you @cseickel going to do the window manager refactoring anytime soon?

And will it be a major version jump (`v3`)?

cseickel:
@pysan3 @nhat-vo I don't see myself having time to do anything like this for a while. As it is I can't barely keep up with PRs. If either of you want to take it on, go ahead and claim it when you are ready.

It would likely require a major version bump. We are overdue for 3.0 anyhow. I also never intended for major versions to last as long as v2.x did, so I wouldn't shy away from more frequent major versions as opposed to trying to plan and complete all possible changes in 3.0.

PR #760, opened 2023-02-19
fix(renderer): remove closed windows from floating_windows

### LINKED ISSUE #759: [Bug] Resizing in tmux pane errors
Resizing in a tmux pane gives an error message:

```
[Neo-tree ERROR] Error in event handler for event vim_resized[table: 0x7fcde9dc7048]: ...q/.local/share/nvim/lazy/nui.nvim/lua/nui/utils/init.lua:23: Invalid window id: 1027
```

nhat-vo:
Fixes #759 by cleaning up invalid windows from `floating_windows`, which is called on `events.VIM_RESIZED`


lua/neo-tree/ui/renderer.lua
```diff
@@ -933,6 +922,22 @@ M.is_window_valid = function(winid)
   end
 end
 
+M.update_floating_window_layouts = function()
+  for i, win in ipairs(floating_windows) do
+    --first remove any invalid windows
+    if M.is_window_valid(win.winid) == false then
+      table.remove(floating_windows, i)
+    else
```
cseickel, line 930: Removing values from the middle of a table you are iterating on is a bad idea and could produce bugs. In practice we don't usually have more than one window in this collection so you may not have noticed any bugs in testing, but one day this could come back to bite somebody.

It's also generally a good idea to use truthy assertions rather than negative ones where possible, because it is a little easier to read them.

I would suggest:
```suggestion
    valid_windows = {}
    if M.is_window_valid(win.winid) then
      table.insert(valid_windows, win)
```

and then at the end of the loop you can assign:
```lua
  floating_windows = valid_windows
```
to get rid of any invalid windows you may have found along the way.

nhat-vo: Yes, thank you for the comment. I forgot the fact that `table.remove()` could cause bugs in a loop.

cseickel (approved):

PR #762, opened 2023-02-21
fix(command): do not collapse root node

### LINKED ISSUE #646: Issue when closing node together with hide_root_node
👋🏼,

I've been trying the `hide_root_node` setting, since it gives a cleaner look but noting that it doesn't play nicely when triggering `close_node`. It basically leaves you in a state that is completely empty and triggering `open` results in an error. The only way, that I've found, to repair the state is to press <kbd>R</kbd>. 

Reproduce:
- Enable `hide_root_node`.
- Open `Neotree`.
- Press `C`.
- You're now in an empty state.
- Press `CR`: it should give you an error.
- Press `R` to reset and repair the state.

nhat-vo:
Fixes #646. I added some checks in `close_node` to prevent it from collapsing the root node (similarly to `close_all_nodes`).

### DIFF
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -152,7 +152,10 @@ M.close_node = function(state, callback)
     target_node = parent_node
   end
 
-  if target_node and target_node:has_children() then
+  local root = tree:get_nodes()[1]
+  local is_root = target_node:get_id() == root:get_id()
+
+  if target_node and target_node:has_children() and not is_root then
     target_node:collapse()
     renderer.redraw(state)
     renderer.focus_node(state, target_node:get_id())


lua/neo-tree/sources/common/commands.lua
```diff
@@ -152,7 +152,10 @@ M.close_node = function(state, callback)
     target_node = parent_node
   end
 
-  if target_node and target_node:has_children() then
+  local root = tree:get_nodes()[1]
```
cseickel, line 155: I'm not sure if this is the best way to determine if it the root or not. What happens when the hide root config is used? Is the root still in this collection at position 1 or could that be a normal folder that should be collapsible?

nhat-vo: I have tested the code with the `hide_root_node` config set, and things work well. According to the [nui's documentation](https://github.com/MunifTanjim/nui.nvim/tree/main/lua/nui/tree#treeget_nodes), `get_nodes` returns a list of roots of the tree, and in our case it is a singular list containing the directory's root.

If I'm not wrong then `hide_root_node` is only used [here](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/main/lua/neo-tree/ui/renderer.lua#L1128), which only prevent the root node from being rendered from the screen. The tree itself is never modified, so I believe the `root` node should remain the same.

cseickel: OK, just wanted to make sure that situation was tested. Thanks for the explanation.

cseickel (approved):

PR #779, opened 2023-02-27
fix: do not auto clean up buffers after session restore, unless opted in

### LINKED ISSUE #778: Automatic buffer cleaning after session restore should be opt-in
@nhat-vo @cseickel this breaks when using `views` in neovim with `mkview` and `loadview`. It seems that Neo-tree now thinks that it is restoring a session, so it closes the neotree sidebar that's open even though I'm not restoring an entire session. This needs to be fixed

_Originally posted by @mehalter in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/748#issuecomment-1447124435_

cseickel:
fixes #778

This moves the session cleanup functionality to an opt-in feature instead of enabling it automatically for everyone.

The new config option is:
```lua
  auto_clean_after_session_restore = false, -- Automatically clean up broken neo-tree buffers saved in sessions
```

I also moved the first cleanup action from the renderer to the setup function.

### DIFF
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -9,6 +9,7 @@ local config = {
     "git_status",
   },
   add_blank_line_at_top = false, -- Add a blank line at the top of the tree.
+  auto_clean_after_session_restore = false, -- Automatically clean up broken neo-tree buffers saved in sessions
   close_if_last_window = false, -- Close Neo-tree if it is the last window left in the tab
   -- popup_border_style is for input and confirmation dialogs.
   -- Configurtaion of floating window is done in the individual source sections.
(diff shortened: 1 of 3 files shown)


cseickel:
@nhat-vo Can you test this out and confirm it still works as expected when you opt-in to the functionality by setting `auto_clean_after_session_restore = false,` in your config? 

Also, let me know if you think that is a good name for the config option.

cseickel:
Thanks for confirming @nhat-vo. 

After thinking about this, I realize that any new functionality needs to be opt-in because you never now how it may interact with all of the different setups people have, especially when doing something like deleting buffers with an autocmd. Even if the default value is eventually `true`, you still need an escape hatch for people to be able to disable it when they need to.

PR #790, opened 2023-03-08
fix: error opening file when window.width is not lua number

### LINKED ISSUE #789: error opening file when window.width is not lua number and neo-tree is only opened window
When I set window.width to a string or a function, neo-tree throws an error when trying to open a file when neo-tree is the only window open.

 It works fine when I have neo-tree open and another window open, only when neo-tree is the only buffer open, it breaks.

Example setup:
```lua
require("neo-tree").setup({
  window = {
    width = "30%"
  }
})
```
Error code:
```
E5108: Error executing lua: ...cal/share/nvim/lazy/neo-tree.nvim/lua/neo-tree/utils.lua:508: Expected lua number
stack traceback:
        [C]: in function 'nvim_win_set_width'
        ...cal/share/nvim/lazy/neo-tree.nvim/lua/neo-tree/utils.lua:508: in function 'open_file'
        ...y/neo-tree.nvim/lua/neo-tree/sources/common/commands.lua:561: in function 'open'
        ...y/neo-tree.nvim/lua/neo-tree/sources/common/commands.lua:592: in function 'open_with_cmd'
        ...y/neo-tree.nvim/lua/neo-tree/sources/common/commands.lua:601: in function 'open'
        ...o-tree.nvim/lua/neo-tree/sources/filesystem/commands.lua:184: in function 'func'
        ...are/nvim/lazy/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:782: in function <...are/nvim/lazy/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:780>
```

From my understanding of the code, I'm guessing it works normally because neo-tree uses nui to open the split, and nui accepts percentages and functions. However, when opening a file when neo-tree is the only window, `vim.api.nvim_win_set_width` is called and `window.width` is passed without any extra processing.

Perhaps the code similar to [calc_container_width](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/aec592bb1f0cf67f7e1123053d1eb17700aa9ed4/lua/neo-tree/sources/common/container.lua#LL20C4-L20C4) could be used, as container width support strings and functions already.

williamhCode:
fixes #789. One comment is that if the width is invalid, normally neo-tree still works and doesn't throw an error. However, since I added this fix, errors only appear if you happen to open a file when there's no other windows (which only happens rarely). This is a bit inconsistent and might confuse users, should we bother at all?

### DIFF
--- a/lua/neo-tree/utils.lua
+++ b/lua/neo-tree/utils.lua
@@ -495,7 +495,28 @@ M.open_file = function(state, path, open_cmd)
       -- TODO: make this configurable, see issue #43
       if is_neo_tree_window then
         -- Neo-tree must be the only window, restore it's status as a sidebar
-        local width = M.get_value(state, "window.width", 40, false)
+        local default_width = 40
+        local width = M.get_value(state, "window.width", default_width, false)
+        local available_width = vim.api.nvim_win_get_width(0)
+        if type(width) == "string" then
+          if string.sub(width, -1) == "%" then
+            width = tonumber(string.sub(width, 1, #width - 1)) / 100
+          else
+            width = tonumber(width)
+          end
+          width = math.floor(available_width * width)
+        elseif type(width) == "function" then
+          width = width()
+          if type(width) ~= "number" then
+            width = default_width
+          else
+            width = math.floor(width)
+          end
+        elseif type(width) == "number" then
+          width = math.floor(width)
+        else
+          width = default_width
+        end
         local nt = require("neo-tree")
         local split_command = "vsplit "
         -- respect window position in user config when Neo-tree is the only window


cseickel:
> errors only appear if you happen to open a file when there's no other windows (which only happens rarely). This is a bit inconsistent and might confuse users, should we bother at all?

I would say that in this context, we should just set `width` to the default value of `40` and not throw any errors. If we are going to validate the value, it should really be done in the setup method.

williamhCode:
I updated the commit so it matches nui's size handling more and sets the default value of 40. There's still things that don't quite match, such as if the width is very small and if the width is invalid. Because of this, I'm thinking maybe instead of doing a vsplit, we could just switch the current window to the file, then toggle the tree. Then, we don't have to try to match the width handling to nui.

cseickel:
> we could just switch the current window to the file, then toggle the tree. Then, we don't have to try to match the width handling to nui.

That won't work. A lot of people get really finicky about seeing the window open and close. There's also the issue of the built-in window stealing prevention and having to keep the cursor and scroll in the same positions.

I think what you have here is good.

cseickel (approved):

PR #800, opened 2023-03-11
feat(filesystem): use fd/find as first level filter in fzy, cancel jobs as needed

cseickel:
The advantage here is that the fzy search will work in much larger directories because we no longer try to load **all** of the files and fzy search in memory. Now we will have a first level search using the external program which searches for the existence of all of the characters in the term.

The downside is that it will no longer find matches that are in a different order than what you type. I think that is a reasonable trade off because the old way was only usable in small projects.

If we want to recreate a perfect fzy search, we should just use fzy itself. I think we may be able to use [nvim-fzf](https://github.com/vijaymarupudi/nvim-fzf) or at least lean from its code base.

The other change made here is that we now keep track of running jobs for the external find command and kill them when a new one comes in or when the search is exited.

### DIFF
--- a/lua/neo-tree/sources/filesystem/init.lua
+++ b/lua/neo-tree/sources/filesystem/init.lua
@@ -184,6 +184,8 @@ end
 
 M.reset_search = function(state, refresh, open_current_node)
   log.trace("reset_search")
+  -- Cancel any pending search
+  require("neo-tree.sources.filesystem.lib.filter_external").cancel()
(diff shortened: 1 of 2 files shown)


cseickel:
@pysan3 FYI: I made some changes that will make it work in larger directories without bringing the computer to a grinding halt. You do lose the ability to find results that are out of order, but I think it is a reasonable trade off.

Issue #451, opened 2022-07-21
How I exit fuzzy finder without close tree?


guilhermeprokisch:
How I do search with '/' and exit to navigate the result tree?

<img width="559" alt="image" src="https://user-images.githubusercontent.com/12011070/180267007-936138de-176d-4bb3-a9b4-11251ad77598.png">

guilhermeprokisch:
Thanks, but it's doesn't working 

 Why not just exit the the filter mode and keep the tree without refresh keeping the filter?

cseickel:
> Why not just exit the filter mode and keep the tree without refresh and keeping the filter?

That's not really how a fuzzy finder should work to me. You can use the `filter_on_submit` command to keep a filtered list after leaving the search popup.

I think the more constructive thing would be to figure out why the navigation mappings are not working. Could it be they are being overridden with different mappings in some autocmd you have configured?

guilhermeprokisch:
> That's not really how a fuzzy finder should work to me. You can use the filter_on_submit command to keep a filtered list after leaving the search popup. That command is mapped to f by default.

That's exactly the functionality that I was looking for :)

cseickel:
> > You can use the `<Up>`/`<C-p>` and `<Down>`/`<C-n>` keys to select items in the search results, then hit enter to open it.
> 
> and or the arrows don't work for me... Also, I can't find it documented in :help neo-tree.

You're right that it's not documented, it's just common convention. I assume everyone has used Telescope or fzf and that these movements would be intuitive.

If it's not working, you probably have something in your config that is overriding the mappings.

adoyle-h:
`<Up>/<C-p>` and `<Down>/<C-n>` keys not work for me too. And there is no keymap binding for `:map <C-n>`.
@cseickel

cseickel:
Same answer @adoyle-h, there must be something in your config which is overriding these mappings. If it's not explicitly written in your own config files, it could be another plugin.

Those mapping would be commonly used by an autocomplete plugin, maybe it's in that config.

nyngwang:
> `<Up>/<C-p>` and `<Down>/<C-n>` keys not work for me too. And there is no keymap binding for `:map <C-n>`. @cseickel

@adoyle-h And it's possible that you set `setup { use_default_mappings = false }`. Did you resolve it?

adoyle-h:
@nyngwang I have found it is my fault. It not key mapping overridden, but the cursor line off.
[...]

PR #818, opened 2023-03-19
feat: add fuzzy_finder_mappings to customize mappings for filter popup in fuzzy_finder_mode

adoyle-h:
related issue: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/451#issuecomment-1475284693

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,7 +292,14 @@ use {
               ["<c-x>"] = "clear_filter",
               ["[g"] = "prev_git_modified",
               ["]g"] = "next_git_modified",
-            }
+            },
+
+            fuzzy_finder_mappings = { -- define keymaps for filter popup window in fuzzy_finder_mode
+              ["<down>"] = "move_cursor_down",
+              ["<C-n>"] = "move_cursor_down",
+              ["<up>"] = "move_cursor_up",
+              ["<C-p>"] = "move_cursor_up",
+            },
           }
         },
         buffers = {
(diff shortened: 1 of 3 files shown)


cseickel:
@adoyle-h While this certainly works, it seems a little rushed. For one, it doesn't make sense for the fuzzy finder mappings config to be at the root level of the config when it only applies to the filesystem source. Also, you are exposing a lot of internal implementation details with the arguments being passed and your example.

It would probably be better to have a mappings config for the filter popup that is similar to the mappings used on the main window. In this particular example, the logic for how the focus is moved up or down should not be exposed to the end user. We should be giving them ready made command names that they just need to provide the mapping(s) for.

adoyle-h:
> It would probably be better to have a mappings config for the filter popup that is similar to the mappings used on the main window. 

Agree. I will fix it.

> We should be giving them ready made command names that they just need to provide the mapping(s) for.

How to create mappings and bind them with the filter popup? The `input` node is created in show_filter function. Its `bufnr` is unpredictable.

README.md
```diff
@@ -292,7 +292,14 @@ use {
               ["<c-x>"] = "clear_filter",
               ["[g"] = "prev_git_modified",
               ["]g"] = "next_git_modified",
-            }
+            },
+
```
cseickel, line 296:
```suggestion

```
Remove extra blank line.

cseickel (approved):
Looks good, thanks!

cseickel (approved):

ditsuke:
Works very well, thank you!

PR #837, opened 2023-03-30
feat: add "none" as an option for cwd_target

### LINKED ISSUE #835: Add support for cwd_target = "none"
Would it be possible to add `"none"` as a `cwd_target` option? I use neo-tree in two ways:

1. Position = `sidebar`: When I change the root directory here, I would like the `cwd` in Neovim to update. This already works great with `cwd_target = "global"`.
2. Position = `current`: Here, I'm navigating for a nearby file. As I browse, I don't want the `cwd` to change at all.

I hope this might be as simple as returning `nil, nil` in `get_params_for_cwd` and some `if` checks in the two call sites, but I don't know if there are other reasons why this would be insufficient. Happy to put up a PR if that's all the work that's involved. Thank you!

apazzolini:
Closes #835 

I've been using this locally for a few weeks and it's worked well. I'm not sure if anything special needs to be done with the nil result where `get_params_for_cwd` is called inside the `get_cwd` function, but I haven't seen any errors so far.


cseickel (commented):
The only thing I don't really like is that `M.get_cwd` will have an error on line 317. It will be handled appropriately in the end but I don't like that we are knowingly creating an error.

Maybe it can be changed to something like this:

```lua
M.get_cwd = function(state)
  local winid, tabnr = get_params_for_cwd(state)
  local success, cwd = false, ""
  if winid or tabnr then
    success, cwd = pcall(vim.fn.getcwd, winid, tabnr)
  end
  if success then
    return cwd
  else
    success, cwd = pcall(vim.fn.getcwd)
    if success then
      return cwd
    else
      return state.path
    end
  end
end
```

apazzolini:
> Maybe it can be changed to something like this:

Makes sense to me! I pushed up that change.

Issue #743, opened 2023-02-11
Feature request: add highlight for open files


serhez:
It would be nice to have a highlight for files which are open in existing buffers. This is a nvim-tree feature which I miss having.

miversen33:
To be clear, this request is for _more_ than simply having the cursorline on the file that is open, correct? That is already something that Neo-tree supports, so I want to be sure this is an ask that is not already solved via configuration :)

serhez:
Hey @miversen33 ! Thanks for the reply. No, I don't mean just having automatic cursor positioning over the line of the currently open file. I mean that every file on the tree which is currently open in some buffer would be highlighted differently from other files.

Here is an example of how it looks in `nvim-tree` (the highlighting group can be changed of course):
![Screenshot 2023-02-14 at 21 03 16](https://user-images.githubusercontent.com/18057093/218832631-3cd3ae1f-cff5-43a7-ac43-a268c28f21ea.jpg)

nhat-vo:
@serhez The highlight open files feature that you requested has been merged in #838. Have a look to see if it matches your expectation 😁.

serhez:
Hey @nhat-vo, thanks so much! Unfortunately, I can't seem to get it working. I updated the plugin to the latest commit and have added the following snippet to my config, but I don't get any highlighting on open files, maybe I'm missing something?

```lua
default_component_configs = {
  name = {
    highlight_opened_files = true,
  },
},
```

miversen33:
I am not able to get this functioning either. Instead, I am noticing the file names that are opened are being italicized?

![image](https://user-images.githubusercontent.com/2640668/230749070-78c7295a-2134-4189-be18-2c67ee74ca86.png)


My config for this
```lua
default_component_configs = {
    name = {
        highlight_opened_files = true
    }
}
```

nhat-vo:
@serhez @miversen33 That's odd... Could you check the value of the highlight group `NeoTreeFileNameOpened`?

serhez:
The highlight group exists but it's defined to be the same as `Normal`, additionally with `gui=bold`.

nhat-vo:
@serhez That is the group that is used to highlight the opened files. What happens when you set it to some other value (maybe changing the fg/bg)?

serhez:
I'm not familiar with your code base, but maybe `highlights.lua:245` is the culprit?

```lua
[...]

PR #838, opened 2023-03-30
feat: add highlight open files

nhat-vo:
Added highlight open file feature as described in #743. I'm not sure if there is any better way to cache this so that we do not have to loop over all buffers every time. Do you think adding to `state` is a good idea?

### DIFF
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -328,7 +328,7 @@ M.name = function(config, node, state)
     end
   end
 
-  if config.highlight_open_files then
+  if config.highlight_opened_files then
     local opened_buffers = state.opened_buffers or {}
     if opened_buffers[node.path] ~= nil then
       highlight = highlights.FILE_NAME_OPENED
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -292,9 +292,10 @@ M.icon = function(config, node, state)
 end
 
 M.modified = function(config, node, state)
-  local modified_buffers = state.modified_buffers or {}
+  local opened_buffers = state.opened_buffers or {}
+  local buf_info = opened_buffers[node.path]
 
-  if modified_buffers[node.path] then
+  if buf_info and buf_info.modified then
     return {
       text = (make_two_char(config.symbol) or "[+] "),
       highlight = config.highlight or highlights.MODIFIED,
@@ -328,8 +329,8 @@ M.name = function(config, node, state)
   end
 
   if config.highlight_open_files then
-    local modified_buffers = state.modified_buffers or {}
-    if modified_buffers[node.path] ~= nil then
+    local opened_buffers = state.opened_buffers or {}
+    if opened_buffers[node.path] ~= nil then
       highlight = highlights.FILE_NAME_OPENED
     end
   end
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -327,6 +327,18 @@ M.name = function(config, node, state)
     end
   end
 
+  if config.highlight_open_files then
+    for _, buf in ipairs(vim.api.nvim_list_bufs()) do
+      if
+        vim.api.nvim_buf_is_loaded(buf)
+        and vim.fn.buflisted(buf)
+        and vim.api.nvim_buf_get_name(buf) == node.path
+      then
+        highlight = highlights.FILE_NAME_OPENED
+      end
+    end
+  end
+
   if type(config.right_padding) == "number" then
     if config.right_padding > 0 then
       text = text .. string.rep(" ", config.right_padding)
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -328,14 +328,9 @@ M.name = function(config, node, state)
   end
 
   if config.highlight_open_files then
-    for _, buf in ipairs(vim.api.nvim_list_bufs()) do
-      if
-        vim.api.nvim_buf_is_loaded(buf)
-        and vim.fn.buflisted(buf)
-        and vim.api.nvim_buf_get_name(buf) == node.path
-      then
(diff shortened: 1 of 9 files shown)


miversen33:
I love this idea :) That said, honestly I feel like (IMO) looping over the buffers each time isn't that big of a deal, _especially_ if we only perform said change on a Buffer Change event (such as BufDelete/BufAdd event).  At the end of the day, lua is fast and a loop operation (even over several thousand items) per event loop iteration is still effectively instant. Thats not to say that thinking about "caching" or reducing the work is a bad thing, just that its **probably** not going to have a massive performance impact. Even less of an impact if we tie it only to event changes.

One thing I would be curious about, this will highlight _all_ open files with the same highlight right? Is there a way we can highlight the currently focused file _different_ if its one of the files visible in Neotree?

nhat-vo:
@miversen33 Thanks for your insights! I also think that caching might not result in a noticeable impact in performance, and adding the buffer list into `state` seems rather complicated so I'm not sure if I should implement this.

Regarding the focused file, are you talking about the `follow_current_file` option?

miversen33:
> Regarding the focused file, are you talking about the follow_current_file option?

Yes. Does the proposed highlighting overrule that option? I am at work so I cannot test this right now lol. My main concern is that while having all open files highlighted makes a ton of sense, I wouldn't want this to supersede other options (such as `follow_current_file`)

miversen33:
Reading code is hard lol. It looks like you are already using the event system for triggering the highlight logic. As long as the current focused file is distinct from the other open files, LGTM :)

lua/neo-tree/sources/common/components.lua
```diff
@@ -327,6 +327,18 @@ M.name = function(config, node, state)
     end
   end
 
+  if config.highlight_open_files then
+    for _, buf in ipairs(vim.api.nvim_list_bufs()) do
```
cseickel, line 331: This is really the wrong place to do this, because then it will do this loop for every row in the tree which is wasteful.

lua/neo-tree/sources/filesystem/init.lua
```diff
@@ -369,6 +369,17 @@ M.setup = function(config, global_config)
     })
   end
 
+  if global_config.highlight_open_files then
+    for _, event in ipairs({ events.VIM_BUFFER_ADDED, events.VIM_BUFFER_DELETED }) do
+      manager.subscribe(M.name, {
+        event = event,
+        handler = function()
+          manager.refresh(M.name)
```
cseickel, line 377: This is the right place to collect all of the open buffers and add it to the state. Search for `modified_buffers_changed` for a good example of how to do this.

nhat-vo: Ah yes I misunderstood the logic behind state and didn't manage to pass it to the update function.

lua/neo-tree/utils.lua
```diff
@@ -246,11 +246,13 @@ end
 M.get_modified_buffers = function()
   local modified_buffers = {}
   for _, buffer in ipairs(vim.api.nvim_list_bufs()) do
-    local buffer_name = vim.api.nvim_buf_get_name(buffer)
-    if buffer_name == nil or buffer_name == "" then
-      buffer_name = "[No Name]#" .. buffer
+    if vim.api.nvim_buf_is_loaded(buffer) and vim.fn.buflisted(buffer) then
```
nhat-vo, line 249: Modified buffers should be loaded and listed, so I added this instead of creating a new method to check for open files to avoid code repetition.

lua/neo-tree/utils.lua
```diff
@@ -246,11 +246,13 @@ end
...
+      local buffer_name = vim.api.nvim_buf_get_name(buffer)
+      if buffer_name == nil or buffer_name == "" then
+        buffer_name = "[No Name]#" .. buffer
+      end
+      modified_buffers[buffer_name] = vim.api.nvim_buf_get_option(buffer, "modified")
```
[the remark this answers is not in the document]
nhat-vo: Good idea! I'll fix it later today.

lua/neo-tree/sources/filesystem/init.lua
```diff
@@ -369,6 +369,15 @@ M.setup = function(config, global_config)
     })
   end
 
+  if global_config.default_component_configs.name.highlight_open_files then
```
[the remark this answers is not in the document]
nhat-vo: So in that case we'll have 2 configs  (1 global config and 1 for the name component) to turn this on?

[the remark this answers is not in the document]
cseickel: Yes, but I would default it to enabled since it's not really expensive the way it is handled. In that way, it's really one config to turn it on (the name component config) and one config to turn it off. The global config is really just fail safe for troubleshooting or people that want extreme optimization.

The alternative is to read the finalized merged config on every source and look for whether this feature is enabled, but that would be too complicated and then there's also the possibility that the data get's used for other reasons down the line.

nhat-vo:
@cseickel I have refactored the `modified_buffers` list to `opened_buffers` and added the configs as you suggested. I find `opened` is less ambiguous than `open` in describing the status, as `open` also refers to the action.
To make the change more uniform, I also renamed the library function in `buffers.items` to `opened` instead of `open`, in case you don't mind 😅.

cseickel (changes requested):
This looks great, there's just one more thing. Technically, the `utils` module is part of the public api and changing the signature or return type of a function in the utils module that does not start with a `_` is a breaking change that is not allowed within a major version: https://github.com/nvim-neo-tree/neo-tree.nvim#what-is-a-breaking-change

To work around that, can you just add a function that mimics the old `get_modified_buffers()`? It can be marked as deprecated using a comment:
```lua
--- DEPRECATED: This will be removed in v3
---Gets a lookup of all open buffers keyed by path with the modifed flag as the value
---@return table
M.get_modified_buffers = function()
```

Internally, it should just call the new function and convert it to the simple table of booleans that was returned before.

nhat-vo:
@cseickel Done!

cseickel (approved):
Looks good, thanks!

PR #840, opened 2023-04-02
fix(config): prioritize default source-specific configs over user general configs

nhat-vo:
Currently, any user-defined generic window configuration (`user_config.window`) will override default source-specific window configuration (e.g `default_config.git_status.window`). The current priority is
```
default_config.window < default_config.source.window < user_config.window < user_config.source.window
```
So for example, if I map `A` to `add_directory` in my user config for all windows, which will override the source-specific keymap (`["A"] = git_add_all`).

I think the expected behavior should be 
```
default_config.window < user_config.window < default_config.source.window < user_config.source.window
```

### DIFF
--- a/lua/neo-tree/setup/init.lua
+++ b/lua/neo-tree/setup/init.lua
@@ -572,8 +572,8 @@ M.merge_config = function(user_config, is_auto_config)
     source_default_config.window = vim.tbl_deep_extend(
       "force",
       default_config.window or {},
-      source_default_config.window or {},
-      user_config.window or {}
+      user_config.window or {},
+      source_default_config.window or {}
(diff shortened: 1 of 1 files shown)


cseickel:
I really think that the current priority is the correct one. User config should always override defaults.

The proposed behavior is confusing:
```lua
default_config.window < user_config.window < default_config.source.window < user_config.source.window
```
because it would only make sense if the user reads and understands the entire default config and which things are set at what level, which is really an implementation detail that they should not have to understand and we should not be forced to maintain that implementation detail to avoid breaking user configs.

The current behavior is obvious and consistent and should be kept the way it is.

PR #841, opened 2023-04-03
feat(symbols): Added symbol source

nhat-vo:
A simple source to view LSP symbols in the current file. Currently WIP.
Features:
- [ ] UI:
	- [x] Display all symbols in the current file with symbol kinds
	- [x] Symbols nesting
	- [ ] Configurable kinds' name and icon
	- [x] Auto-refresh symbol list
- [ ] Commands
	- [x] Jump to symbols, open symbol in split,... (`open_split` and friends)
	- [x] Rename symbols (`rename`)
	- [x] Preview symbol (`preview` and friends)
	- [ ] Moving / Copying symbols (not sure if needed)
	- [ ] Hover docs
	- [ ] Call hierarchy 

Feel free to let me know if you have any comments on what additional features should be desired.


nhat-vo:
The `document_symbol` source is good for now (it is not enabled by default as there might still be bugs). I will implement the `hover_docs`, `call_hierarchy`, and `CoC Support` in a future PR.

lua/neo-tree/defaults.lua
```diff
@@ -491,6 +493,61 @@ local config = {
...
+          {"kind_name", zindex = 20, align = "right"},
+          {"kind_icon", zindex = 20, align = "right"},
+          }
+        }
+      },
+    },
+    window = {
+      mappings = {
+        ["<cr>"] = "toggle_node",
+        ["o"] = "jump_to_symbol",
```
cseickel, line 518: I think it is more natural for `<cr>` to be mapped to `jump_to_symbol`, that follows the convention of the rest of the sources. `toggle_node` should be mapped to `<space>` by default.

lua/neo-tree/defaults.lua
```diff
@@ -491,6 +493,61 @@ local config = {
...
+      },
+    },
+    server_filter = "first",
+    kinds = {
+      [1] = { name = "File", icon = "", hl = "Tag" },
```
cseickel, line 523: Why are kinds keyed by numbers? It seems more natural that what is currently the `name` field would be the key. Do the numbers have a special meaning?

lua/neo-tree/setup/init.lua
```diff
@@ -71,6 +71,11 @@ local define_events = function()
   events.define_autocmd_event(events.VIM_COLORSCHEME, { "ColorScheme" }, 0)
   events.define_autocmd_event(events.VIM_CURSOR_MOVED, { "CursorMoved" }, 100)
   events.define_autocmd_event(events.VIM_AFTER_SESSION_LOAD, { "SessionLoadPost" }, 200)
+  events.define_autocmd_event(
+    events.VIM_LSP_UPDATE,
```
cseickel, line 75: The convention is that events that start with `VIM_` are direct mappings to those built-in events. Multiple events are only combined in special situations where their meanings are synonymous for our purposes.

You should just be adding the individual events that do not already exist, rather than creating one super event.

lua/neo-tree/sources/document_symbols/commands.lua
```diff
@@ -0,0 +1,71 @@
...
+    vim.lsp.buf.rename(new_name)
+    M.refresh(state)
+  end
+  local msg = string.format('Enter new name for "%s":', old_name)
+  inputs.input(msg, old_name, callback)
+end
+
+M.open = M.jump_to_symbol
+
+-- mask away default commands
```
cseickel, line 59: If you don't want to take most of the common commands, then just don't use the `cc._add_common_commands` function. Add the ones you want individually instead.

lua/neo-tree/defaults.lua
```diff
@@ -491,6 +493,61 @@ local config = {
...
+  document_symbols = {
+    renderers = {
+      root = {
+        {"indent"},
+        {"icon", default="C" },
+        {"name", zindex = 10},
+      },
+      symbol = {
+        {"indent"},
+        {"icon", default="C" },
```
cseickel, line 505: I feel like the `kind_icon` belongs here by default, between the expander and the name.

lua/neo-tree/sources/document_symbols/components.lua
```diff
@@ -0,0 +1,76 @@
...
+-- contains the following keys:
+--    text:      The text to display for this item.
+--    highlight: The highlight group to apply to this text.
+
+local highlights = require("neo-tree.ui.highlights")
+local common = require("neo-tree.sources.common.components")
+
+local M = {}
+
+M.icon = function(config, node, state)
```
cseickel, line 18: This really isn't an icon, it's an expander. I'm pretty sure the built-in `indent` component already draws these for you if it's configured to do so sp you may not need this at all.

cseickel (commented):
First off, it's awesome that you are undertaking this, I can't wait to start using it!

I do think there is more work to do before it's ready for merging though. I haven't dug deep into the code but I have some suggestions at the surface level here.

I also noticed that it does not automatically update when I change files, which is something that I think would be expected by anyone using it.

nhat-vo:
I'm very glad to hear that you like this! Regarding the automatic update, I will have a look at why it is not auto-updating. Previously, I used the `User LspRequest` event, which updates the tree automatically on change and when the current buffer is changed (and every time any LSP Request is made), but this means that `navigate` is called twice every time the source is opened, so I switched to using these events instead.

lua/neo-tree/defaults.lua
nhat-vo: According to the [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol), the kinds are defined as enums and are returned as integers in the server. 

The conventional list of kinds looks like this
<details>
<summary>Kinds list</summary>
```
export namespace [SymbolKind](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#symbolKind) {
	export const File = 1;
	export const Module = 2;
	export const Namespace = 3;
	export const Package = 4;
	export const Class = 5;
	export const Method = 6;
	export const Property = 7;
	export const Field = 8;
	export const Constructor = 9;
	export const Enum = 10;
	export const Interface = 11;
	export const Function = 12;
	export const Variable = 13;
	export const Constant = 14;
	export const String = 15;
	export const Number = 16;
	export const Boolean = 17;
	export const Array = 18;
	export const Object = 19;
	export const Key = 20;
	export const Null = 21;
	export const EnumMember = 22;
	export const Struct = 23;
	export const Event = 24;
	export const Operator = 25;
	export const TypeParameter = 26;
}
```
</details>

It seems that other servers could define more kinds than this, for example in [here](https://github.com/glepnir/lspsaga.nvim/blob/fb476086012e18e0001c3dcc5b18fd34a847e5fe/lua/lspsaga/lspkind.lua#L59), so I make the kind list configurable by index to accommodate these additional kinds.

nhat-vo: I think it would be a good idea to define the common kinds keyed by `name` as you said and have an option for additional symbols.

nhat-vo:
@cseickel I can't seem to reproduce your issue with automatic file changes. For me, whenever I change the file, the `document_symbols` source is updated accordingly. Could you provide the steps to reproduce this?

lua/neo-tree/setup/init.lua
nhat-vo: @cseickel I think it is better to bundle these events into one `autocmd`, so that they have a unified debouncing time. This will limit the number of calls to the LSP server, which could slow down nvim. I'll name it `NEO_TREE_LSP_UPDATE` if that is fine for you.

cseickel: Actually, the debounce can happen on the subscriber side instead of of on the producer (vim events) side. The important thing here is that we don't want to subscribe to the same autocommand ore then once, that's why one at a time is the way to go.

The function the end up calling should have it's own debounce if you worried about multiple vim events happening at the same time. That is how the filesystem refresh works.

nhat-vo: Currently, I'm switching back to the `LspRequest` event for auto-refresh. It's quite fast with a debounce time of 200ms, and covers many events, so I think it is a good idea for now.

nhat-vo:
Actually, I ended up changing it to several separate events and using `utils.debounce`. Seems to refresh at the right moment and the right amount now.

cseickel (approved):
Looks good.

cseickel:
@nhat-vo I think this looks good as an initial release and I'll merge it now. The two things I would add to make it complete are:

- [ ] Follow cursor: just like "follow current file", I would expect the symbol tree to always have the symbol I'm currently on be selected.
- [ ] fuzzy search: I would implement the search commands to be able to search/filter the list of document symbols.

nhat-vo:
@cseickel Great! I'm very glad to see this merged! I'll create an issue to track the features of this source.

PR #842, opened 2023-04-03
fix: use correct conversion between tabid and tabnr when tracking cwd

jedrzejboczar:
I noticed that neo-tree cwd was sometimes getting out of sync with the real value. After some investigation I found that the manager uses `nvim_get_current_tabpage` to get the value stored in state (tabid but is called tabnr in code), but then uses the same value to get tab cwd using `vim.fn.getcwd` (which expects tab number). The values used in API and in vimscript functions are slightly different - tabnr is the visible position of a tab, while tabid refers to the tab object. 

To reproduce the bugs that I was expecting start Neovim, open two tabs, then set different tabcwd for one of the tabs and open neo-tree in both tabs. Then use `:tabmove +1` in the first (leftmost) tab - neo-tree will be out of sync with the actual tab cwd (as returned by `vim.fn.getcwd(-1, 0)`). You can also compare `vim.api.nvim_get_current_tabpage()` with `vim.fn.tabpagenr()`.

I changed all usages of "tabnr" in code to "tabid" whenever we are referring to API tab handles. To avoid breaking changes I left the following incorrect names: 
* `tabnr` in event args (NEO_TREE_WINDOW_BEFORE_CLOSE, NEO_TREE_WINDOW_BEFORE_OPEN)
* buffer variable `neo_tree_tabnr`
* mentions about these in neo-tree.txt

One notable change that I had to do was in TabClosed event handler. It passes `<afile>` which is a tab number. It was previously being used incorrectly to dispose tab (state was stored by tabid). I found no way to get tabid of the tab that has just been closed. Normally to convert tabnr to tabid we would have to find it iterating over output of `nvim_list_tabpages`, but inside the TabClosed handler this tab is already deleted. Instead of this I created dispose_invalid_tabs which deletes tabs based on `nvim_tabpage_is_valid`. This needs to be scheduled to next tick as inside the event handler the tab still appears to be valid. If scheduling to next tick is not acceptable then we would have to use `nvim_list_tabpages` and find `state.tabid` in the returned list as `nvim_list_tabpages` won't list the already deleted tab even when not scheduled.

### DIFF
--- a/lua/neo-tree/setup/init.lua
+++ b/lua/neo-tree/setup/init.lua
@@ -51,7 +51,7 @@ local define_events = function()
     args.opened_buffers = utils.get_opened_buffers()
     return args
   end
-  
+
   events.define_autocmd_event(events.VIM_AFTER_SESSION_LOAD, { "SessionLoadPost" }, 200)
   events.define_autocmd_event(events.VIM_BUFFER_ADDED, { "BufAdd" }, 200, update_opened_buffers)
   events.define_autocmd_event(
@@ -225,9 +225,9 @@ M.buffer_enter_event = function()
       return
     end
 
-    local current_tabnr = vim.api.nvim_get_current_tabpage()
-    local neo_tree_tabnr = vim.api.nvim_buf_get_var(prior_buf, "neo_tree_tabnr")
-    if neo_tree_tabnr ~= current_tabnr then
+    local current_tabid = vim.api.nvim_get_current_tabpage()
+    local neo_tree_tabid = vim.api.nvim_buf_get_var(prior_buf, "neo_tree_tabid")
+    if neo_tree_tabid ~= current_tabid then
       -- This a new tab, so the alternate being neo-tree doesn't matter.
       return
     end
@@ -268,12 +268,12 @@ M.win_enter_event = function()
   manager.close_all("float")
 
   if M.config.close_if_last_window then
-    local tabnr = vim.api.nvim_get_current_tabpage()
-    local wins = utils.get_value(M, "config.prior_windows", {})[tabnr]
+    local tabid = vim.api.nvim_get_current_tabpage()
+    local wins = utils.get_value(M, "config.prior_windows", {})[tabid]
     local prior_exists = utils.truthy(wins)
     local non_floating_wins = vim.tbl_filter(function(win)
       return not utils.is_floating(win)
-    end, vim.api.nvim_tabpage_list_wins(tabnr))
+    end, vim.api.nvim_tabpage_list_wins(tabid))
     local win_count = #non_floating_wins
     log.trace("checking if last window")
     log.trace("prior window exists = ", prior_exists)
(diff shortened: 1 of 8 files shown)


jedrzejboczar:
> It's a shame we don't have any tests that covers tabs right now. Would you have any time to add some in this PR?

I'll try to add a few tests if I find some free time.

rafi:
The tabid vs. tabnr makes a lot of sense, nice job. I was hoping this fixes the issue with two tabs, each has a window with different cwd and both windows are editing a buffer with a file in each cwd respectively; `:Neotree reveal` on the second tab almost always gets out of sync and says "File not in cwd`. Only after `:Neotree filesystem dir=./` is run, the `:Neotree reveal` starts working again. Unfortunately issue persists with your branch.

lua/neo-tree/ui/renderer.lua
```diff
@@ -838,7 +838,7 @@ create_window = function(state)
   local event_args = {
     position = state.current_position,
     source = state.name,
-    tabnr = state.tabnr,
+    tabnr = state.tabid,
```
cseickel, line 841: see prior comment about tabid/tabnr.

lua/neo-tree/ui/renderer.lua
```diff
@@ -913,7 +913,7 @@ create_window = function(state)
 
   if type(state.bufnr) == "number" then
     vim.api.nvim_buf_set_var(state.bufnr, "neo_tree_source", state.name)
-    vim.api.nvim_buf_set_var(state.bufnr, "neo_tree_tabnr", state.tabnr)
+    vim.api.nvim_buf_set_var(state.bufnr, "neo_tree_tabnr", state.tabid)
```
cseickel, line 916: see prior comment about tabid/tabnr.

lua/neo-tree/setup/init.lua
```diff
@@ -215,9 +215,9 @@ M.buffer_enter_event = function()
       return
     end
 
-    local current_tabnr = vim.api.nvim_get_current_tabpage()
-    local neo_tree_tabnr = vim.api.nvim_buf_get_var(prior_buf, "neo_tree_tabnr")
-    if neo_tree_tabnr ~= current_tabnr then
+    local current_tabid = vim.api.nvim_get_current_tabpage()
+    local neo_tree_tabid = vim.api.nvim_buf_get_var(prior_buf, "neo_tree_tabnr")
```
cseickel, line 219: see next comment about tabid/tabnr.

lua/neo-tree/ui/renderer.lua
```diff
@@ -123,7 +123,7 @@ M.close = function(state)
               position = state.current_position,
               source = state.name,
               winid = state.winid,
-              tabnr = state.tabnr,
+              tabnr = state.tabid,
```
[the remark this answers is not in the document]
jedrzejboczar: I added `tabid` as separate variable, while keeping `tabnr` to have the value of tab number and marked `tabnr` as deprecated in the docs (not sure if I should mark it somewhere in code too?).

lua/neo-tree/ui/renderer.lua
jedrzejboczar: Done

lua/neo-tree/ui/renderer.lua
jedrzejboczar: DoneDone

jedrzejboczar:
Finally found time to get back to this. I followed your suggestion regarding buf variables/event args. I also added simple test for this issue under `tests/sources/manager_spec.lua`.

cseickel (approved):
This looks great, thanks!

PR #846, opened 2023-04-05
feat: commands can be defined globally. fixes #707

### LINKED ISSUE #707: Allow global commands for window
It is great that I can use the `commands` table to define commands that can easily be used in mappings and also provide context to the help menu, but this only works with specific sources like `filesystem`. It would be great to be able to define commands for the entire window that work for all sources or give a way to name a function so that when I press `?` it doesn't just say a mapping is `<function>`

lnc3l0t:
Added **global commands** feature. A user can specify a table with functions that will automatically be added to all sources. 
The table is `config.commands` and it's content is added to `config[source_name].commands` with `vim.tbl_extend`, and the "keep" option, to allow overriding the global implementation with a custom one in any source.

```lua
   require("neo-tree").setup({
     commands = {
       hello = function()            -- define a global "hello world" function
         print("Hello world")
       end
     },

     window = {
       mappings = {
         ["<C-c>"] = "hello"
                    -- define a global mapping to call 'hello' in every source
       }
     },

     filesystem = {
       commands = {
         -- override implementation of the 'hello' action in filesystem source
         hello = function()
           print("Hello inside filesystem")
         end
       }
     }
   })
```
This means that pressing `CTRL-c` in the `filesystem` source will result in a different effect than in other sources.
`hello` is therefore a global command, but in filesystem it is overwritten.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,11 @@ use {
             }
           },
         },
+        -- A list of functions, each representing a global custom command
+        -- that will be available in all sources (if not overridden in `opts[source_name].commands`)
+        -- see `:h neo-tree-global-custom-commands`
+        commands = {}
+
         window = {
           position = "left",
           width = 40,
@@ -299,7 +304,9 @@ use {
               ["<up>"] = "move_cursor_up",
               ["<C-p>"] = "move_cursor_up",
             },
-          }
+          },
+
+          commands = {} -- Add a custom command or override a global one using the same function name
         },
         buffers = {
           follow_current_file = true, -- This will find and focus the file in the active buffer every
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -9,6 +9,7 @@ Mappings .................... |neo-tree-mappings|
   View Changes .............. |neo-tree-view-changes|
   File Actions .............. |neo-tree-file-actions|
   Filter .................... |neo-tree-filter|
+  Global custom commands .... |neo-tree-custom-commands-global|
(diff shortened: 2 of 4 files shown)


cseickel:
> I tried to make a test for the feature, but on my setup Plenary didn't work and the test mechanism seemed rather complex.

As a quick drive by comment, you can run the tests by running
```
make test
```
in the root of the project.

There is also a dockerized version, although I haven't used it in a while:
```
make test-docker
```

mehalter:
@lnc3l0t the [`scripts/test.sh`](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/v2.x/scripts/test.sh) file has the code for setting up the environment. Basically cloning the plugins to the necessary location.

Are tests necessary here? I looked and there currently aren't testing for custom defined `commands` for each source as it currently exists either.

cseickel:
> Are tests necessary here? I looked and there currently aren't testing for custom defined `commands` for each source as it currently exists either.

I'm not a stickler for tests because I don't have the time so I assume most other contributors also don't, but as a general rule I think we need way more tests than we have so don't use what tests already exists as a good benchmark for what is appropriate.

cseickel (approved):
Thanks for contributing and actually adding documentation! Sorry for the long delay on the review.

cseickel:
Oops, this should have been targeting `main` instead of the release branch. I missed that one and merged it to `2.x` accidentally, I'll try to move it to main now...

PR #852, opened 2023-04-08
fix(source_selector): fix bug with tabs_layout

### LINKED ISSUE #848: `tabs_layout="start"` is broken
I'm currently on branch v2.x at https://github.com/nvim-neo-tree/neo-tree.nvim/commit/2b2f74828eeb02cf29d6b21aa32eedadadc94ca7

With the config:
```lua
require"neo-tree".setup {
  source_selector = {
    winbar = true,
    tabs_layout = "start",
  }
}
```
I get:
Screenshot_20230407_222246

The winbar is correctly set to `%{%v:lua.require'neo-tree.ui.selector'.get()%}`
and that function returns `"%#NeoTreeTabInactive#%4009@v:lua.___neotree_selector_click@%…%<%0@v:lua.___neotree_selector_click@"` (missing all tabs?)

nhat-vo:
Fixes #848. 

The function `text_layout` doesn't work with the whole tab representation, which relies on `vim.fn.strdisplaywidth`. It seems that this function considers the tab representation string literally, and gives a length of more than 400 for just 3 tabs. The `text_layout` then truncates the tab representation, which results in weird values.

To fix this, I moved the `text_layout` logic to truncate `tab.text` instead.

### DIFF
--- a/lua/neo-tree/ui/selector.lua
+++ b/lua/neo-tree/ui/selector.lua
@@ -388,7 +388,6 @@ M.get_selector = function(state, width)
           tab.tab_hl,
           calc_click_id_from_source(winid, tab.index)
         )
-        .. text_with_hl("", hl_background)
     end
   end
   return return_string .. "%<%0@v:lua.___neotree_selector_click@"
--- a/lua/neo-tree/ui/selector.lua
+++ b/lua/neo-tree/ui/selector.lua
@@ -341,19 +341,36 @@ M.get_selector = function(state, width)
       if width == 0 then
         break
       end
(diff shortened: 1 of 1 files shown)


nhat-vo:
Actually I realized that I made the wrong implementation but haven't got time to redo it, so this shouldn't be merged yet.😓

cseickel:
> Actually I realized that I made the wrong implementation but haven't got time to redo it, so this shouldn't be merged yet.

Can you mark this as a "draft" PR to make it clearer that it's not ready?

nhat-vo:
@cseickel This should be ready for review.

cseickel (approved):

PR #857, opened 2023-04-09
doc: make longer example collapsible

nhat-vo:
Not sure if this is a good idea, but I find the longer example section in the README too long to scroll through, making it hard to get to the next part of the guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,10 @@ Press `?` in the Neo-tree window to view the list of mappings.
 ## Quickstart
 
 #### Longer Example for Packer:
+<details>
+  <summary>
+    Click to view
+  </summary>
   
 ```lua
 use {
@@ -334,7 +338,8 @@ use {
     end
 }
 ```
-
+</details>
+  
 _The above configuration is not everything that can be changed, it's just the
 parts you might want to change first._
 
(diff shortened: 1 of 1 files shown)


miversen33:
I wonder if instead of doing that, we opt for a Table of contents at the top of the README? 

I don't have any strong opinions one way or the other, IMO this is also fine. I'm just not sure which is "better", if either really are. I could also see users not realizing there is a configuration shown and us having to point them to "click the expand button the readme to see the config" (regardless of it literally saying that on the readme)

If we do go with this, I would change the verbiage from "Click to View" to something more specific. Like "Click to view Packer.nvim example configuration"

Again though, I don't have strong feelings in any direction here lol

nhat-vo:
@miversen33 Thanks for the comments 😁. IMO the README file serves more as an overview of the project, and detailed configs should be put into the help docs and the wiki. In that sense, I think it is better to hide detailed configs like this as collapsible or move it to the wiki, so that's why I create this PR. 

I agree with your point on adding a ToC, but github has a good builtin one already, and (maybe I'm biased) I think README is generally viewed through github.

image

cseickel:
I think I like it because too many people just copy and paste the whole thing and then don't change anything, which they really shouldn't. In that way it might prevent some of that by not having it immediately visible.

On the other hand, some portion of users will never see it and this may produce some extra issues... Let me think about this a little.

pynappo (approved):

Issue #859, opened 2023-04-09
Inconsistent view of elements across different sources


ghostbuster91:
First of all, thanks for this great plugin. I really think that it makes a lot of sense to have such generic view for tree-like data structures. I have plenty of ideas what to integrate next but first I need to get familiar with the basics, hence my migration from nvim-tree. 

I modified `file` and `directory` renderers slightly to my liking but I noticed that my modifications look inconsistent across different sources. 

File source:
![image](https://user-images.githubusercontent.com/5662622/230792417-37739728-4b59-4182-99a6-7390a506b267.png)

Buffer source:
![image](https://user-images.githubusercontent.com/5662622/230792443-346dcb85-4b8b-4065-8aea-8e7c31a4f3fc.png)

Git source:
![image](https://user-images.githubusercontent.com/5662622/230792457-fb5dc8b8-f86b-4228-894e-60e51c9d9345.png)

Code for renderes:
```lua
renderers = {
            directory = {
                { "indent" },
                { "icon" },
                { "current_filter" },
                {
                    "container",
                    content = {
                        { "name",       zindex = 10 },
                        { "clipboard",  zindex = 10 },
                        { "git_status", zindex = 10, hide_when_expanded = true },
                        {
                            "diagnostics",
                            errors_only = true,
                            zindex = 20,
                            align = "right",
                            hide_when_expanded = true
                        },
                    },
                },
            },
            file = {
                { "indent" },
                { "icon" },
                {
                    "container",
                    content = {
                        { "name",        zindex = 10 },
                        { "clipboard",   zindex = 10 },
                        { "bufnr",       zindex = 10 },
                        { "git_status",  zindex = 10 },
                        { "modified",    zindex = 20, align = "right" },
                        { "diagnostics", zindex = 20, align = "right" },
                    },
                },
            },
        }
```

I also have a `name` component modified:
```lua
    name = {
       use_git_status_colors = false,
    }
```

neotree rev: 205184aa0e0f08e8a1249d9bb37b45bae85f01b9

miversen33:
The only difference I can see here is how the directory structure is displayed on the file source vs the git and buffer source. Am I missing something?

ghostbuster91:
Sorry, my description was incomplete. 

Here is a list of differences:
1. As you noticed the directory structure differs between the git source and buffer source. In the buffer source multiple directories are collapsed into a single node, while in git source every directory is its own node
[...]

PR #864, opened 2023-04-11
fix(components): fix `use_git_status_colors` config and component's padding

nhat-vo:
Fixes two issues as highlighted in #859:

- [x] `buffers` and `git_status` sources are not using `use_git_status_colors`
- [x] Missing padding. We can move padding logic to renderers instead of making components handle themselves.

For the missing padding, I have added the padding logic to `renderer` and `container` when they render their components. To avoid spurious padding caused by fading logic, I have added a `no_padding` flag to the return results of the components. I think there is also the choice of directly returning a `NuiLine` to enforce this behavior, but I ended up using `no_padding` for simplicity. It might be interesting to add this `NuiLine` option in the future, though.

In implementing this, I noticed that the truncation logic in `container` uses the lua string function (e.g. `string.sub()`) instead of  `vim.fn.strcharpart` and its counterparts. Therefore, in some cases, the icons are rendered incorrectly, and the truncated length does not correspond to what is expected. I am unsure whether this is actually the intended behavior, so I only fixed the `truncate_layer_keep_right` and not its counterpart `truncate_layer_keep_left`.

The PR is good for now, so it should be ready for merge. For the truncation logic, I think it is better to fix the `keep_left` and add additional tests in another PR.


cseickel (approved):
This looks great, and thanks so much for adding tests!

pysan3:
Hi @cseickel @nhat-vo 

This PR seems to break when `config.default_component_config.indent.indent_size == 1`.
I noticed this when digging into #924.

## Issue

This happens only and only if value is exactly 1. See the indent of `./src/B/{a,b}.js`.

- `= 0`

image

- `= 1`

image

- `= 2` (`>= 2` all works)
image

## Expected

When the indent is set to 1, the indent should only show `indent_marker` (or `last_indent_marker`) so that the indent width will be 1.
- `== 1`
image

## Cause

This PR forces at least one space between each component (if it is not a space). This automatically adds a space after `indent_marker` (`│`) (to be exact, it adds padding before `icon` since the char before it is not a space) which forces the indent width to become 2.

This is caused by these lines.

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8c89efb93b8383666a6f989762c2f610f5f2c930/lua/neo-tree/ui/renderer.lua#L403-L427

To test this, if I put `indent_marker` after `B/` dir, the indents align when `indent_size == 1`, although a padding is added before icons.
image

If I'm not mistaken, we actually don't need any of the padding logic here since it is already calculated inside each component as you mentioned (for example where `no_padding` are set), and is actually corrupting the logic for indents.

```lua
  for _, component in ipairs(renderer) do
    local component_data, component_wanted_width =
      M.render_component(component, item, state, remaining_cols)
    local actual_width = 0
    if component_data then
      for _, data in ipairs(component_data) do
        if data.text then
          actual_width = actual_width + vim.api.nvim_strwidth(data.text)
          line:append(data.text, data.highlight)
          remaining_cols = remaining_cols - vim.fn.strchars(data.text)
        end
      end
    end
    component_wanted_width = component_wanted_width or actual_width
    wanted_width = wanted_width + component_wanted_width
  end
```

This code generates the screenshot I posted in `# Expected` section.
However, I do not know if this passes the tests added in this PR.

Could you elaborate why you added this code here @nhat-vo ?

cseickel:
@pysan3 Ensuring there is a padding between components is actually a difficult thing when you leave it to the components themselves because teh components do not know in what order they should be placed and whether that padding should be on the right or left hand side.

The second code block you posted:
```lua
  for _, component in ipairs(renderer) do
    local component_data, component_wanted_width =
      M.render_component(component, item, state, remaining_cols)
    local actual_width = 0
    if component_data then
      for _, data in ipairs(component_data) do
        if data.text then
          actual_width = actual_width + vim.api.nvim_strwidth(data.text)
          line:append(data.text, data.highlight)
          remaining_cols = remaining_cols - vim.fn.strchars(data.text)
        end
      end
    end
    component_wanted_width = component_wanted_width or actual_width
    wanted_width = wanted_width + component_wanted_width
  end
```
... is in the `container` component and it is calculating width in order to decide how to right align the right hand components and what needs to be cut off if there is not enough space. It also determines where the right hand column is if the tree is in a split window and not a sidebar.

I think the solution here is to just add an exception and not add any spaces if the previous component was the indent component.

PR #868, opened 2023-04-12
fix(setup): save and restore foldcolumn option

### LINKED ISSUE #867: Foldcolumn getting overriden and not restored
Similar to #684, I noticed that for some reason `neo-tree` is overriding `foldcolumn` and not restoring it. Here is a minimal `init.lua` file that is able to replicate this. All it does is set up neo-tree with `netrw` hijacking and then sets `foldcolumn=1`.

```lua
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--branch=stable",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
  {
    dependencies = { "MunifTanjim/nui.nvim", "nvim-lua/plenary.nvim" },
    opts = {
      filesystem = {
        hijack_netrw_behavior = "open_current",
      },
    },
  },
})

vim.opt.foldcolumn = "1"
```

### Steps to Replicate

1. Use this minimal `init.lua`
2. run `nvim` and see that `:set foldcolumn` shows it's set to `1` which is what we want
3. Close Neovim and rerun with the `netrw` hijacking: `nvim .`
4. See `:set foldcolumn` shows `0` which makes sense in the file browser, select a file and see that `:set foldcolumn` is still showing `0`

mehalter:
Fixes #867

I've gone through the `neo-tree` source code and nowhere does it set `foldcolumn` so I'm honestly not sure when/where it's getting set to `0`, but this does resolve the issue perfectly. Let me know what you think!

### DIFF
--- a/lua/neo-tree/setup/init.lua
+++ b/lua/neo-tree/setup/init.lua
@@ -105,6 +105,7 @@ local store_local_window_settings = function(winid)
   prior_window_options[tostring(winid)] = {
     cursorline = vim.wo.cursorline,
     cursorlineopt = vim.wo.cursorlineopt,
+    foldcolumn = vim.wo.foldcolumn,
     wrap = vim.wo.wrap,
     list = vim.wo.list,
     spell = vim.wo.spell,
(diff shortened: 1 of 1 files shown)


cseickel:
> I've gone through the `neo-tree` source code and nowhere does it set `foldcolumn` so I'm honestly not sure when/where it's getting set to `0`, but this does resolve the issue perfectly

It's possible that it is set by Nui as well.

cseickel (approved):
This looks good, thanks for contributing! In the future, please use the `main` branch as your target when creating a PR.

Issue #822, opened 2023-03-21
Error executing vim.schedule lua callback in the latest version from nui.nvim


lucax88x:
Since yesterday I'm getting the following error:

<img width="1719" alt="Screenshot 2023-03-21 at 19 08 16" src="https://user-images.githubusercontent.com/6294464/226702182-826ac097-d906-4412-8e35-fb4c83a89f7b.png">

I guess there's some breaking changes from nui?

cseickel:
The most important part of that message is cutoff. Can you post the full message, particularly the top line?

I just updated this morning and I'm not getting that error, so I think it's something other than a breaking change in nui.

lucax88x:
for some reason notify is showing a different error.. interesting.

I had to remove it and noice in order to show the real message..

[Neo-tree ERROR] debounce  filesystem_navigate  error:  ...zi/.local/share/nvim/lazy/nui.nvim/lua/nui/tree/init.lua:185: missin
g bufnr

benplante:
Hi, I'm noticing this issue as well. It happens if I open nvim on a directory and if I run the 'Neotree' command with no arguments. I found that the commit fee76c06e2bfd7d3dbb39b197fc208f6d5c0410c introduced the error message for me though since that commit only adds minimal error checking I'm guessing the issue is somewhere else. Minimal config to reproduce is below.

`packer.lua`
```
local fn = vim.fn
local packer_install_path = fn.stdpath('data') ..'/site/pack/packer/start/packer.nvim'
local packer_compile_path = fn.stdpath('config') .. '/lua/packer_compiled.lua'

if fn.empty(fn.glob(packer_install_path)) > 0 then
  packer_bootstrap = fn.system({ 'rm', '-f', packer_compile_path })
  fn.system({'git', 'clone', '--depth', '1', 'https://github.com/wbthomason/packer.nvim', packer_install_path})
  print "Installing packer, close and reopen nvim"
  vim.cmd [[packadd packer.nvim]]
end

if fn.filereadable(packer_compile_path) == 1 then
  require('packer_compiled')
end

vim.cmd [[
  augroup packer_user_config
    autocmd!
    autocmd BufWritePost packer.lua source <afile> | PackerSync
  augroup end
]]

local ok_packer, packer = pcall(require, 'packer')
if not ok_packer then
  return
end
local ok_util, util = pcall(require, 'packer.util')
if not ok_util then
  return
end

packer.init({
[...]

PR #874, opened 2023-04-15
fix: set correct winid when window.position = current

### LINKED ISSUE #845: Neotree setting position=current is broken
Hi! Just wanted to report that since about a couple of weeks ago Neotree breaks if it's declared to open as part of the existing buffer (Netrw-like)`window = { position = "current" }`. Here's the error I'm getting when trying to run `:Neotree toggle` after that.

CleanShot 2023-04-05 at 09 35 47@2x

Curiously, running `:Neotree show position=current` works fine...🤔

Steps to reproduce the issue:
1. Install Neotree + nui dependencies
2. In the settings set the following:
```lua
require("neo-tree").setup({
  window = {
    position = "current"
  }
})
```
3. Run `:Neotree toggle`

benplante:
Fixes #845 and #822

`:Neotree` and the default `hijack_netrw` behaviour would fail if `config.window.postion = current`. It seems that both `execute` in`command/init.lua` and `hijack` in `setup/netrw.lua` didn't set the correct winid in state when `config.window.position = current`. Changes to fix this are minimal and just involve getting the window position from the config.

This is my first PR to open source 🥳 and I'm quite new to neovim plugin development, any feedback would be appreciated.

### DIFF
--- a/lua/neo-tree/command/init.lua
+++ b/lua/neo-tree/command/init.lua
@@ -72,7 +72,8 @@ M.execute = function(args)
 
   -- Now get the correct state
   local state
-  if args.position == "current" then
+  local requested_position = args.position or nt.config[args.source].window.position
+  if requested_position == "current" then
     local winid = vim.api.nvim_get_current_win()
     state = manager.get_state(args.source, nil, winid)
   else
--- a/lua/neo-tree/setup/netrw.lua
+++ b/lua/neo-tree/setup/netrw.lua
@@ -5,14 +5,16 @@ local command = require("neo-tree.command")
 local M = {}
 
 local get_position = function(source_name)
-  local pos = utils.get_value(M, "config." .. source_name .. ".window.position", "left")
+  local nt = require("neo-tree")
+  local default_pos = utils.get_value(nt.config, "window.position", "left", true)
+  local pos = utils.get_value(nt.config, source_name .. ".window.position", default_pos, true)
   return pos
 end
 
 M.get_hijack_netrw_behavior = function()
   local nt = require("neo-tree")
   local option = "filesystem.hijack_netrw_behavior"
-  local hijack_behavior = utils.get_value(nt.config, option, "open_default")
+  local hijack_behavior = utils.get_value(nt.config, option, "open_default", true)
   if hijack_behavior == "disabled" then
     return hijack_behavior
   elseif hijack_behavior == "open_default" then
--- a/lua/neo-tree/setup/netrw.lua
+++ b/lua/neo-tree/setup/netrw.lua
@@ -6,7 +6,7 @@ local M = {}
 
 local get_position = function(source_name)
   local nt = require("neo-tree")
-  local pos = utils.get_value(nt.config, source_name .. ".window.position", default_pos, true)
+  local pos = utils.get_value(nt.config, source_name .. ".window.position", "left", true)
   return pos
 end
 
--- a/lua/neo-tree/setup/netrw.lua
+++ b/lua/neo-tree/setup/netrw.lua
@@ -5,7 +5,9 @@ local command = require("neo-tree.command")
 local M = {}
 
 local get_position = function(source_name)
-  local pos = utils.get_value(M, "config." .. source_name .. ".window.position", "left", true)
+  local nt = require("neo-tree")
+  local default_pos = utils.get_value(nt.config, "window.position", "left", true)
+  local pos = utils.get_value(nt.config, source_name .. ".window.position", default_pos, true)
   return pos
 end
 
--- a/lua/neo-tree/setup/netrw.lua
+++ b/lua/neo-tree/setup/netrw.lua
@@ -5,9 +5,7 @@ local command = require("neo-tree.command")
 local M = {}
 
 local get_position = function(source_name)
-  local nt = require("neo-tree")
-  local default_pos = utils.get_value(nt.config, "window.position", "left", true)
-  local pos = utils.get_value(nt.config, source_name .. ".window.position", default_pos, true)
+  local pos = utils.get_value(M, "config." .. source_name .. ".window.position", "left", true)
   return pos
 end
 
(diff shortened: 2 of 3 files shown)


lua/neo-tree/setup/netrw.lua
```diff
@@ -5,14 +5,16 @@ local command = require("neo-tree.command")
 local M = {}
 
 local get_position = function(source_name)
-  local pos = utils.get_value(M, "config." .. source_name .. ".window.position", "left")
+  local nt = require("neo-tree")
+  local default_pos = utils.get_value(nt.config, "window.position", "left", true)
+  local pos = utils.get_value(nt.config, source_name .. ".window.position", default_pos, true)
```
cseickel, line 10: This part should not be necessary as the global default should have already been merged down to the source config by the initial call to `setup()`. If that does not currently work, it should be fixed here:
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/039aff445280dd4ffce3c4b97e25d7962c0bcfd4/lua/neo-tree/setup/init.lua#L572-L577

cseickel (commented):
Thanks for contributing! The only issue, which I have fixed for you,  is that all PRs should be targeting the `main` branch instead of the release branch. Other than that, there is just one comment.

lua/neo-tree/setup/netrw.lua
benplante: It seems that `M.config` in `netrw.lua` is never set to the global config, thus why getting position from `nt.config` worked. My latest commit changes that in the `setup()` function which also fixes netrw hijack, though I'm really not sure if this is the right solution.

https://github.com/nvim-neo-tree/neo-tree.nvim/pull/874/files#diff-3032d6dddfe6b408e380eae368f03366da4adcc797cc2b8e05a27420573efa72

cseickel: Actually, the netrw module is not a source at all and should not have any config. the config belongs to the sources, such as `filesystem`, `buffers`, `git_status`, etc. That is what is being pulled by:
```
local pos = utils.get_value(nt.config, source_name .. ".window.position", "left", true)
```
and that section of code should not have been touched at all. The global config is already being applied to the source config in the lines of code that I quoted originally. **If** that code has a bug, it should be fixed at this location: 
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/039aff445280dd4ffce3c4b97e25d7962c0bcfd4/lua/neo-tree/setup/init.lua#L572-L577

I would double check that though, because I don't think there is any bug to be fixed there.

lua/neo-tree/command/init.lua
```diff
@@ -72,7 +72,8 @@ M.execute = function(args)
 
   -- Now get the correct state
   local state
-  if args.position == "current" then
+  local requested_position = args.position or nt.config[args.source].window.position
+  if requested_position == "current" then
```
cseickel, line 76: Just to clarify, I think that these two lines are the bug fix, and everything else should be reverted to the original code and left unchanged.

benplante: Actually with just that fix in place it leaves the netrw hijack broken.

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/039aff445280dd4ffce3c4b97e25d7962c0bcfd4/lua/neo-tree/setup/netrw.lua#L83-L95

Logging here shows that `should_open_current` is false and control goes to the `hijack_netrw: opening default` block.

cseickel: OH! I get it now. The problem is that the `get_position` method was not updated when it was refactored out of the `setup/init/lua` module into the the separate `netrw.lua` module, so it has just been broken the whole time.

What you did works but it's not correct. The `netrw.get_position` method should either be changed to:
```lua
local get_position = function(source_name)
  local nt = require("neo-tree")
  local pos = utils.get_value(nt, "config." .. source_name .. ".window.position", "left", true)
  return pos
end
```
or that function should just be completely removed and the above code inlined to the one place that it is actually called from.

benplante: I've reverted the previous commit. With the snippet above, the netrw hijack won't respect the global `config.window.position` when the `config.filesystem.window.position` is unset and `netrw_hijack_behaviour = open_default`. Is this the expected behaviour?

cseickel: > config.window.position when the config.filesystem.window.position is unset and netrw_hijack_behaviour = open_default

That means there is a problem in the setup function that needs to be fixed. When `config.filesystem.window.position` should never be unset, if the user doesn't specify a value then it should already be set to the global value before any code runs. If that is not working correctly, it could be causing other bugs as well and should be fixed.

> Also is it preferable to overwrite history to remove a bad commit in a PR or just revert and push that?

It doesn't matter, whatever you think is easiest to follow. I tend to keep adding commits and only force push when rebasing to include changes merged to the main branch. In the end it will be squashed to a single commit when it is merged.

benplante: Apologies, it actually was set correctly and getting the global config win position wasn't necessary. I've pushed another commit to fix that. Thank you for hand-holding my way through this :)

lua/neo-tree/setup/netrw.lua
```diff
@@ -5,14 +5,15 @@ local command = require("neo-tree.command")
 local M = {}
 
 local get_position = function(source_name)
-  local pos = utils.get_value(M, "config." .. source_name .. ".window.position", "left")
+  local nt = require("neo-tree")
+  local pos = utils.get_value(nt.config, source_name .. ".window.position", default_pos, true)
```
cseickel, line 9: It doesn't actually have any effect with how this is used right now, but I think the `default_pos` variable is undefined here and should be replaced with `"left"`.

benplante: Right, fixed that now.

cseickel (approved):
Looks good, thanks!

Issue #873, opened 2023-04-15
Neo-tree source selector crashes with user provided external sources


miversen33:
Related: https://github.com/miversen33/netman.nvim/issues/155

When providing a "third party" source, the source selector appears to fail on loading the state of that source.

The specific error being thrown is
```
E5108: Error executing lua ...nvim/lazy/neo-tree.nvim/lua/neo-tree/sources/manager.lua:39: attempt to index local 'state' (a nil value)
stack traceback:
        ...nvim/lazy/neo-tree.nvim/lua/neo-tree/sources/manager.lua:39: in function 'create_state'
        ...nvim/lazy/neo-tree.nvim/lua/neo-tree/sources/manager.lua:111: in function 'get_state'
        ...re/nvim/lazy/neo-tree.nvim/lua/neo-tree/command/init.lua:79: in function 'execute'
        ...are/nvim/lazy/neo-tree.nvim/lua/neo-tree/ui/selector.lua:384: in function <...are/nvim/lazy/neo-tree.nvim/lua/neo-tree/ui/selector.lua:373>
```

That gibberish may not make a ton of sense so below I will attempt to explain what I am seeing and what I believe the problem is.

-----

In order to provide a third party source, a user must specify that source in their neo-tree configuration. This is usually done as follows
```lua
require("neo-tree").setup({
    sources = {
        "path.to.custom.source", -- The third party source we care about
    }
})
```

Following the setup path, this module is checked [here](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/0a2e518d3169aeb62d544329e6aa1a1925d1e0fd/lua/neo-tree/setup/init.lua#L527-L543)
```lua
    else
      -- fully qualified module name
      -- or just a root level module name
      is_external_ns, module = pcall(require, source)
      if is_external_ns then
        name = module.name or name
        all_sources[name] = source
      else
        log.error("Source module not found", source)
        name = nil
      end
    end
    if name then
      default_config[name] = module.default_config or default_config[name]
      table.insert(all_source_names, name)
    end
  end
```

The important bit to note here is how `name` is being populated and used later.
If not provided, name defaults to the last bit of the module path (though this is just to clarify what name will be here, it doesn't have a ton of importance).
Name default source: https://github.com/nvim-neo-tree/neo-tree.nvim/blob/0a2e518d3169aeb62d544329e6aa1a1925d1e0fd/lua/neo-tree/setup/init.lua#L515-L516

```lua
[...]

PR #876, opened 2023-04-16
873 source selector shenanigans

miversen33:
This accomplishes 3 things

1) Resolves the issues initially reported in #873 (Source selector now reads names instead of module paths)  
2) Corrects potential error in `default_source` being nil (and tosses a warning to the user in the event that they do set it to nil)  
3) Adds a new attribute that sources can return called `display_name`. This is then interpreted into the source selectors "tab" name. If this attribute isn't provided, the source's name is used instead. These can still currently be overridden by the `config.source_selector.tab_labels`. I have also updated the internal sources to use this `display_name` instead of relying on the configuration to provide it for them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,6 @@ on:
       - v2.x
       - v3.x
   pull_request:
-    branches:
-      - main
 
 jobs:
   stylua-check:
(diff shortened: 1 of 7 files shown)


miversen33:
I don't really know where to document the addition of the new `display_name` attribute, I didn't see a good place in the documentation to put it. If @cseickel  can tell me where they want that documented, I will get that documentation added as well.

miversen33:
A nice little screenshot :)
image

The code behind the dummy source is super simple

```lua
local name = "dummy"
if vim.g.neo_tree_dummy_extension_no_name then
    name = nil
end

local commands = {}
require("neo-tree.sources.common.commands")._add_common_commands(commands)

local M = {
    name = name,
    display_name = 'ﭧ Dummy',
    components = {},
    commands = commands,
    navigate = function(state)
        require("neo-tree.ui.renderer").show_nodes({}, state)
    end,
    setup = function() end,
}

return M
```

cseickel (approved):

PR #877, opened 2023-04-16
Configurable source selector entries

miversen33:
RE: #479

Replaces `source_selector.tab_labels` with `source_selector.sources`.

Why? This allows users to have **much** more choice over the things that are displayed in their source selector. Previously, all sources in Neo-tree were added to the source selector and there was no way to remove any of them. This will get especially painful as Neo-tree grows and more developers write sources for it.

Additionally I added some logic to ensure that if `source_selector` is active, (IE one of the display bar options is enabled), that the first option in source selector is set to the `default_source` so that when users execute `:Neotree`, their source selector properly displays the first item in the sources.

Lastly, I fixed a few logs that had invalid function names in them, and updated the name of a mispelled variable because it annoyed me lol.

Please let me know your feedback!

Tagging users from #479 that participated as that PR was closed:
@cseickel @mrbjarksen @pysan3 @mehalter

### DIFF
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -771,7 +771,7 @@ source to add to the bar. A second optional `display_name` key can be
 provided to modify how you wish that source to appear in the bar. It is
 safe to add sources that do not exist, they will simply be omitted from
 the bar if they cannot be found.
-NOTE: If `source_selector` is enabled (via `winbar=true` or `statusbar=true`)
+NOTE: If `source_selector` is enabled (via `winbar=true` or `statusline=true`)
 then the `default_source` will be updated to be the first entry of
 `sources`.
 
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -729,12 +729,6 @@ type notations.
                     display_name = "  Git "                          -- string | nil
                 },
             },
-            tab_labels = {                                            -- table
-                filesystem = "  Files ",                             -- string | nil
-                buffers =    "  Buffers ",                           -- string | nil
-                git_status = "  Git ",                               -- string | nil
-                diagnostics = " 裂Diagnostics " ,                     -- string | nil
-            },
             content_layout = "start",                                 -- string
             tabs_layout = "equal",                                    -- string
             truncation_character = "…",                               -- string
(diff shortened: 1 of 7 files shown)


miversen33:
I might consider this a breaking change as even though I added the migration path from `source_selector.tab_labels` to `source_selector.sources` I did still remove `tab_labels`.

miversen33:
Also, I am not terribly keen on `source_selector.sources` so if anyone has a _better_ name for that variable, lemme know! Maybe something like `source_selector.display_sources`? Idk, `sources` made the most sense when making this but it might get confusing as there is also a top level `sources` key that is going to have different values than `source_selector.sources`.

doc/neo-tree.txt
```diff
@@ -715,6 +715,20 @@ type notations.
...
+                {
+                    source = "buffers",                               -- string
+                    display_name = "  Buffers                        -- string | nil
+                },
+                {
+                    source = "git_status",                            -- string
+                    display_name = "  Git "                          -- string | nil
+                },
+            },
             tab_labels = {                                            -- table
```
cseickel, line 732: I assume that the `tab_labels` are kept to avoid a breaking change. I would mark them as deprecated though.

lua/neo-tree/setup/init.lua
```diff
@@ -636,6 +647,25 @@ M.merge_config = function(user_config, is_auto_config)
     M.config.git_status_async = false
   end
 
+  -- Validate that the source_selector.sources are all available and if any
+  -- aren't remove, remove them
```
cseickel, line 651: Typo:
```suggestion
  -- aren't, remove them
```

doc/neo-tree.txt
miversen33: Actually I forgot to remove them! I didn't see deprecation labels in the readme so I intended to yoink this! If you would prefer I mark it as deprecated, I can do so though I don't know how exactly to do that (following the readme) as I don't see any other deprecated options in the readme.

cseickel: No, actually you're right, it should just be removed.

cseickel (approved):

cseickel:
> Also, I am not terribly keen on `source_selector.sources` so if anyone has a _better_ name for that variable, lemme know!

I think it's a good name.

Issue #828, opened 2023-03-26
Can not open file on first <cr> in lazy load.


mtfcd:
**condition**: Using lazy.nvim and lazy load neotree. use nvim open a directory like this `nvim .`
**behavior**: The first time toggle neotree, and the first time open a file, it will not show the file, only a empty split window. The second time open a file it will show the file normally.

barreiroleo:
I'm pretty sure it's an old bug. When you open a file after starting nvim in a directory, you get the following error;
```
[Neo-tree ERROR] Error opening file: vim/_editor.lua:0: nvim_exec2(): Vim(edit):E201: *ReadPre autocommands must not change current buffer
```
_Sorry if the error doesn't match, I've had to search and translate_

You have to close the buffer and reopen it to get the correct content.
![image](https://user-images.githubusercontent.com/48270301/230757895-5846b842-580c-4063-af41-99a93d4affcf.png)
![image](https://user-images.githubusercontent.com/48270301/230757949-8adfc2d0-dfcd-48a8-b758-7a0eb3e9d8e3.png)

Oscarteg:
I am experiencing the same bug. Not sure why this happens. Is there a fix? @barreiroleo @mtfcd

barreiroleo:
Not that I am aware of. The most usable I found is to set "hijack_netrw_behavior" to "open_current" (read the docs), which is quite weird in some workflow. Neotree toggle mapping causes duplicate file explorer for example.
I wish the "open_default" to work fine. I tried to find a relevant autocmd in the source but I was unlucky.

miversen33:
I am not able to recreate this issue. Could someone provide a minimal lazy config that I can use to reproduce this? 

Below is a very simple lazy config that allows me to open a directory with `nvim .` and open a file by pressing `<CR>`
```lua
        {
            "nvim-neo-tree/neo-tree.nvim", -- File Explorer
            branch = "v2.x",
            dependencies = {
                "nvim-lua/plenary.nvim",
                "kyazdani42/nvim-web-devicons", -- not strictly required, but recommended
                "MunifTanjim/nui.nvim",
            },
            config = {}
        }
```

miversen33:
Thank you! I will see if I can reproduce the issue with your config

lcrownover:
I know this isn't really a fix to the overall issue but setting `lazy = false` for this plugin is a workaround.

miversen33:
[...]

PR #907, opened 2023-04-30
fix(hijack): Hijack on setup for lazy load

pysan3:
closes: #828 Can not open file on first `<cr>` in lazy load.

[My comment](https://github.com/nvim-neo-tree/neo-tree.nvim/issues/828#issuecomment-1529131383) explains the detail of this bug.

With this fix, when the setup function is called, neo-tree examines current buffer once to check if its a dir buffer or not.

Usually, this check is done in an autocmd [defined here](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8d485aab32da9b8841d4af977f992b82b14af469/lua/neo-tree/setup/init.lua#L150) but this is too late in the case of lazy loading.

In order to make neo-tree work properly, we need to guarantee that the alternating buffer `"#"` must NOT be a _`dir buffer`_, and this hijack in setup function guarantees that even on lazy loading.

### Please Test It

I am pinging people over on the issue thread. I kindly ask you to test my change out if it works in your environment.
@mtfcd @barreiroleo @Oscarteg @miversen33 @lcrownover

```lua
return {
  "pysan3/neo-tree.nvim",
  branch = "hijack-on-setup-for-lazy-load",
  lazy = true,
  dependencies = {
    { "MunifTanjim/nui.nvim" },
    { "nvim-tree/nvim-web-devicons" },
  },
  cmd = { "Neotree" },
  keys = {
    { "<Leader>e", "<Cmd>Neotree toggle<CR>", remap = false, silent = true, desc = "<Cmd>Neotree toggle<CR>" },
  },
  opts = {} -- YOUR SETUP
}
```

### Old Behavior

record

- Please ignore `NVIM_LIGHT_ENV`. It's doing nothing.
- I'm pressing `<CR>` two times, one for each launch of neo-tree.

### DIFF
--- a/lua/neo-tree.lua
+++ b/lua/neo-tree.lua
@@ -246,6 +246,7 @@ M.setup = function(config, is_auto_config)
   local netrw = require("neo-tree.setup.netrw")
   if not is_auto_config and netrw.get_hijack_netrw_behavior() ~= "disabled" then
     vim.cmd("silent! autocmd! FileExplorer *")
+    netrw.hijack()
   end
 end
 


barreiroleo:
Hi @pysan3, thanks for pinging me

This fixes the bad initialization of root directory but doesn't solve the issue itself. I had to set the Newtr flags for solve the weird behaviour (at least `vim.g.loaded_netrwPlugin = 1`, I don't have the newtr disabled by Lazy because gx).

Using main branch with newtr flags (also I needed to close the first neotree window):
 main-newtr

Using the pr with newtr flags:
pr-newtr

Using the pr without newtr flags:
pr-nonewtr

pysan3:
Thanks for testing @barreiroleo !

I'm sorry but could you elaborate your issue? I cannot find anything wrong in your gif.

pysan3:
Thanks for your reply.

I really cannot diagnose your problem.
At least, I'd have to say that it is not (directly) related to the issue we are currently tackling.

Could you open a new issue and paste your configuration?
And does that error not happen when you do the same thing with `lazy = false`?

Anyways, @cseickel in the scope of fixing lazy loading, this PR should be ok to merge.

barreiroleo:
> The PR is good and with the netrwPlugin note (...)

Just to be clear here. At least in my test using lazy, Netrw loads before Neotree and you have to mark Netrw plugin as loaded in order to jump to the Neotree setup. You can't use lazy with netrw enabled. 

Netrw disabled: all are good
pr-lazy-nonetrw
Netrw enabled: Netrw loads first
pr-lazy-netrw

> And does that error not happen when you do the same thing with lazy = false?

In facts, I don't have neo-tree lazy loaded in my config. The error still there. Don't worry, I can live with it.

mtfcd:
@pysan3 I have tested. My problem has gone. Now it's works as expected.

cseickel (approved):
Wow, so much work and explanation for such a simple fix!

pysan3:
Yah, it was quite hard to diagnose and with a lot of try and error. 

Pretty proud that I could simply the fix so much and still make it stable :)

PR #908, opened 2023-05-01
feat(renderer): Ignore never show on empty dirs

### LINKED ISSUE #906: Ignore never show items while grouping dirs
I saw the number of hidden items increased and found it was from system files like .DS_Store" & "thumbs.db".
I added `never_show` to the config to stop them from showing. 
However they are still counted for grouping by `group_empty_dirs` so the dir was not grouped.

pysan3:
closes #906. Please read the issue for detailed explanation.

When a directory contains a `never_show` file (e.g. `.DS_Store` in the examples), those files are counted and prevent `group_empty_dirs` to be executed.

This PR fixes that behavior and ignores `never_show` files in the first place.

FYI, `e.filtered_by.never_show` is set here.

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8d485aab32da9b8841d4af977f992b82b14af469/lua/neo-tree/sources/common/file-items.lua#L114-L148

## Old Behavior

record

## New Behavior

record

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/fs_scan.lua
+++ b/lua/neo-tree/sources/filesystem/lib/fs_scan.lua
@@ -180,13 +180,22 @@ local function get_children_async(path, callback)
   end, 1000)
 end
 
+local function filter_never_shows(paths, context)
+  if paths == nil or #paths < 1 then
+    return paths
+  end
+  return vim.tbl_filter(function(c)
+    return not context.is_a_never_show_file(c.path)
+  end, paths)
+end
+
 local function scan_dir_sync(context, path)
   process_node(context, path)
   local children = get_children_sync(path)
   for _, child in ipairs(children) do
     create_node(context, child)
     if child.type == "directory" then
-      local grandchild_nodes = get_children_sync(child.path)
+      local grandchild_nodes = filter_never_shows(get_children_sync(child.path), context)
       if
         grandchild_nodes == nil
         or #grandchild_nodes == 0
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -1176,7 +1176,7 @@ M.show_nodes = function(sourceItems, state, parentId, callback)
       local scan_mode = require("neo-tree").config.filesystem.scan_mode
       if scan_mode == "deep" then
         for i, item in ipairs(sourceItems) do
-          sourceItems[i] = group_empty_dirs(item)
+          sourceItems[i] = filter_never_shows(group_empty_dirs(item))
         end
       else
         -- this is a lazy load of a single sub folder
@@ -1211,7 +1211,7 @@ M.show_nodes = function(sourceItems, state, parentId, callback)
       for _, item in ipairs(sourceItems) do
         if item.children ~= nil then
           for i, child in ipairs(item.children) do
-            item.children[i] = group_empty_dirs(child)
+            item.children[i] = filter_never_shows(group_empty_dirs(child))
           end
         end
       end
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -1095,13 +1095,24 @@ draw = function(nodes, state, parent_id)
   M.position.restore(state)
 end
 
+local function filter_never_shows(node_tbl)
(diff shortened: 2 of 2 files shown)


pysan3:
Caution: this actually filters out the never show nodes in some occasions from the list, so if we want to implement such thing as `H` to toggle the visibility of never show files (tho I think that's nonsense) it becomes problematic.

cseickel:
I think it is a good idea but my question is whether there is a simpler way to accomplish the same thing. I noticed that you are filtering the `never_show` items out of the collection in many different places. If it is never going to be used for anything, wouldn't it be easier to short circuit as soon is it is decided that it is a never show item and just not create the node at all?

I would try to return here for never show items:
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8d485aab32da9b8841d4af977f992b82b14af469/lua/neo-tree/sources/common/file-items.lua#L118-L125

It might just work, but if there is an issue I would try to handle that function returning nil if it all possible.

cseickel:
> if we want to implement such thing as `H` to toggle the visibility of never show files

That is actually the entire purpose of the `never_show` setting, to create a filter for things that are not shown when you toggle hidden with H.

pysan3:
Ok, let me look into it.

cseickel:
I am not quite following @pysan3. It seems like in every possible situation, we should treat "never show" files as if they did not exist, so being able to short circuit without creating the item at the lowest level possible seems like the correct thing to do.

Can you explain more about what situation it would be a problem to skip creating a "never show" item in the `create_item` function?

As for the error pop-up, that's likely just a bug to fix.

pysan3:
It's so hard to explain code in my second language lol. plz feel free to ask any more questions if you don't get it.

Let's say we have folder structure like this.

```txt
- root/
  - A1/
    - B1/
      - foo.txt
    - B2/
  - A2/
    - B3/
      - bar.txt
```

On first load of the plugin, Neotree does not scan through the whole subfolders but only load root and it's direct children. So in the above case, `root, A1, A2` and does not know the content inside A1 nor A2.

But in order to do the `group_empty_dirs` (to make it look like `- A2/B3/`) from the beginning, the scan mechanism checks the number of children for each dirs just in case.

When the number of children in the dirs is exactly 1 and that grandchild is a folder, neotree goes deeper and go scan B3 as well.

(When scanning children of A1, it stops there cuz there're B1 and B2 (it does not even check if they are files or folders) resulting to only showing `- A1/` in the tree initially.)

For the case of A1, B1 and B2 are only scanned when `<CR>` is pressed at A1 (scan is delayed), so neo-tree does not have any node about B1, B2.

Note that this mechanism only checks the number of children in the folders and does not do any high-level check. 

`create-item` is called after the whole scan, in this example, to create 3 nodes (`root/`, `A1/`, `A2/B3/`). It's not `A2/` _and_ `B3/` but are merged into a single node `A2/B3/` while scanning.

When we have a (what is going to be a `never_show` item at `create_item`) file like below, the scan stops at A3 and we do not have the information of child items until the folder is expanded. But actually what we want is to merge to `A3/B4/` at this moment (when scanning contents of `root`).

```txt
- root/
  - A3/
    - B4/
    - .DS_Store
```

What makes it worse is that there are multiple functions that does this scan `scan_dir_sync`, `scan_dir_async` and there should be more? So there is no silver bullet to change the above action.

> As for the error pop-up, that's likely just a bug to fix.

No, this is expected. Calls to `create_item` is always wrapped in a pcall. Error guard in done by checking success or fail of pcall and the code after that expects the return value to be an item (table) if success.

```lua
-- It is always like this
local suc, item = pcall(create_item, ...)
if suc then
  item.whatever = xxx -- would fail: error: cannot access `whatever` of nil
else
  show_error!!
end

-- this would be error proof, but will still raise the error notification.
local suc, item = pcall(create_item, ...)
if suc and item then
  ...
```

So, to actually not break, we have to make `suc = false` with some kind of raise exception, or assert error. But then, it triggers the `show_error!!` part which is what the notification I posted earlier is.

pysan3:
@cseickel Any updates?

cseickel:
Sorry @pysan3, this one slipped through the cracks and I forgot to follow up!

In the meantime, @ghostbuster91 has been working on a major refactor of the fs scanning functions in #957, so we should probably wait for that to land and then proceed from that point.

PR #909, opened 2023-05-02
fix(icons): replace obsolete Nerd Fonts icons

loichyan:
Replaces obsolete icons which were removed in [the Nerd Fonts v3.0.0](https://github.com/ryanoasis/nerd-fonts/releases/tag/v3.0.0) with same icons (some are not exactly the same, but very similar).

Note: This assumes you are using Nerd Fonts after v2.3.3, check out [nerdfix](https://github.com/loichyan/nerdfix) for more details.


kevinrobayna:
Came here to do this change 😅 thanks for doing this!

wookayin:
Unfortunately some of widely used other plugins lspkind.nvim, nvim-web-devicons already have moved to nerd font v3.0 and already made breaking changes without the compatibility layer. I hate such breaking changes but that somehow happened. I think it'd make sense to switch to the *new* icon unicode characters sooner than later.

If you care not having breaking changes (which is bad!), then we could provide some compatibility option (e.g. some option/flag for using v3-compatible codepoints) so that users can switch easily between two if they haven't migrated yet.

cseickel:
> Unfortunately some of widely used other plugins lspkind.nvim, nvim-web-devicons already have moved to nerd font v3.0 and already made breaking changes without the compatibility layer. 

> If you care not having breaking changes (which is bad!), then we could provide some compatibility option (e.g. some option/flag for using v3-compatible codepoints) so that users can switch easily between two if they haven't migrated yet.

Ugh, what a mess! Luckily every icon is configurable. I think the right thing to do here is to not change the defaults, but have a config prominently displayed in the README that will set all of the affected icons to their 3.o equivalent for users that have already switched.

cseickel:
> Unfortunately some of widely used other plugins lspkind.nvim, nvim-web-devicons already have moved to nerd font v3.0 and already made breaking changes without the compatibility layer. 

> If you care not having breaking changes (which is bad!), then we could provide some compatibility option (e.g. some option/flag for using v3-compatible codepoints) so that users can switch easily between two if they haven't migrated yet.

Ugh, what a mess! Luckily every icon is configurable. I think the right thing to do here is to not change the defaults, but have a config prominently displayed in the README that will set all of the affected icons to their 3.o equivalent for users that have already switched.

The next thing we need to do is to move to a new major release (v3.0) where we will make these changes part of the defaults.

cseickel:
> Unfortunately some of widely used other plugins lspkind.nvim, nvim-web-devicons already have moved to nerd font v3.0 and already made breaking changes without the compatibility layer. 

> If you care not having breaking changes (which is bad!), then we could provide some compatibility option (e.g. some option/flag for using v3-compatible codepoints) so that users can switch easily between two if they haven't migrated yet.

Ugh, what a mess! Luckily every icon is configurable. I think the right thing to do here is to not change the defaults, but have a config prominently displayed in the README that will set all of the affected icons to their 3.o equivalent for users that have already switched.

The next thing we need to do is to move to a new major release (v3.0) where we will make these changes part of the defaults.

cseickel:
> Unfortunately some of widely used other plugins lspkind.nvim, nvim-web-devicons already have moved to nerd font v3.0 and already made breaking changes without the compatibility layer. 

> If you care not having breaking changes (which is bad!), then we could provide some compatibility option (e.g. some option/flag for using v3-compatible codepoints) so that users can switch easily between two if they haven't migrated yet.

Ugh, what a mess! Luckily every icon is configurable. I think the right thing to do here is to not change the defaults, but have a config prominently displayed in the README that will set all of the affected icons to their 3.o equivalent for users that have already switched.

The next thing we need to do is to move to a new major release (v3.0) where we will make these changes part of the defaults.

Can someone make a new PR that includes a section in the README that has instructions on how to configure all affected icons to the 3.o equivalents? I will then just hold this PR for the 3.0 release which I will have to hurry up on.

cseickel:
> Unfortunately some of widely used other plugins lspkind.nvim, nvim-web-devicons already have moved to nerd font v3.0 and already made breaking changes without the compatibility layer. 

> If you care not having breaking changes (which is bad!), then we could provide some compatibility option (e.g. some option/flag for using v3-compatible codepoints) so that users can switch easily between two if they haven't migrated yet.

Ugh, what a mess! Luckily every icon is configurable. I think the right thing to do here is to not change the defaults, but have a config prominently displayed in the README that will set all of the affected icons to their 3.o equivalent for users that have already switched.

The next thing we need to do is to move to a new major release (v3.0) where we will make these changes part of the defaults.

Can someone make a new PR that includes a section in the README that has instructions on how to configure all affected icons to the 3.o equivalents? I will then just hold this PR for the 3.0 release which I will have to hurry up on.

cseickel:
> Unfortunately some of widely used other plugins lspkind.nvim, nvim-web-devicons already have moved to nerd font v3.0 and already made breaking changes without the compatibility layer. 

> If you care not having breaking changes (which is bad!), then we could provide some compatibility option (e.g. some option/flag for using v3-compatible codepoints) so that users can switch easily between two if they haven't migrated yet.

Ugh, what a mess! Luckily every icon is configurable. I think the right thing to do here is to not change the defaults, but have a config prominently displayed in the README that will set all of the affected icons to their 3.o equivalent for users that have already switched.

The next thing we need to do is to move to a new major release (v3.0) where we will make these changes part of the defaults.

Can someone make a new PR that includes a section in the README that has instructions on how to configure all affected icons to the v3 equivalents? I will then just hold this PR for the Neotree 3.0 release which I will have to hurry up on.

cseickel:
The instructions for manually fixing this were merged in #921

Alan-Kuan:
Hi @loichyan,

For the symbol in the display name of "Buffers", I noticed that you chose `󰈙` to replace ``. However, I found `󰈚` a better one, which looks the same as the old one.

I think you can make another commit to update it.

cseickel (approved):

Issue #911, opened 2023-05-02
Error when creating a file on SMB share


RomanoZumbe:
When I create a file on an SMB share from within NeoTree I get the following error message:

`Could not create file <filename you entered>`

The file is created though but is not immediately visible in Neotree and the file is readonly. When I create a file with the `:w` command everything works as expected.

I attached a screenrecord as requested by @pysan3 

https://user-images.githubusercontent.com/16334447/235638252-eeb6d682-3ad4-4cb3-b0a1-bf005010424b.mp4

RomanoZumbe:
It's at the end of the video.  Or do you mean something else? Just in case here is the output shown in the video:

-rwxr-xr-x 1 romano wheel 5  2. Mai 12:07 test2.txt
-r-xr-xr-x 1 romano wheel 0  2. Mai 12:07 test.txt

pysan3:
I couldn't diagnose the problem sadly.

I instead created a PR (#912) to workaround the issue we are having here.

**IF** that PR is merged, you will be able do something like this to force overwrite the permission so that you can work as normal.

```lua
event_handlers = {
  {
    event = "file_added",
    handler = function (destination)
      local uv = vim.loop
      local file_info = uv.fs_stat(destination)
      if not file_info then
        return
      elseif file_info.type == "file" then
        if vim.startswith(destination, "/mnt/share/samba/or/whatever") then
          vim.fn.setfperm(destination, "rw-r--r--") -- read `:h setfperm` for more options
        end
      end
    end,
  }
}
```

For more info, read this discussion as well.
https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/708

RomanoZumbe:
Thank you!

pysan3:
Merged. Closing.

pysan3:
Just wanted to notify that I think it's _properly_ fixed now with #1200.

PR #912, opened 2023-05-02
feat(fs_actions): check file existence on fs_open error

pysan3:
This might be more of a workaround than a fix but in case of working with directories mounted from samba, `fs_open` seems to fail internally after file creation but before assigning file permission.

https://github.com/nvim-neo-tree/neo-tree.nvim/issues/911#issue-1692135414

I could not diagnose the core problem but it seems to fail regardless of neo-tree as shown in the following code which end up creating a file but with permission set to `r-xr-xr-w` (I guess it's more of a samba bug?).

So, to workaround this, I'd like neo-tree to not simply bail out when `fd` is nil [here](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8d485aab32da9b8841d4af977f992b82b14af469/lua/neo-tree/sources/filesystem/lib/fs_actions.lua#L367) but rather actually check the file existence and continue on when file was indeed created.

After that, to solve the samba issue, user can overwrite the permission with `vim.fn.setfperm` in the `FILE_ADDED` event handler. (Tested that this method works with samba files)

Best,

```lua
-- test `fs_open` in samba folder
local loop = vim.loop

local open_mode = loop.constants.O_CREAT + loop.constants.O_WRONLY + loop.constants.O_TRUNC
local dst = "g"
local fd, err = loop.fs_open(dst, "w", open_mode)
assert(not err, err)
```

```txt
E5108: Error executing lua [string ":source (no file)"]:6: EACCES: permission denied: g
stack traceback:
	[C]: in function 'assert'
	[string ":source (no file)"]:6: in main chunk
```

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/fs_actions.lua
+++ b/lua/neo-tree/sources/filesystem/lib/fs_actions.lua
@@ -365,11 +365,16 @@ M.create_node = function(in_directory, callback, using_root_directory)
             + loop.constants.O_TRUNC
           local fd = loop.fs_open(destination, "w", open_mode)
           if not fd then
-            api.nvim_err_writeln("Could not create file " .. destination)
-            return
+            if not loop.fs_stat(destination) then
+              api.nvim_err_writeln("Could not create file " .. destination)
+              return
+            else
+              log.warn("Failed to complete file creation of " .. destination)
+            end
+          else
+            loop.fs_chmod(destination, 420)
+            loop.fs_close(fd)
           end
-          loop.fs_chmod(destination, 420)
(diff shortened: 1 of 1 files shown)


pysan3:
> All checks have failed

Did I do something wrong??

cseickel:
@pysan3 Github was having an outage related to Actions earlier. If you add another commit or amend your last one to trigger the actions to run again it will probably be fine.

pysan3:
Thanks ;)

Added an empty commit.

Now it's a green light

cseickel (approved):
More precise error handling is always good!

PR #938, opened 2023-05-19
add copy path

daog1:
Recently, I used nvim to read the code, and felt that neo-tree is very useful. I suddenly found that there is no path copy function, so I added it.


lua/neo-tree/sources/common/commands.lua
```diff
@@ -480,6 +480,32 @@ M.copy = function(state, callback)
...
   fs_actions.copy_node(node.path, nil, callback, using_root_directory)
 end
+---Copies a node relative path to clipboard.
+---@param state table The state of the source
+M.copy_path = function(state)
+  local node = state.tree:get_node()
+  if node.type == "message" then
+    return
+  end
+  local pwdpath = vim.fn.getcwd()
```
cseickel, line 490: Shouldn't this path be relative to the root of the tree? That may or may not be the same thing as what `vim.fn.getcwd()` will return. I would do it like this:
```suggestion
  local pwdpath = state.path
```

cseickel:
> When will this patch be merged into the main branch?

When the relative path bug is fixed @acsezen.

acsezen:
> > When will this patch be merged into the main branch?
> 
> When the relative path bug is fixed @acsezen.

Thank you @cseickel !

PR #944, opened 2023-05-22
Update README.md

goququ:
HI team, I just tried to find this setting recently and spend some time to find it. Could we place it in documentation?

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,7 @@ use {
         window = {
           position = "left",
           width = 40,
+	  auto_expand_width = false,
           mapping_options = {
             noremap = true,
             nowait = true,


pysan3:
Hi, thanks for your PR but the indentation is wrong and you should be targeting the `main` branch instead.

Please read https://github.com/nvim-neo-tree/neo-tree.nvim/blob/v2.x/CONTRIBUTING.md.

cseickel:
TBH, I don't really want to make `auto_expand_width` an option that's highlighted on the README. It's one of many options that exists but is not advertised on the README. We don't need to list every option there.

This one I would rather keep off the README because I suspect it could occasionally cause infinite loop bugs so I'd rather less people be using it. I also think that the implimentation is now outdated because neovim has finally added a window resize event that we should be using.

PR #946, opened 2023-05-24
feat(ui): hide hidden files irrespective of `hide_gitignored` option

### LINKED ISSUE #929: `hide_gitignored` takes precedence over `hide_by_*` and `never_show*`
It seems that the `hide_gitignored = false` option takes precedence over `hide_by_name`, `hide_by_pattern`, `never_show` and `never_show_by_pattern`, which I believe to be counter-intuitive.

My situation is that I have a javascript project with a .gitignored `node_modules` folder. I want to be able to hide the `node_modules` folder while still keeping other .gitignored files visible.

I also have `.DS_Store` set to `never_show`, but this again doesn't work if `.DS_Store` is .gitignored.

As soon as I remove `node_modules` or `.DS_Store` from the `.gitignore` file they then get hidden.

The issue appears with `hide_by_pattern` and `never_show_by_pattern` as well.

These are the relevant config lines:
```lua
{
  filesystem = {
    filtered_items = {
      hide_dotfiles = false,
      hide_gitignored = false,
      hide_by_name = {
        ".git",
        "node_modules",
      },
      never_show = {
        ".DS_Store",
      },
    },
  }
}
```

I would expect all files in `hide_by_name`, `hide_by_pattern`, `never_show` and `never_show_by_pattern` to be hidden irrespective of the `hide_gitignored` option. This is what happens with `hide_dotfiles` - hiding the `.git` directory works even though I have `hide_dotfiles = false`. 

This is a similar issue to #682.

pysan3:
closes #929

## Note

`show_anyways` is only used in this context (there is no other reference to `show_anyways` other than my fix), so deleted in line 200.

### DIFF
--- a/lua/neo-tree/git/ignored.lua
+++ b/lua/neo-tree/git/ignored.lua
@@ -84,14 +84,14 @@ M.mark_ignored = function(state, items, callback)
   end
 
   local function finalize(all_results)
-    local show_anyway = state.filtered_items and state.filtered_items.hide_gitignored == false
-    log.trace("IGNORED: Comparing results to mark items as ignored, show_anyway:", show_anyway)
+    local show_gitignored = state.filtered_items and state.filtered_items.hide_gitignored == false
+    log.trace("IGNORED: Comparing results to mark items as ignored:", show_gitignored)
     local ignored, not_ignored = 0, 0
     for _, item in ipairs(items) do
       if M.is_ignored(all_results, item.path, item.type) then
         item.filtered_by = item.filtered_by or {}
         item.filtered_by.gitignored = true
-        item.filtered_by.show_anyway = show_anyway
+        item.filtered_by.show_gitignored = show_gitignored
         ignored = ignored + 1
       else
         not_ignored = not_ignored + 1
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -201,6 +201,8 @@ local remove_filtered = function(source_items, filtered_items)
       if not fby.never_show then
         if filtered_items.visible or child.is_nested or fby.always_show then
           table.insert(visible, child)
+        elseif fby.name or fby.pattern or fby.dotfiles or fby.hidden then
+          table.insert(hidden, child)
         elseif fby.show_gitignored and fby.gitignored then
           table.insert(visible, child)
         else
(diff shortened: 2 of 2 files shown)


pysan3:
I'm sorry @cseickel 

Forgot to turn it into a draft. 

I was intending to first get a reply over on the issue and then start working on it.

cseickel:
> I was intending to first get a reply over on the issue and then start working on it.

Ah, well yes, I do think this is something that should be fixed!

pysan3:
Hey @cseickel,

With further exploration, I think the easiest way would be to skip updating `item.filtered_by.gitignored` by short circuiting everything here.

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/e5594d53986b34e584e8afe2ea6ad99d6f6d2105/lua/neo-tree/git/ignored.lua#L86-L102

image

I at first thought this would affect other places since it does not flag `filtered_by.gitignored`, but in fact this is what is done for other `hide_*` configs so I think it is rather a feature? `config.hide_dotfiles == false` makes it never update `filtered_by.hidden` in the first place.

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/e5594d53986b34e584e8afe2ea6ad99d6f6d2105/lua/neo-tree/sources/common/file-items.lua#L139-L146

Theoretically this short circuit will be a breaking change, but I would say that updating `filtered_by.gitignored` when `config.hide_gitignored == false` was unnecessary in the first place.

Would you agree with this fix @cseickel ?

Another option would be to add `if-else`s [here](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/e5594d53986b34e584e8afe2ea6ad99d6f6d2105/lua/neo-tree/ui/renderer.lua#L195) as below. Which however I feel is a bit inconsistent.

image

Which one would you prefer?

pysan3:
@cseickel Any updates?

cseickel:
The part you have to consider is that even when a gitignored item is visible, we still need to be able to mark it as ignored in the `name` and `git_status` components so it shows with teh right font and with the correct status symbol.

Your second proposal is the correct strategy. I don't know for sure if the logic is correct or not (short on time), but it is definitely the right place to handle this.

pysan3:
@cseickel Thanks for the response.

> Your second proposal is the correct strategy

I've updated the code and is ready to be merged :)

cseickel (approved):
You're on a roll @pysan3!

PR #953, opened 2023-05-27
fix: Prevent window from moving when reloading scrolled buffer

### LINKED ISSUE #781: Prevent scrolling on refresh/render
Currently it seems that refreshing neo-tree/re-render will scroll within the neo-tree view with a very large tree, see:

https://user-images.githubusercontent.com/7985687/222311811-3e663541-9eb4-48a7-ac83-1896563161d5.mp4

This is a bit jarring and throws the user off a smooth workflow. 

Would it be possible to prevent that scrolling from happening (basically cursor and buffer stays exactly the same)?

ghostbuster91:
Save current window scroll offset using `vim.fn.winsaveview` when saving
current node position. Then, when restoring current node position,
restore saved window scroll offset using `vim.fn.winrestview`.

This fixes #781 

I also extracted one function to improve readability. I understand that this is a subjective assessment, so it is done as a separate commit and can be easily reverted if needed.

cc @miversen33

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -822,46 +822,8 @@ local set_window_mappings = function(state)
   state.resolved_mappings = resolved_mappings
 end
 
-create_window = function(state)
-  local default_position = utils.resolve_config_option(state, "window.position", "left")
-  local relative = utils.resolve_config_option(state, "window.relative", "editor")
-  state.current_position = state.current_position or default_position
-
-  local bufname = string.format("neo-tree %s [%s]", state.name, state.id)
-  local size_opt, default_size
-  if state.current_position == "top" or state.current_position == "bottom" then
-    size_opt, default_size = "window.height", "15"
-  else
-    size_opt, default_size = "window.width", "40"
-  end
-  local win_options = {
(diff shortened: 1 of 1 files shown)


cseickel (approved):
This looks good to me. Thanks!

lua/neo-tree/ui/renderer.lua
```diff
@@ -621,6 +621,8 @@ M.position = {
         _, state.position.node_id = pcall(node.get_id, node)
       end
     end
+    local win_state = vim.fn.winsaveview()
+    state.position.topline = win_state.topline
```
cseickel, line 625: @ghostbuster91 I've notice recently that when I use the `:Neotree current` command from a terminal, the neotree window will usually be scrolled down so that the last line is the top line. I don't have time to look into it right now, but I would suspect that it is the result of this change.

cseickel: My guess would be that the above two lines should be pushed up into the block above:
```lua
      if success and node then
        _, state.position.node_id = pcall(node.get_id, node)
        local win_state = vim.fn.winsaveview()
        state.position.topline = win_state.topline
      end
```

PR #955, opened 2023-05-27
feat: Add ability to hijack-cursor

### LINKED ISSUE #604: [Feature request] nvim-tree `hijack_cursor` functionality
"Keeps the cursor on the first letter of the filename when moving in the tree."

In my opinion it's little easier (or more "pleasant" to eye) to follow where cursor is when it's always on first letter of dir/file.

When moving cursor with `jk` and open/close nodes with `hl`, with this feature in nvim-tree that feels much better.

https://github.com/nvim-tree/nvim-tree.lua/blob/9e87ee2d6e86f37ff09cb74ec7dcf2ac984a01e9/doc/nvim-tree-lua.txt#L478

ghostbuster91:
This closes #604 

TODO:
- [x] docs
- [x] config
- [x] proper initialization


lua/neo-tree/setup/init.lua
```diff
@@ -728,6 +728,8 @@ M.merge_config = function(user_config, is_auto_config)
 
   local rt = utils.get_value(M.config, "resize_timer_interval", 50, true)
   require("neo-tree.ui.renderer").resize_timer_interval = rt
+  --TODO how to initialize properly
+  require("neo-tree.sources.common.hijack_cursor").setup(M.config, nil)
```
ghostbuster91, line 732: I wasn't able to figure out how this should be initialized. I would be grateful for some guidance here.

lua/neo-tree/sources/common/hijack_cursor.lua
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+end
+
+---Configures the plugin, should be called before the plugin is used.
+---@param config table Configuration table containing any keys that the user
+---wants to change from the defaults. May be empty to accept default values.
+M.setup = function(config, global_config)
+  local modules = { "filesystem", "git_status", "buffers" }
+  for _, module in ipairs(modules) do
```
ghostbuster91, line 30: Is this the proper way to initialize something for all sources?

cseickel: No, because the list of sources is not a static thing and will change according to what extensions a user has. The best place to do it right now would be in this block:

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8c89efb93b8383666a6f989762c2f610f5f2c930/lua/neo-tree/setup/init.lua#L671-L682

lua/neo-tree/setup/init.lua
cseickel: Generally all initialization that is needed happens in setup.merge_config. There's no particular pattern to that except for the sources themselves that have to conform to an api.

lua/neo-tree/sources/common/hijack_cursor.lua
ghostbuster91: Done :+1:

ghostbuster91:
@cseickel Thanks for the quick review. I have addressed your comments. Is there anything left that I should do? Initially I was thinking also about mentioning it in the vim docs, but I see that not all of the flags are there, so I am not sure if I should put it there.

lua/neo-tree/sources/common/hijack_cursor.lua
```diff
@@ -0,0 +1,36 @@
+local events = require("neo-tree.events")
+local manager = require("neo-tree.sources.manager")
+local log = require("neo-tree.log")
+
+local M = {}
+
+local setup_for_module = function(module)
```
cseickel, line 7: I feel like this function name can be named better. Maybe something like `create_handler_for_module`?

lua/neo-tree/sources/common/hijack_cursor.lua
```diff
@@ -0,0 +1,36 @@
...
+  return function()
+    local state = manager.get_state(module)
```
cseickel, line 9: This will actually only give you the state of the sidebar for the current tab. There may also be state objects for other tabs along with many state objects for windows where position = "current".

There is a `manager._for_each_state()` function that you can use for this:
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8c89efb93b8383666a6f989762c2f610f5f2c930/lua/neo-tree/sources/manager.lua#L54

cseickel (commented):
As far as docs go, not everything is in the help file because I was just too lazy to write docs for every single option. The general rule is that anything that can easily be explained with an inline comment in the config doesn't need to be explicitly listed in the help file.

If you wanted to put in a doc entry, you could just start a new section that could someday contain all of the options that just need one-line explanations like this one. Or you can skip it.

lua/neo-tree/sources/common/hijack_cursor.lua
ghostbuster91: Nice, this simplifies everything :)

lua/neo-tree/sources/common/hijack_cursor.lua
ghostbuster91: done

lua/neo-tree/sources/common/hijack_cursor.lua
ghostbuster91: Although, this doesn't seem to work for some reason. Any idea what could that be?

ghostbuster91: It seems that `all_states` is not populated in the `manager` at the time I initialize hijack.

cseickel: I'm sorry, I kind of mislead you here. This doesn't really work because states are created and destroyed as the user opens new trees in new tabs/windows and destroys those tabs/windows.

The event should just be subscribed once, and then every time it fires you should figure out which neo-tree window you are in (if any), then retrieve the state for that window or tab and move the cursor.

You can tell if a neo-tree window is focused by checking the buffer local variables. This code is an example of getting state for the current window:
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8c89efb93b8383666a6f989762c2f610f5f2c930/lua/neo-tree/setup/init.lua#L282-L287

If the position IS current, the get_state call needs to include the window id.

ghostbuster91:
Hi @cseickel, it seems that I can now get back to this one, though I still don't know where to subscribe for the events. Could you drop some hint? :pray:

lua/neo-tree/sources/common/hijack_cursor.lua
cseickel: Here's one example of subscribing to an event, there are actually a lot of subscriptions in the setup/init methods:

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/7f6fa04dbd8e8c79d1af33bc90e856b65d8641da/lua/neo-tree/setup/init.lua#L704-L708

You can also view the help file for more details about the events and arguments:
```
:h neo-tree-events
```

ghostbuster91: all clear now, thanks

lua/neo-tree/sources/common/hijack_cursor.lua
```diff
@@ -0,0 +1,38 @@
...
+    if state == nil then
+      return
+    end
+    local winid = state.winid
+    if vim.api.nvim_get_current_win() == winid then
+      local node = state.tree:get_node()
+      log.debug("Cursor moved in tree window, hijacking cursor position")
+      local cursor = vim.api.nvim_win_get_cursor(0)
+      local row = cursor[1]
+      local current_line = vim.api.nvim_get_current_line()
```
cseickel, line 21: I think this could be:
```suggestion
      local current_line = cursor[0]
```

cseickel (commented):
I'm not sure why, but it does not seem to work with `position = "current"`.

lua/neo-tree/sources/common/hijack_cursor.lua
ghostbuster91: I think that you meant `cursor[2]` but that returns only the column position. Here I need to fetch the text for the current line.

lua/neo-tree/sources/common/hijack_cursor.lua
```diff
@@ -0,0 +1,38 @@
+local events = require("neo-tree.events")
+local log = require("neo-tree.log")
+
+local M = {}
+
+local hijack_cursor_handler = function()
+    if vim.o.filetype ~= "neo-tree" then
+        return
+    end
+    local source = vim.api.nvim_buf_get_var(0, "neo_tree_source")
```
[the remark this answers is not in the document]
ghostbuster91: Good idea, done.

ghostbuster91:
> I'm not sure why, but it does not seem to work with `position = "current"`.

I will look into that.

lua/neo-tree/sources/common/hijack_cursor.lua
cseickel: You're right, I was very confused. I was thinking that you were looking for the current line number for some reason, and also that tables were zero indexed.

ghostbuster91:
@cseickel I looked into that. It doesn't work because `state.winid` is `nil`. I thought that maybe this is some special case that needs to be handled. 
I found this https://github.com/ghostbuster91/neo-tree.nvim/blob/11f6042509d8e4dc3ffbddf0bf1d8b7d01ac3d94/lua/neo-tree/ui/selector.lua#L254 and made a similar change. However after that change neotree failed because `state.tree` is also `nil` :eyes: (in this line https://github.com/ghostbuster91/neo-tree.nvim/blob/11f6042509d8e4dc3ffbddf0bf1d8b7d01ac3d94/lua/neo-tree/sources/common/hijack_cursor.lua#L21) 
Any idea what is happening here?

cseickel:
> Any idea what is happening here?

Yes. The problem is that if you call `manager.get_state()` without specifying a window id, you will get back the sidebar state for the tab. What you need to do is first check if the window was open in `position = "current"` or not, and if the position is `"current"` then you need to pass the winid as the third parameter to `get_state(source, tabid, winid)` to get the state for that window and not the tab.

You'd need something like this:

```lua
local winid = nil
local _, position = pcall(vim.api.nvim_buf_get_var, 0, "neo_tree_position")
if position == "current" then
  winid = vim.api.get_current_win()
end
local state = manager.get_state(source, nil, winid)
```

That block of logic should probably be a utility function, or mayabe the `get_state` function itself should be checking if the currently focused window is a neo-tree window and doing that work for you.

ghostbuster91:
I see, thanks for the explanation. For now I will just add this check here but I think that it would make sense to move it eventually to `get_state`. I can do that as a follow-up.

cseickel:
Hey @ghostbuster91, sorry I think I dropped the ball on this. Is this good to go?

ghostbuster91:
@cseickel I fixed the issue with `position=current`. Could you give it a try?

cseickel (approved):
Looks good and it works for me in all situations. Thanks!

Issue #701, opened 2023-01-16
`expand_all_nodes` does not work


cseickel:
It only expands a few levels, but it used to work to any arbitrary depth.

swarn:
I'll repeat this here, because I think it's relevant, and I put too much noise in #445.

I have my own `recursive_open` function that uses `fs.toggle_directory`. I find that it works properly if I add a delay after `fs.toggle_directory`, like so:

``` lua
-- Expand a node and load filesystem info if needed.
local function open_dir(state, dir_node)
  local fs = require "neo-tree.sources.filesystem"
  fs.toggle_directory(state, dir_node, nil, true, false)
  vim.wait(50)  -- required for correct operation (?)
end

-- Expand a node and all its children, optionally stopping at max_depth.
local function recursive_open(state, node, max_depth)
  local max_depth_reached = 1
  local stack = { node }
  while next(stack) ~= nil do
    node = table.remove(stack)
    if node.type == "directory" and not node:is_expanded() then
      open_dir(state, node)
    end

    local depth = node:get_depth()
    max_depth_reached = math.max(depth, max_depth_reached)

    if not max_depth or depth < max_depth - 1 then
      local children = state.tree:get_nodes(node:get_id())
      for _, v in ipairs(children) do
        table.insert(stack, v)
      end
    end
  end

  return max_depth_reached
end
```

miversen33:
So doing a bit of quick digging. I added some logging to the expand nodes function and it appears that there are no children when you go to further expand them down. Peeking at [lines 105 - 134 of commands.lua](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8af176ed7a8ce7b042ab95e11af81ee1a70244ef/lua/neo-tree/sources/common/commands.lua#L105-L134), I added some logging into the recursive loop that gets children and found that after the loop processes the currently visible set of nodes, it gets no children for those nodes. I am _guessing_ that @swarn 's adding of the sleep function works because its giving the expand function enough time to populate the children for the expanded node. 

So I wonder if the issue is actually related to how children are being populated for a node?

miversen33:
![Before Expand](https://user-images.githubusercontent.com/2640668/215239005-eae0bc49-1464-45e6-a3a3-b84086ef9f3a.png)
![After Expand](https://user-images.githubusercontent.com/2640668/215239011-ca5cea97-dd99-4288-ace6-02eda3bdc611.png)
[...]

PR #957, opened 2023-05-28
refactor(filesystem): fix recursive expand of nodes

ghostbuster91:
This relates to #701 

This is not a full fix for the builtin `expand_all_nodes` method but rather a fix that enables fixing it. It also enables fixing function from wiki to recursively expand nodes under the cursor (more important in my opinion).

List of changes broken down by commit:
1. Add callback to `toggle_directory` - this is the most important. Luckily it is not a breaking change because callback was previously `nil` anyway. 
    Justification:
   `fs.toggle_directory` calls `fs_scan.get_items` that calls `sync_scan` that for the 
   `scan_mode=="shallow"` calls `sync_scan` and then `job_complete`. 

   Here is where the problem starts - `job_complete` checks if `git_status_async` is set and 
   if so it registers a callback in `git.mark_ignored` method. That callback is important 
   because it eventually calls `render_context` that updates the nodes. That callback will be 
   called at some point later in time, but the execution continues in the meantime. Because 
   of that it is not guaranteed that we will have updated results by the time we start 
   expanding nodes. 
3. Small refactoring - method extraction. Can be removed from the list of changes if desired.
4. Small fix for the recursive traversals when `deep` strategy was used. 

I did not fix the `expand_all_nodes` because I didn't know how can I obtain the root element of the whole filesystem tree. The root element is needed so that we can start expanding from it rather than expanding from its children like it is now:
```lua
  for _, node in ipairs(state.tree:get_nodes()) do
    expand_node(node)
  end
```
This is important so that we can have a single callback that we pass to `toggle_directory`. Otherwise, this method will spawn as many processes as there are nodes on the top level. Tbh I would remove this method. 

And here is a fixed method to expand recursively nodes under the cursor `recursive_open`:
```lua
    local function expand_all(state, root)
        local toggle_directory = function(_, node)
            node:expand()
        end

        local expand_node
        expand_node = function(node)
            local id = node:get_id()
            if node.type == "directory" and not node:is_expanded() then
                toggle_directory(state, node)
                node = state.tree:get_node(id)
            end
            local children = state.tree:get_nodes(id)
            if children then
                for _, child in ipairs(children) do
                    if child.type == "directory" then
                        expand_node(child)
                    end
                end
            end
        end

        expand_node(root)
        renderer.redraw(state)
    end

    local function open_dir(state, dir_node, callback)
        local fs = require "neo-tree.sources.filesystem"
        fs.toggle_directory(state, dir_node, nil, true, true, callback)
    end

    -- Expand a node and all its children
    -- does not handle max_depth
    local function recursive_open(state, node, max_depth)
        open_dir(state, node, function()
            expand_all(state, node)
        end)
    end
```

Tested both with `deep` and `shallow` strategies.

### DIFF
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -114,9 +114,6 @@ end
 ---@param state table The state of the source
 ---@param node table A node to expand
 M.expand_all_nodes = function(state, node, source_expander)
-  if node == nil then
-    node = state.tree:get_node(state.path)
-  end
   log.debug("Expanding all nodes under " .. node:get_id())
   if source_expander == nil then
     source_expander = node_expander.default_expander
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -115,6 +115,7 @@ end
 ---@param node table A node to expand
 M.expand_all_nodes = function(state, node)
   log.debug("Expanding all nodes under " .. node:get_id())
+  renderer.position.set(state, nil)
   local task = function ()
     fs.expand_directory(state, node)
   end
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -113,16 +113,17 @@ end
 ---Expand all nodes
 ---@param state table The state of the source
 ---@param node table A node to expand
-M.expand_all_nodes = function(state, node, source_expander)
+---@param prefetcher table an object with two methods `prefetch(state, node)` and `should_prefetch(node) => boolean`
+M.expand_all_nodes = function(state, node, prefetcher)
   log.debug("Expanding all nodes under " .. node:get_id())
-  if source_expander == nil then
-    source_expander = node_expander.default_expander
+  if prefetcher == nil then
+    prefetcher = node_expander.default_prefetcher
   end
 
   renderer.position.set(state, nil)
 
   local task = function ()
-    node_expander.expand_directory_recursively(state, node, source_expander)
+    node_expander.expand_directory_recursively(state, node, prefetcher)
   end
   async.run(
       task,
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -117,7 +117,7 @@ M.expand_all_nodes = function(state, node)
   log.debug("Expanding all nodes under " .. node:get_id())
   renderer.position.set(state, nil)
   local task = function ()
-    fs.expand_directory(state, node)
+    fs.expand_directory_recursively(state, node)
   end
   async.run(
       task,
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -2,6 +2,7 @@
 
 local vim = vim
 local fs_actions = require("neo-tree.sources.filesystem.lib.fs_actions")
+local fs = require("neo-tree.sources.filesystem")
 local utils = require("neo-tree.utils")
 local renderer = require("neo-tree.ui.renderer")
 local events = require("neo-tree.events")
@@ -10,6 +11,7 @@ local popups = require("neo-tree.ui.popups")
 local log = require("neo-tree.log")
 local help = require("neo-tree.sources.common.help")
 local Preview = require("neo-tree.sources.common.preview")
+local async = require("plenary.async")
 
 ---Gets the node parent folder
 ---@param state table to look for nodes
@@ -108,35 +110,23 @@ M.add_directory = function(state, callback)
   fs_actions.create_directory(in_directory, callback, using_root_directory)
 end
 
-M.expand_all_nodes = function(state, toggle_directory)
-  if toggle_directory == nil then
-    toggle_directory = function(_, node)
-      node:expand()
-    end
+---Expand all nodes
+---@param state table The state of the source
+---@param node table A node to expand
+M.expand_all_nodes = function(state, node)
+  log.debug("Expanding all nodes under " .. node:get_id())
+  local task = function ()
+    fs.expand_directory(state, node)
   end
-  --state.explicitly_opened_directories = state.explicitly_opened_directories or {}
-
-  local expand_node
-  expand_node = function(node)
-    local id = node:get_id()
-    if node.type == "directory" and not node:is_expanded() then
-      toggle_directory(state, node)
-      node = state.tree:get_node(id)
-    end
-    local children = state.tree:get_nodes(id)
-    if children then
-      for _, child in ipairs(children) do
-        if child.type == "directory" then
-          expand_node(child)
-        end
+  async.run(
+      task,
+      function ()
+        log.debug("All nodes expanded - redrawing")
+        vim.schedule_wrap(function()
+          renderer.redraw(state)
+        end)
       end
-    end
-  end
-
-  for _, node in ipairs(state.tree:get_nodes()) do
-    expand_node(node)
-  end
-  renderer.redraw(state)
+    )
 end
 
 M.close_node = function(state, callback)
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -122,9 +122,7 @@ M.expand_all_nodes = function(state, node)
       task,
       function ()
         log.debug("All nodes expanded - redrawing")
-        vim.schedule_wrap(function()
-          renderer.redraw(state)
-        end)
+        renderer.redraw(state)
       end
     )
 end
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -2,7 +2,6 @@
 
 local vim = vim
 local fs_actions = require("neo-tree.sources.filesystem.lib.fs_actions")
-local fs = require("neo-tree.sources.filesystem")
 local utils = require("neo-tree.utils")
 local renderer = require("neo-tree.ui.renderer")
 local events = require("neo-tree.events")
@@ -12,6 +11,7 @@ local log = require("neo-tree.log")
 local help = require("neo-tree.sources.common.help")
 local Preview = require("neo-tree.sources.common.preview")
 local async = require("plenary.async")
+local node_expander = require("neo-tree.sources.common.node_expander")
 
 ---Gets the node parent folder
 ---@param state table to look for nodes
@@ -113,11 +113,19 @@ end
 ---Expand all nodes
 ---@param state table The state of the source
 ---@param node table A node to expand
-M.expand_all_nodes = function(state, node)
+M.expand_all_nodes = function(state, node, source_expander)
+  if node == nil then
+    node = state.tree:get_node(state.path)
+  end
   log.debug("Expanding all nodes under " .. node:get_id())
+  if source_expander == nil then
+    source_expander = node_expander.default_expander
+  end
+
   renderer.position.set(state, nil)
+
   local task = function ()
-    fs.expand_directory_recursively(state, node)
+    node_expander.expand_directory_recursively(state, node, source_expander)
   end
   async.run(
       task,
--- a/lua/neo-tree/sources/common/node_expander.lua
+++ b/lua/neo-tree/sources/common/node_expander.lua
@@ -7,9 +7,9 @@ local M = {}
 ---@param node table a node to expand
 ---@param state table current state of the source
 ---@return table discovered nodes that need to be loaded
-local function expand_loaded(node, state, node_expander)
+local function expand_loaded(node, state, prefetcher)
     local function rec(current_node, to_load)
-      if node_expander.should_prefetch(current_node) then
+      if prefetcher.should_prefetch(current_node) then
         log.trace("Node " .. current_node:get_id() .. "not loaded, saving for later")
         table.insert(to_load, current_node)
       else
@@ -39,14 +39,14 @@ end
 --- async method
 ---@param node table a node to expand
 ---@param state table current state of the source
-local function expand_and_load(node, state, node_expander)
+local function expand_and_load(node, state, prefetcher)
     local function rec(to_load, progress)
-      local to_load_current = expand_loaded(node, state, node_expander)
+      local to_load_current = expand_loaded(node, state, prefetcher)
       for _,v in ipairs(to_load_current) do
         table.insert(to_load, v)
       end
       if progress <= #to_load then
-        M.expand_directory_recursively(state, to_load[progress], node_expander)
+        M.expand_directory_recursively(state, to_load[progress], prefetcher)
         rec(to_load, progress + 1)
       end
     end
@@ -54,26 +54,28 @@ local function expand_and_load(node, state, node_expander)
 end
 
 --- Expands given node recursively loading all descendant nodes if needed
+--- Nodes will be loaded using given prefetcher
 --- async method
 ---@param state table current state of the source
 ---@param node table a node to expand
-M.expand_directory_recursively = function(state, node, node_expander)
+---@param prefetcher table an object with two methods `prefetch(state, node)` and `should_prefetch(node) => boolean`
+M.expand_directory_recursively = function(state, node, prefetcher)
   log.debug("Expanding directory " .. node:get_id())
   if node.type ~= "directory" then
     return
   end
   state.explicitly_opened_directories = state.explicitly_opened_directories or {}
-  if node_expander.should_prefetch(node) then
+  if prefetcher.should_prefetch(node) then
     local id = node:get_id()
     state.explicitly_opened_directories[id] = true
-    node_expander.prefetch(state, node)
-    expand_loaded(node, state, node_expander)
+    prefetcher.prefetch(state, node)
+    expand_loaded(node, state, prefetcher)
   else
-    expand_and_load(node, state, node_expander)
+    expand_and_load(node, state, prefetcher)
   end
 end
 
-M.default_expander = {
+M.default_prefetcher = {
   prefetch = function (state, node)
     log.debug("Default expander prefetch does nothing")
   end,
--- a/lua/neo-tree/sources/common/node_expander.lua
+++ b/lua/neo-tree/sources/common/node_expander.lua
@@ -1,0 +1,85 @@
+local log = require("neo-tree.log")
+
+local M = {}
+
+--- Recursively expand all loaded nodes under the given node
+--- returns table with all discovered nodes that need to be loaded 
+---@param node table a node to expand
+---@param state table current state of the source
+---@return table discovered nodes that need to be loaded
+local function expand_loaded(node, state, node_expander)
+    local function rec(current_node, to_load)
+      if node_expander.should_prefetch(current_node) then
+        log.trace("Node " .. current_node:get_id() .. "not loaded, saving for later")
+        table.insert(to_load, current_node)
+      else
+        if not current_node:is_expanded() then
+          current_node:expand()
+          state.explicitly_opened_directories[current_node:get_id()] = true
+        end
+        local children = state.tree:get_nodes(current_node:get_id())
+        log.debug("Expanding childrens of " .. current_node:get_id())
+        for _, child in ipairs(children) do
+          if child.type == "directory" then
+             rec(child, to_load)
+          else
+            log.trace("Child: " .. child.name .. " is not a directory, skipping")
+          end
+        end
+      end
+    end
+
+    local to_load = {}
+    rec(node, to_load)
+    return to_load
+end
+
+--- Recursively expands all nodes under the given node
+--- loading nodes if necessary.
+--- async method
+---@param node table a node to expand
+---@param state table current state of the source
+local function expand_and_load(node, state, node_expander)
+    local function rec(to_load, progress)
+      local to_load_current = expand_loaded(node, state, node_expander)
+      for _,v in ipairs(to_load_current) do
+        table.insert(to_load, v)
+      end
+      if progress <= #to_load then
+        M.expand_directory_recursively(state, to_load[progress], node_expander)
+        rec(to_load, progress + 1)
+      end
+    end
+    rec({}, 1)
+end
+
+--- Expands given node recursively loading all descendant nodes if needed
+--- async method
+---@param state table current state of the source
+---@param node table a node to expand
+M.expand_directory_recursively = function(state, node, node_expander)
+  log.debug("Expanding directory " .. node:get_id())
+  if node.type ~= "directory" then
+    return
+  end
+  state.explicitly_opened_directories = state.explicitly_opened_directories or {}
+  if node_expander.should_prefetch(node) then
+    local id = node:get_id()
+    state.explicitly_opened_directories[id] = true
+    node_expander.prefetch(state, node)
+    expand_loaded(node, state, node_expander)
+  else
+    expand_and_load(node, state, node_expander)
+  end
+end
+
+M.default_expander = {
+  prefetch = function (state, node)
+    log.debug("Default expander prefetch does nothing")
+  end,
+  should_prefetch = function (node)
+    return false
+  end
+}
+
+return M
--- a/lua/neo-tree/sources/common/node_expander.lua
+++ b/lua/neo-tree/sources/common/node_expander.lua
@@ -34,23 +34,18 @@ local function expand_loaded(node, state, prefetcher)
     return to_load
 end
 
---- Recursively expands all nodes under the given node
---- loading nodes if necessary.
+--- Recursively expands all nodes under the given node collecting all unloaded nodes
+--- Then run prefetcher on all unloaded nodes. Finally, expand loded nodes. 
 --- async method
 ---@param node table a node to expand
 ---@param state table current state of the source
 local function expand_and_load(node, state, prefetcher)
-    local function rec(to_load, progress)
-      local to_load_current = expand_loaded(node, state, prefetcher)
-      for _,v in ipairs(to_load_current) do
-        table.insert(to_load, v)
+      local to_load = expand_loaded(node, state, prefetcher)
+      for _, _node in ipairs(to_load) do
+        prefetcher.prefetch(state, _node)
+        -- no need to handle results as prefetch is recursive
+        expand_loaded(_node, state, prefetcher)
       end
-      if progress <= #to_load then
-        M.expand_directory_recursively(state, to_load[progress], prefetcher)
-        rec(to_load, progress + 1)
-      end
-    end
-    rec({}, 1)
 end
 
 --- Expands given node recursively loading all descendant nodes if needed
--- a/lua/neo-tree/sources/filesystem/commands.lua
+++ b/lua/neo-tree/sources/filesystem/commands.lua
@@ -68,6 +68,9 @@ M.delete_visual = function(state, selected_nodes)
 end
 
 M.expand_all_nodes = function(state, node)
+  if node == nil then
+    node = state.tree:get_node(state.path)
+  end
   cc.expand_all_nodes(state, node, fs.expander)
 end
 
--- a/lua/neo-tree/sources/filesystem/commands.lua
+++ b/lua/neo-tree/sources/filesystem/commands.lua
@@ -68,10 +68,7 @@ M.delete_visual = function(state, selected_nodes)
 end
 
 M.expand_all_nodes = function(state)
-  local toggle_dir_no_redraw = function(_state, node)
-    fs.toggle_directory(_state, node, nil, true, true)
-  end
-  cc.expand_all_nodes(state, toggle_dir_no_redraw)
+  cc.expand_all_nodes(state, state.tree:get_node(state.path))
 end
 
 ---Shows the filter input, which will filter the tree.
--- a/lua/neo-tree/sources/filesystem/commands.lua
+++ b/lua/neo-tree/sources/filesystem/commands.lua
@@ -71,7 +71,7 @@ M.expand_all_nodes = function(state, node)
   if node == nil then
     node = state.tree:get_node(state.path)
   end
-  cc.expand_all_nodes(state, node, fs.expander)
+  cc.expand_all_nodes(state, node, fs.prefetcher)
 end
 
 ---Shows the filter input, which will filter the tree.
--- a/lua/neo-tree/sources/filesystem/commands.lua
+++ b/lua/neo-tree/sources/filesystem/commands.lua
@@ -67,8 +67,8 @@ M.delete_visual = function(state, selected_nodes)
   cc.delete_visual(state, selected_nodes, utils.wrap(refresh, state))
 end
 
-M.expand_all_nodes = function(state)
-  cc.expand_all_nodes(state, state.tree:get_node(state.path))
+M.expand_all_nodes = function(state, node)
+  cc.expand_all_nodes(state, node, fs.expander)
 end
 
 ---Shows the filter input, which will filter the tree.
--- a/lua/neo-tree/sources/filesystem/init.lua
+++ b/lua/neo-tree/sources/filesystem/init.lua
@@ -471,7 +471,7 @@ local function expand_and_load(node, state)
         table.insert(to_load, v)
       end
       if progress <= #to_load then
-        M.expand_directory(state, to_load[progress])
+        M.expand_directory_recursively(state, to_load[progress])
         rec(to_load, progress + 1)
       end
     end
--- a/lua/neo-tree/sources/filesystem/init.lua
+++ b/lua/neo-tree/sources/filesystem/init.lua
@@ -428,74 +428,14 @@ M.toggle_directory = function(state, node, path_to_reveal, skip_redraw, recursiv
   end
 end
 
---- Recursively expand all loaded nodes under the given node
---- returns table with all discovered nodes that need to be loaded 
----@param node table a node to expand
----@param state table current state of the source
----@return table discovered nodes that need to be loaded
-local function expand_loaded(node, state)
-    local function rec(_node, to_load)
-      if _node.loaded == false then
-        table.insert(to_load, _node)
-      else
-        if not _node:is_expanded() then
-          _node:expand()
-          state.explicitly_opened_directories[_node:get_id()] = true
-        end
-        local children = state.tree:get_nodes(_node:get_id())
-        log.debug("Expanding childrens of " .. _node:get_id())
-        for _, child in ipairs(children) do
-          if child.type == "directory" then
-             rec(child, to_load)
-          else
-            log.trace("Child: " .. child.name .. " is not a directory, skipping")
-          end
-        end
-      end
-    end
-
-    local to_load = {}
-    rec(node, to_load)
-    return to_load
-end
-
---- Recursively expands all nodes under the given node
---- loading nodes if necessary.
---- async method
----@param node table a node to expand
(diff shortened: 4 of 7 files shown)


cseickel:
> I did not fix the expand_all_nodes because I didn't know how can I obtain the root element of the whole filesystem tree.

I think you can get the root with `state.tree:get_node(state.path)`, because the ID's of teh nodes is just the absolute path.

Would you like to continue and fix that expand all nodes built-in as well or just merge waht you have now?

BTW - I do agree with all of your function extractions. Some of the functions have grown way too big and breaking them up into smaller functions with descriptive names is always good.

ghostbuster91:
> I think you can get the root with state.tree:get_node(state.path), because the ID's of the nodes is just the absolute path.

:+1: Thanks

> Would you like to continue and fix that expand all nodes built-in as well or just merge what you have now?

I am happy to work on it further. In fact I realized that there might be some cases where the expanding might not work correctly.

ghostbuster91:
Ok so here is what I discover/realized: 
Even when expanding nodes under cursor, some child nodes might already be loaded/expanded. 
Consider following situation:

```
dir1 <- cursor here
\- child1 - loaded
\- child2 - not loaded
\- child3 - loaded and expanded
  \- file.lua
  \- grandchild1 - not loaded
```
If dir1 is unloaded then the situation is simple, we enter this path:
```lua
  if node.loaded == false then
    local id = node:get_id()
    state.explicitly_opened_directories[id] = true
    renderer.position.set(state, nil)
    fs_scan.get_items(state, id, path_to_reveal, callback, true, recursive)
```

If `dir1` is loaded then we need to basically call that method for each of its unloaded children. The problem is that this method is async and it has to be called recursively. 

I am still quite new to lua and its concurrency features/gotchas but from what I saw, this seems like a job for `plenary.async`.

I tried rewriting `expand_all_nodes` to use it, but for some reason the execution of coroutines seems to be stopped before all of them have a chance to finish. 

```lua
M.expand_all_nodes = function(state, toggle_directory)
  local async = require("plenary.async")

  -- toggle_directory = async.wrap(toggle_directory, 3)
  local expand_node

  local expand_async = function (node)
    local id = node:get_id()
    local toggle_async = async.wrap(function (callbackF)
      if node.type == "directory" and not node:is_expanded() then
          log.debug("before toggle " .. node.name)
          toggle_directory(state, node, function ()
              log.debug("after toggle " .. node.name)
              node = state.tree:get_node(id)
              callbackF(node)
          end)
      else
        callbackF(node)
      end
    end,1)
    toggle_async()
    log.debug("after toggle_async " .. node.name)
    local children = state.tree:get_nodes(id)
    local tasks = {}
    if children then
      log.debug("children of " .. node.name .. " - "  .. #children)
      for _, child in ipairs(children) do
        if child.type == "directory" then
          local task = async.wrap(function (callback2)
            expand_node(child, callback2)
          end, 1)
          table.insert(tasks, task)
        else
          log.debug("skipping child " .. vim.inspect(child.name))
        end
      end
    else
      log.debug("no kids")
    end
    if #tasks > 0 then
		async.util.join(tasks)
	end
  end



  expand_node = function(node, callback)
    async.run(function ()
        log.debug("async run in expand_node " .. node.name )
        return expand_async(node)
    end, callback)
  end

  local tasks = {}
  for _, node in ipairs(state.tree:get_nodes()) do
    local task = async.wrap(function (callback)
        log.debug("async expand node" .. node.name)
        expand_node(node, callback)
    end, 1)
    table.insert(tasks, task)
  end
    log.debug("run all on top")
    async.util.run_all(
      tasks,
      function ()
        log.debug("run all on top - finish")
        vim.schedule_wrap(function()
          renderer.redraw(state)
        end)
      end
    )
end
```

here is how logs ends
```
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:123: after toggle scripts
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:132: after toggle_async scripts
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:136: children of scripts - 1
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:144: skipping child "test.sh"
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:123: after toggle plugin
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:132: after toggle_async plugin
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:136: children of plugin - 1
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:144: skipping child "neo-tree.vim"
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:123: after toggle lua
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:132: after toggle_async lua
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:136: children of lua - 2
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:144: skipping child "neo-tree.lua"
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:159: async run in expand_node neo-tree
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:121: before toggle neo-tree
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:123: after toggle doc
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:132: after toggle_async doc
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:136: children of doc - 1
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:144: skipping child "neo-tree.txt"
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:123: after toggle tests
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:132: after toggle_async tests
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:136: children of tests - 3
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:144: skipping child "mininit.lua"
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:159: async run in expand_node neo-tree
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:121: before toggle neo-tree
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:159: async run in expand_node utils
[DEBUG Tue 30 May 2023 10:43:39 AM CEST] .//lua/neo-tree/sources/common/commands.lua:121: before toggle utils
```

Any idea what could be the reason or how to do it without `plenary.async`?

cseickel:
TBH I don't have enough time left today to load recursive async code into my working memory and reason about where the problem is. My gut is telling me it would be better to add directories to be scanned to a collection as they are discovered and keep looping until that collection is exhausted.

The async part probably breaks that technique... Maybe every time an async methods adds a new directory to that collection, the loop can be restarted if it had previously been exhausted and completed. This may be a situation where OOP is better than functional. Neo-tree has more functional style code but we do use OOP at times where it makes sense.

Anyway, I would be thrilled if you rewrote it because I know you write good code. This section of the code has gotten messy and confusing.

ghostbuster91:
> Anyway, I would be thrilled if you rewrote it because I know you write good code. This section of the code has gotten messy and confusing.

Crap, now I will have to deliver on these expectations xD

> The async part probably breaks that technique...

Yeah, that wouldn't work. I need to dive more into plenary async and threading in lua. 

> This may be a situation where OOP is better than functional. Neo-tree has more functional style code but we do use OOP at times where it makes sense.

Idk. I have been writing functional code for the last couple of years and this seems fine like every other place. I am glad that you took that approach. It is much easier to reason about that codebase than the nvim-tree one (at least for me). Sure there are some messy parts but nothing that cannot be cleaned up. 

In general I am happy to continue working on it but it will take more time than I initially anticipated.

ghostbuster91:
just a small update from my side:

[Kooha-2023-06-02-10-46-16.webm](https://github.com/nvim-neo-tree/neo-tree.nvim/assets/5662622/f0f336ab-521d-48ba-9852-459111319d38)

The code seems to work now. Though I have made some significant changes. I am not pushing it yet as I need to clean it. 

After that I will publish it and describe my changes with recommendations of moving forward. In general I would avoid using callbacks as they are hard to follow and error-prone (e.g. calling it twice or forgot to call it at all).

ghostbuster91:
@cseickel I have it almost ready. Just one small bug left. Could you tell me how `node.is_loaded` gets changed?

cseickel:
> @cseickel I have it almost ready. Just one small bug left. Could you tell me how `node.is_loaded` gets changed?

@ghostbuster91 that `is_loaded` property is junk that got slipped in somehow. It could be that it was just a typo. You are probably looking for `node.loaded` which is meaningful and used in many places. That is set here:

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8c89efb93b8383666a6f989762c2f610f5f2c930/lua/neo-tree/ui/renderer.lua#L1091

lua/neo-tree/sources/filesystem/init.lua
```diff
@@ -428,4 +429,76 @@ M.toggle_directory = function(state, node, path_to_reveal, skip_redraw, recursiv
...
+    return
+  end
+  state.explicitly_opened_directories = state.explicitly_opened_directories or {}
+  if node.loaded == false then
+    local id = node:get_id()
+    state.explicitly_opened_directories[id] = true
+    renderer.position.set(state, nil)
+    fs_scan.get_dir_items_async(state, id, true)
+    -- ignore results as we know here that all descendant nodes have been already loaded
+    expand_loaded(node ,state)
```
ghostbuster91, line 498: Here is the problem - for some reason `fs_scan.get_dir_items_async` does set `loaded=true` for all the nodes in the tree despite recursively scanning whole directory. 

`fs_scan.get_dir_items_async` is a simplification of `fs_scan.get_items`

If you could point me to a place that sets `loaded` I will be able to debug it further. 

Right now, expanding `all_nodes` works but is inefficient due to redundant directory scans.

lua/neo-tree/sources/filesystem/commands.lua
```diff
@@ -68,10 +68,7 @@ M.delete_visual = function(state, selected_nodes)
 end
 
 M.expand_all_nodes = function(state)
-  local toggle_dir_no_redraw = function(_state, node)
-    fs.toggle_directory(_state, node, nil, true, true)
-  end
-  cc.expand_all_nodes(state, toggle_dir_no_redraw)
+  cc.expand_all_nodes(state, state.tree:get_node(state.path))
```
ghostbuster91, line 71: Thanks to such interface expanding all nodes recursively under the cursor will be as simple as `expand_all_nodes(state, state.tree:get_node())`.

lua/neo-tree/sources/common/commands.lua
```diff
@@ -108,35 +110,23 @@ M.add_directory = function(state, callback)
...
-        if child.type == "directory" then
-          expand_node(child)
-        end
+---Expand all nodes
+---@param state table The state of the source
+---@param node table A node to expand
+M.expand_all_nodes = function(state, node)
+  log.debug("Expanding all nodes under " .. node:get_id())
+  local task = function ()
+    fs.expand_directory(state, node)
```
ghostbuster91, line 119: I switched semantics from toggle to purely expand because if there is a state like:
```
root
- dir1 (collapsed, loaded)
- - child1 (expanded, loaded)
```

Then a recursive toggle would expand dir1 but collapse child1 which is not what we want here.

ghostbuster91:
A general comment to my change:

I think that we should migrate the codebase to plenary.async as relying on callbacks is too error prone and not easy to follow. 

Thanks to that we should be able to get rid of `context` that is just a bunch of variables that are set in some contexts and in others not. 

This will makes the functions signatures clearer and allow to better reason about the codebase. 

Now, I don't expect anyone to do that work. However, I like neo-tree idea to that degree that I am willing to do the most of heavy lifting provided that you are interested in such changes.

lua/neo-tree/sources/filesystem/init.lua
ghostbuster91: Ok, I misunderstood how the `scan_dir_async` and `scan_dir_sync` behave. I thought that since we can pass there `recursive` flag they should be able to load all nodes recursively from the file system, and then all the nodes can be processed (like e.g. by expanding them). 

However, these functions load only one level below and changing them is not so easy as `renderContext` that I wanted to call for each scanned directory to create corresponding nodes in the tree clears all the variables in the context (there might be other obstacles to setting that). 

So, right now it works both for `expand_all` and `expand_under_cursor`. I am just not sure if it was supposed to work like this.

cseickel:
One general note about plenary: it's a great library and I appreciate the work that has been put into it, but be careful that you don't think of it as a "standard library" like what you would get from other languages. It's not the .Net Standard Library or Lodash, it's a young project maintained by a handful of very busy open source contributors.

Plenary async methods seem to be pretty good and wouldn't mind a refactor that leaned on those, but be careful and make sure that performance does not suffer. I usually use the Linux project to test out changes to the filesystem scanning code when I am worried about performance. Also, be sure that you consider that there may be some shared code with other sources that needs to be maintained.

-----

On the subject of the Plenary dir scanning functions, they are banned from this project. The first version used them of course, but over time I realized that it is slow and just too simple to work for what we need. Also, the gitignored filter is completely wrong. Gitignore is a very complicated thing and I stand by the decision to just relay that logic to the git cmd and not try to re-implement it ourselves.

I think that for a plugin that revolves around scanning the filesystem, it makes sense that we would roll our own functions for that one thing. We just need that level of control and customization.

lua/neo-tree/sources/filesystem/init.lua
```diff
@@ -428,4 +428,73 @@ M.toggle_directory = function(state, node, path_to_reveal, skip_redraw, recursiv
...
+      end
+    end
+    rec({}, 1)
+end
+
+---Expands given node recursively loading all descendant nodes if needed
+---async method
+---@param state table current state of the source
+---@param node table a node to expand
+M.expand_directory = function(state, node)
```
cseickel, line 485: I think the name of this function should indicate that it is a recursive load. If I saw a call to `expand_directory()` in the code somewhere, I would not expect it to be a recursive load.

lua/neo-tree/sources/filesystem/lib/fs_scan.lua
```diff
@@ -228,7 +257,9 @@ local function async_scan(context, path)
     local scan_tasks = {}
     for _, p in ipairs(context.paths_to_load) do
       local scan_task = async.wrap(function(callback)
-        scan_dir_async(context, p, callback)
+        async.run(function ()
```
cseickel, line 260: I haven't seen this before. Is this like a `pipe()` function from rxjs?

lua/neo-tree/sources/common/commands.lua
cseickel, line 119: The problem with this is that it was a generic function that worked across sources, which is why it was in the `common` module, but now it is specifically a `filesystem` only command.

I think the right thing to do is to restore this function to what it was, which would work for any source that does not lazy load, and move this update code into the `filesystem.commands.expand_all_nodes` function.

cseickel (commented):
Looks pretty good, just a few comments. The important thing is to return the `common.commnds.expand_all_nodes` function to being a generic implementation.

lua/neo-tree/sources/filesystem/lib/fs_scan.lua
ghostbuster91: Not really, it is rather like `subscribe`. It allows you to execute an "async" function from the regular(non-async) context.

lua/neo-tree/sources/common/commands.lua
ghostbuster91: Sure, I will do that :+1:

ghostbuster91:
> One general note about plenary: it's a great library and I appreciate the work that has been put into it, but be careful that you don't think of it as a "standard library" like what you would get from other languages. It's not the .Net Standard Library or Lodash, it's a young project maintained by a handful of very busy open source contributors.

I understand that, but there seem to be no other alternative, and my mental capacity is not big enough to load all the callbacks at once to reason about the code. 

> Plenary async methods seem to be pretty good and wouldn't mind a refactor that leaned on those, but be careful and make sure that performance does not suffer. I usually use the Linux project to test out changes to the filesystem scanning code when I am worried about performance. 

That is a great idea!


> On the subject of the Plenary dir scanning functions, they are banned from this project. The first version used them of course, but over time I realized that it is slow and just too simple to work for what we need. Also, the gitignored filter is completely wrong. Gitignore is a very complicated thing and I stand by the decision to just relay that logic to the git cmd and not try to re-implement it ourselves.

Understood, thanks for the explanation!


> I think that for a plugin that revolves around scanning the filesystem, it makes sense that we would roll our own functions for that one thing. We just need that level of control and customization.

Yep, that makes sense. At some point we might contribute it back to plenary if they will be generic enough.

lua/neo-tree/sources/filesystem/init.lua
ghostbuster91: @cseickel Could you also confirm that this is how it is supposed to work? 

I have a strong feeling that this is not exactly right, but maybe I still don't understand some things. I looked once again and it seems that this was the original intention as the `async_scan` with `scan_mode==shallow` scans directories recursively deep. So there is no reason to call it then multiple times for every new child.

ghostbuster91: nvm, I solved it, see this commit: https://github.com/nvim-neo-tree/neo-tree.nvim/pull/957/commits/b5a72242925026dba8de79b279ecf71e2c4fd434

cseickel:
> I understand that, but there seem to be no other alternative, and my mental capacity is not big enough to load all the callbacks at once to reason about the code.

I completely appreciate that. I've been there many times where there are sections of a project that are too complicated to fit in my working memory and I just have to refactor it. TBH - this code has been incrementally changed many times and by multiple people and is overdue for a refactor. The last time I had to change something it took me a while to get my bearings too.

lua/neo-tree/sources/filesystem/lib/fs_scan.lua
```diff
@@ -106,6 +106,26 @@ local render_context = function(context)
...
+local job_complete_async = function(context)
+  local state = context.state
+  local parent_id = context.parent_id
+  if #context.all_items == 0 then
+    log.info("No items, skipping git ignored/status lookups")
+  elseif state.filtered_items.hide_gitignored or state.enable_git_status then
+    local mark_ignored_async = async.wrap(function (_state, _all_items, _callback)
+      git.mark_ignored(_state, _all_items, _callback)
+    end, 3)
+    local all_items = mark_ignored_async(state, context.all_items)
```
ghostbuster91, line 118: **info** here I am always using async version of git_ignore, in contrary to `job_complete` that respects `require("neo-tree").config.git_status_async` and acts accordingly. 

I could merge these two together, but I am hesitant as the old one calls `render_context` which in the context of my function is unnecessary (at least in this place). 

Nevertheless, I still can add `config.git_status_asyn` to `job_complete_async` though I am not sure what are the benefits of that as everything is "async" in this callstack. Wdyt?

lua/neo-tree/sources/common/node_expander.lua
```diff
@@ -0,0 +1,85 @@
+local log = require("neo-tree.log")
```
ghostbuster91, line 1: I extracted this nodes expander as something that is filesystem agnostic and could be used to load also other lazy sources.

lua/neo-tree/sources/common/commands.lua
ghostbuster91: Done.

ghostbuster91:
So far all of my attempts of running it on a big repository like qmk or nixpkgs have failed due to not enough ram. I am quite surprised by that, so I will take a closer look at it.

lua/neo-tree/sources/common/node_expander.lua
cseickel: nice

lua/neo-tree/sources/filesystem/lib/fs_scan.lua
cseickel: I don't remember how everything works off the top of my head and I don't think I have enough time to dig in now, but I can say that if the filesystem is being loaded async it probably doesn't make sense to load git_status synchronously. I can see the other way around where the fs is scanned sync but git status is async.

cseickel:
> So far all of my attempts of running it on a big repository like qmk or nixpkgs have failed due to not enough ram. I am quite surprised by that, so I will take a closer look at it.

I've hit many difficult memory leaks in this project. I start with looking for closures on `state`, git status/ignored info, and diagnostics. 

There have also been memory leaks before in treesitter, neovim, and Nui. If this is redrawing after every recursive load, you could be hitting memory leaks external to your code.

ghostbuster91:
I found the memory issue, it was actually my fault as I was doing some redundant recursion. 

Now, I am facing another problem. My current solution works like that:
1. async scan recursively
2. mark-gitignored
3. expand all nodes
4. render

There is bit of a lag during the render, as there is a lot to render (I am testing it on a qmk repository that has 50k items in the `keyboard` directory). I guess it would fell more responsive if we rendered it more frequently. I think that this is how it was before, so maybe I will revert some of my changes. 

However, the bigger issue is with `mark-gitignored`. Neovim refuses to spawn that many jobs, and kills the neotree thread. Batching elements into smaller chunks before passing them to `mark_gitingored` solves this problem to some degree, as neovim is still unresponsive during that time. 

So, first I will try to redesign my approach a bit, and then I will look into `mark-gitignored` as I think that we could spawn less processes. 

The good thing is that I feel that I am making some progress, so hopefully this will be finished one day :)

lua/neo-tree/sources/filesystem/lib/fs_scan.lua
ghostbuster91: That sounds reasonable :+1:

ghostbuster91:
@pysan3 For now I am mostly focused only on recursive expansion of all nodes to the very bottom. For that purpose I created a separate chain of methods that performs that expansion in more efficient way. 

I guess once I finish it it will be possible to unify the way how we toggle a single directory with a recursive expansion. Though I would say this will be a different work stream. 

Ideally, we will have `expand_until(node => node.is_not_empty)` and others high level methods. But that is rather far away. 

Given the scope of your PR I think that you should not wait for me, since I am not sure when I will be able to finish it (I still want to dig into git_ignore stuff). And porting/adapting your changes will be rather easy.

ghostbuster91:
> I would be hesitant to render the incremental progress for each folder because the renders would probably take up a lot of resources themselves.

I agree.

> Maybe you can render once per level each time you reach a new max depth for the overall tree.

That could be an interesting feature, but let's save it for later. 

> The important thing is that it performs well in the normal operation where the next level may have a lot of items and with group_empty_dirs enabled.

:+1: Let's stick to that. 

@cseickel I came up with two ideas that could speed up recursive expand process that I would like to discuss:

## Scan parent instead of children

given following directory structure:
```
- dir1
\- dir2
\- dir3
\- dir4
```

It is not a surprise that it is faster to perform single `fs_scan` on `dir1` loading it and all each children recursively than to spawn 3 separate fs_scan coroutines to scan each of its children separately. This is more noticeable especially for directories with a lot of direct child directories. 

The current algorithm works like that:
If the dir1 is loaded then it will check each of its children. For any that is not loaded it wil start recursive fs_scan. If the dir1 is not loaded it will start recursive fs_scan only for it. 

Instead of that we could always perform recursive fs_scan on the parent. The obvious downside of that is that we'd perform that fs_scan every time we expand that node recursively. This could be mitigated by introduction of new flag on the node level - `fully_loaded`, which would denote if the node and all of each children has been loaded. Though, invalidation of such flag can be problematic so I am not sure if we would like to go that route. 

## Smarter git_ignore

The current way of checking git_ignored files works like that:
First we group items by directories, and then for each directory we call `git -C $directory --chekc-ignore $items` (which is a system process - heavy). We do this as the git repository doesn't have to be necessarily on the root level, but rather can be arbitrarily nested within the directory tree. There can be also multiple git repositories. 

However, once we detect that a given directory belongs to a git repository, then I think we could check all descendant directories against that one directory. That would greatly decrease the amount of jobs to spawn in certain situations (e.g. when the root directory is the git root). 

Now the question is how to detect that. I see two options - we could either save that information during the fs_scan as a node level flag - `inside_git_repo`, or we could spawn the jobs starting from the top level directories and only descend if they are not git repositories. 

## Summary

The second optimization is much more worth looking into as it can save a ton of resources. Having said that, I see them as additional improvements to the recursive expand feature and don't consider them critical. 

Apart from testing the changes with `group_empty_dir` improvements made by @pysan3, would you like me to do anything else before we merge it?

cseickel:
> ## Scan parent instead of children
> ...
> Instead of that we could always perform recursive fs_scan on the parent. The obvious downside of that is that we'd perform that fs_scan every time we expand that node recursively. This could be mitigated by introduction of new flag on the node level - `fully_loaded`, which would denote if the node and all of each children has been loaded. Though, invalidation of such flag can be problematic so I am not sure if we would like to go that route.

One thing you have to consider is that there is no way to know if files have changed since the last scan so every node has to be loaded from disk every time with no caching of any kind. The `loaded` flag is only for lazy loading.

I think if we wanted to cut down the number of workers, I would refactor to have a specific number of parallel workers that simply work out of a central queue of discovered directories and pop the next arbitrary directory to scan out of that queue whenever they finish, rather than creating exactly one per directory. I didn't personally add the one worker per directory strategy, but I think it was just a simple way of parallelizing the job.

> ## Smarter git_ignore
> The current way of checking git_ignored files works like that: First we group items by directories, and then for each directory we call `git -C $directory --chekc-ignore $items` (which is a system process - heavy). We do this as the git repository doesn't have to be necessarily on the root level, but rather can be arbitrarily nested within the directory tree. There can be also multiple git repositories.
> 
> However, once we detect that a given directory belongs to a git repository, then I think we could check all descendant directories against that one directory. That would greatly decrease the amount of jobs to spawn in certain situations (e.g. when the root directory is the git root).
> 
> Now the question is how to detect that. I see two options - we could either save that information during the fs_scan as a node level flag - `inside_git_repo`, or we could spawn the jobs starting from the top level directories and only descend if they are not git repositories.
> 

The git-ignore code was certainly not written with recursive expand in mind. That being said, this part is complicated and I would hate to complicate it further for a situation I see as an edge use case. It's OK if it is slow to recursively expand on a very large repo. I'd rather that than possibly add bugs for the average use case of s single level expand.

> ## Summary
> The second optimization is much more worth looking into as it can save a ton of resources. Having said that, I see them as additional improvements to the recursive expand feature and don't consider them critical.
> 
> Apart from testing the changes with `group_empty_dir` improvements made by @pysan3, would you like me to do anything else before we merge it?

I agree that you don't need to go further on optimization. My only concern is to ensure that if someone does recursively expand on a large repo, that neovim does not crash or completely freeze for a long period of time.

ghostbuster91:
Thanks for the detailed response, I appreciate that!

> One thing you have to consider is that there is no way to know if files have changed since the last scan so every node has to be loaded from disk every time with no caching of any kind. The loaded flag is only for lazy loading.

Don't we spawn a fs-watcher for every loaded directory that should inform us if anything changes? That way we could invalidate `fully_loaded` flag if changes were detected.

> The git-ignore code was certainly not written with recursive expand in mind. That being said, this part is complicated and I would hate to complicate it further for a situation I see as an edge use case. It's OK if it is slow to recursively expand on a very large repo. I'd rather that than possibly add bugs for the average use case of s single level expand.

That is a fair point. I might give it later a try to simplify it, because as you said it is complicated. 

> I agree that you don't need to go further on optimization. My only concern is to ensure that if someone does recursively expand on a large repo, that neovim does not crash or completely freeze for a long period of time.

Well, to be fair, that current behavior is exactly like that when you use `git_async=false` with recursive expand on big repository. So we are not introducing any regression rather fixing the case when `git_async=true`. 

I can later give it a try and make both of them more user-friendly but I would like not to growth the scope this PR that much.

cseickel:
> Don't we spawn a fs-watcher for every loaded directory that should inform us if anything changes? That way we could invalidate fully_loaded flag if changes were detected.

I suppose that is true when that option is enabled, and even if it's not, it's no change to current behavior since the nodes will be discarded on each explicit refresh.

------

Do you think this is ready for final review and testing?

ghostbuster91:
> Do you think this is ready for final review and testing?

I didn't have time yet to test it with `group_empty` and changes made by @pysan3. I will let you know once I finish it.

cseickel:
> > Do you think this is ready for final review and testing?
> 
> I didn't have time yet to test it with `group_empty` and changes made by @pysan3. I will let you know once I finish it.

Hey @ghostbuster91 I just want to check in on this. If you're out of time you don't need to worry about the fixes for `group_empty` as they can be handled separately. I originally called that out just to stop @pysan3 from doing too much work with a code base that was being refactored at the same time. We can easily just land this and then continue work on #908 separately.

pysan3:
> We can easily just land this and then continue work on https://github.com/nvim-neo-tree/neo-tree.nvim/pull/908 separately.

Indeed, sorry for the interruption @ghostbuster91.

I now think this PR is not that related to mine so please just ignore my comments on this thread lol. I am thinking of revising my PR after v3.0 is released so it'll probly be quite in the future.

ghostbuster91:
Yes, sorry guys, I got totally absorbed by other things. I will try to finish this by the end of the week. At this point I just wanted to test it with the current implementation of `group_empty`. Hopefully I won't find any issue and we will be able to close it :crossed_fingers:

ghostbuster91:
Just tested it with current `group_empty` - didn't spot any problems. I also pushed two small refactors.

ghostbuster91:
One more thing, when there is not enough file descriptors left in the system, plenary job will fail with following error:
```
Error executing luv callback:
.../myNeovimPackages/start/plenary.nvim/lua/plenary/job.lua:404: Failed to spawn process: {
  _additional_on_exit_callbacks = {},
  _shutdown_check = <userdata 1>,
  _stderr_results = {},
  _stdout_results = {},
  _user_on_exit = <function 1>,
  _user_on_start = <function 2>,
  args = { "-C", "/home/kghost/workspace/diffx/core/target/js-2.13/streams/_global/update/_global/streams", "check-ignore", "--stdin" },
  command = "git",
  enable_handlers = true,
  enable_recording = true,
  interactive = true,
  pid = "EMFILE: too many open files",
  stderr = <userdata 2>,
  stdin = <userdata 3>,
  stdout = <userdata 4>,
  user_data = {},
  writer = { "/home/kghost/workspace/diffx/core/target/js-2.13/streams/_global/update/_global/streams/out" },
  <metatable> = <1>{
    __index = <table 1>,
    _create_uv_options = <function 3>,
    _execute = <function 4>,
    _pipes_are_closed = <function 5>,
    _prepare_pipes = <function 6>,
    _reset = <function 7>,
    _shutdown = <function 8>,
    _stop = <function 9>,
    add_on_exit_callback = <function 10>,
    after = <function 11>,
    after_failure = <function 12>,
    after_success = <function 13>,
    and_then = <function 14>,
    and_then_on_failure = <function 15>,
    and_then_on_failure_wrap = <function 16>,
    and_then_on_success = <function 17>,
    and_then_on_success_wrap = <function 18>,
    and_then_wrap = <function 19>,
    chain = <function 20>,
    chain_status = <function 21>,
    co_wait = <function 22>,
    is_job = <function 23>,
    join = <function 24>,
    new = <function 25>,
    pid = <function 26>,
    result = <function 27>,
    send = <function 28>,
    shutdown = <function 29>,
    start = <function 30>,
    stderr_result = <function 31>,
    sync = <function 32>,
    wait = <function 33>
  }
}
stack traceback:
	.../myNeovimPackages/start/plenary.nvim/lua/plenary/job.lua:404: in function '_execute'
	.../myNeovimPackages/start/plenary.nvim/lua/plenary/job.lua:447: in function 'start'
	../neo-tree.nvim/lua/neo-tree/git/ignored.lua:138: in function 'mark_ignored'
	...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:113: in function 'returned_function'
	...mPackages/start/plenary.nvim/lua/plenary/async/async.lua:31: in function 'callback_or_next'
	...mPackages/start/plenary.nvim/lua/plenary/async/async.lua:45: in function 'saved_callback'
	...ackages/start/plenary.nvim/lua/plenary/async/control.lua:126: in function 'tx'
	...imPackages/start/plenary.nvim/lua/plenary/async/util.lua:66: in function 'callback'
	...mPackages/start/plenary.nvim/lua/plenary/async/async.lua:25: in function 'callback_or_next'
	...mPackages/start/plenary.nvim/lua/plenary/async/async.lua:45: in function 'callback'
	...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:195: in function <...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:186>
stack traceback:
	[C]: in function 'error'
	.../myNeovimPackages/start/plenary.nvim/lua/plenary/job.lua:404: in function '_execute'
	.../myNeovimPackages/start/plenary.nvim/lua/plenary/job.lua:447: in function 'start'
	../neo-tree.nvim/lua/neo-tree/git/ignored.lua:138: in function 'mark_ignored'
	...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:113: in function 'returned_function'
	...mPackages/start/plenary.nvim/lua/plenary/async/async.lua:31: in function 'callback_or_next'
	...mPackages/start/plenary.nvim/lua/plenary/async/async.lua:45: in function 'saved_callback'
	...ackages/start/plenary.nvim/lua/plenary/async/control.lua:126: in function 'tx'
	...imPackages/start/plenary.nvim/lua/plenary/async/util.lua:66: in function 'callback'
	...mPackages/start/plenary.nvim/lua/plenary/async/async.lua:25: in function 'callback_or_next'
	...mPackages/start/plenary.nvim/lua/plenary/async/async.lua:45: in function 'callback'
	...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:195: in function <...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:186>
```

We match this field in the error
`pid = "EMFILE: too many open files",`
and display a nice informative message to the user as well as gracefully tear down the whole process. Probably an idea for another PR though.

cseickel:
> One more thing, when there is not enough file descriptors left in the system, plenary job will fail with following error:

Yeah, that's a known issue for another PR.

cseickel (approved):

PR #968, opened 2023-06-03
fix: do not restore bad state for top line when opening neotree

cseickel:
this is a follow up to #953, which was a fix for #781

This fixes the problem where if you open `:Neotree current` in a window that has a terminal that has a lot of lines, then the tree will be scrolled down as far as it can go. I didn't really trace the execution, I just moved the save and restore of the window top line within the same conditionals as the saving of the currently focused node. It make sense that whatever situation would prevent a save/restore of the current node would also apply to the window scroll position.

FYI: @ghostbuster91


ghostbuster91:
Right, that totally makes sense :+1:

Issue #879, opened 2023-04-17
Tracking issue for Document Symbol source


nhat-vo:
As merged in #841, we now have an additional `document_symbol` source. This source is still rather simple, and more features will be added in the future. The original features are listed in the description of #841, and this issue will track additional features to be implemented.

- [ ] UI:
	- [ ] Follow cursor: just like "follow current file", always have the symbol I'm currently on selected.
- [ ] Commands
	- [ ] Hover docs
	- [ ] Call hierarchy 
	- [ ]  Fuzzy search/filter
- [ ] CoC Support (I don't use CoC, so if anyone would like to help with this, please feel free to do so 😅)
- [ ] Workspace symbols support #593 (as @cseickel mentioned, it would be better to finish the other features first before this)

nhat-vo:
> Can we also throw support for [lspkind.nvim](https://github.com/onsails/lspkind.nvim) into this? I attempted to do this myself in `document_symbol.kinds` but `lspkind` doesn't return Highlight Groups (as far as I can tell) for the items so I don't have a great way to do this. It would be nice if we could simply pass something from `lspkind`  to update _just_ the icons, while letting the `document_symbol` source manage the Highlight Group

At first, I did try to use Lspkind's icons, but this plugin provides the icon for a different set of kinds ([Completion kinds](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) vs. [Symbol kinds](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol)). So it's not very straightforward to add this. Instead, it is quite straightforward to just copy their symbols, modify them accordingly, and paste into your config.

> Edit: After reading #593, is my request the same as that?

I don't think so, #593's request is an integration between the `filesystem` and `documents_symbols` sources, rather than a icon config like this.

nhat-vo:
> It _should_ be, but I don't know the appropriate highlight groups being used in the background (IE, it doesn't appear they are exposed). Which means that I have to set the icons and make my own highlight group for each icon.

Oh I see. Just to confirm, do you mean that the default highlight groups should be used when only the names are specified? If so, I think it should be quite an easy tweak - changing `tbl_extend` to `tbl_deep_extend`. I'll add it this weekend.

miversen33:
Yes exactly! If a highlight group isn't provided, the default should be used.

nhat-vo:
[...]

PR #972, opened 2023-06-06
fix(document_symbols): add_blank_line_at_top

nhat-vo:
Fix the bug when `add_blank_line_at_top = true` as mentioned in #879.


cseickel (approved):

MunifTanjim:
btw, `add_blank_line_at_top` can probably be supported by doing `tree:render(2)` instead of adding extra dummy nodes to the tree.

cseickel:
Thanks for the heads up @MunifTanjim! We should probably do this at some point, but I tried to do it quickly and it didn't work out. There seems to be other code in neo-tree that assumes certain things about the number of nodes in the tree vs linenr. Also, any sort of re-render in the same window just added new copies of the tree below the old one.

If anyone else wants to take that on, go for it.

Issue #924, opened 2023-05-11
BUG: Indentation for files seems to be broken.


oneslash:
I am working on a project where the file `main.tf` is located in the following path `AWS/eu-west-1/INFLUX/production/main.tf`.

The tree renders `main.tf` as it is on the same level as `production` folder

![Screenshot 2023-05-11 at 2 43 40 PM](https://github.com/nvim-neo-tree/neo-tree.nvim/assets/15870134/5b55131c-fd01-460f-b9f8-a9011dbbe3a9)

pysan3:
I cannot reproduce it.

Does it happen with other file name as well?

Your screen shot does not show the left end of the buffer. How far is the indentation?

It would be very nice if you could provide more details.

alptibet:
+1

oneslash:
@alptibet good catch, yes indeed it works well if I create another file or folder.

pysan3:
> i think I have found a pattern here

@alptibet Thanks a lot for the detailed explanation and screenshots. But I still cannot reproduce this error...

Could you send your config options for neo-tree you are using now? (CC @oneslash)

I'm also wondering how you managed to bring `src` folder at the same indent as `~/DEV/test`. I thought that option was deleted quite some time ago.

oneslash:
@pysan3, hi thank you 🙏 

I use vanilla Astronvim config, https://github.com/AstroNvim/AstroNvim/blob/main/lua/plugins/neo-tree.lua

pysan3:
@oneslash 

Quick update. Thanks for the link and I was able to reproduce the error.

The cause was this line and I will look into how to fix it.

https://github.com/AstroNvim/AstroNvim/blob/a137a014ef579122482d05cc8a58fd5f0dcc916d/lua/plugins/neo-tree.lua#L22

alptibet:
@pysan3 , sorry for late response. I also use vanilla astronvim config.

pysan3:
Thanks for both of you guys. 

I was able to diagnose the error and will fix it when I've got time. Please wait till this weekend or so ;)

oneslash:
@pysan3, thank you very much!

alptibet:
@pysan3, thank you!

pysan3:
Closing with #977

PR #977, opened 2023-06-09
fix(renderer): no_next_padding to fix indentation set to 1

pysan3:
## TLDR

Indentation is broken when parent `indent_marker` is not shown in the node since it is the last element in the tree.

This happens only and only if value is exactly 1. See the indent of `./src/B/{a,b}.js`.

- `= 0`

image

- `= 1`

image

- `= 2` (`>= 2` all works)
image

## Expected

When the indent is set to 1, the indent should only show `indent_marker` (or `last_indent_marker`) so that the indent width will be 1.
- `= 1`
image

---

### Main Issue

Closes https://github.com/nvim-neo-tree/neo-tree.nvim/issues/924

### Detailed Explanation

https://github.com/nvim-neo-tree/neo-tree.nvim/pull/864#issuecomment-1578027693

### Fix

https://github.com/nvim-neo-tree/neo-tree.nvim/pull/864#issuecomment-1579093376

### DIFF
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -420,7 +420,11 @@ M.indent = function(config, node, state)
       end
     end
 
-    table.insert(indent, { text = char .. string.rep(" ", spaces_count), highlight = highlight })
+    table.insert(indent, {
+      text = char .. string.rep(" ", spaces_count),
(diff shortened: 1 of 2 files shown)


cseickel (approved):
Nice and simple, looks good!

PR #986, opened 2023-06-11
fix: Preventing removal of necessary root separator on Windows

### LINKED ISSUE #985: Fails to list files when at the root of a drive on Windows
Steps to reproduce on Windows:
1. Open neovim at the root of a drive (C:\ for example)
2. Open neo-tree

No files will be listed.

The culprit seems to be [this](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/v2.x/lua/neo-tree/command/init.lua#L103):
```lua
if #args.dir > 1 and args.dir:sub(-1) == utils.path_separator then
	args.dir = args.dir:sub(1, -2)
end
```
It's removing the slash from "C:\ ", making it "C:" and the slash appears to be necessary for neo-tree here.

This fixes the issue for me:
```lua
local root_len = vim.fn.has("win32") and 3 or 1
if #args.dir > root_len and args.dir:sub(-1) == utils.path_separator then
	args.dir = args.dir:sub(1, -2)
end
```
I believe a root drive path on Windows (not including WSL) will always be 3 characters (drive letter, colon and backlash). If you have more than 26 drives, you have to use volume mount points.

Aeceon:
This fixes #985

When a root path was supplied as an argument to the Neotree command or to execute on Windows, the necessary separator was removed. ":Neotree C:\", for instance, would become "C:" and this resulted in an empty listing.

Instead of testing if the length of args.dir is greater than 1, this now compares it to 3 on Windows while remaining 1 on other operating systems, including WSL.
```lua
   -- Root paths on Windows have 3 characters ("C:\")
    local root_len = vim.fn.has("win32") == 1 and 3 or 1
    if #args.dir > root_len and args.dir:sub(-1) == utils.path_separator then
      args.dir = args.dir:sub(1, -2)
    end
```
I added a comment as well just to clarify what was going on. Though, maybe it'd be better to have a constant. In utils, there's an is_windows variable set (which includes WSL). Maybe a root_len should be set here as well? I'm going to leave it at this for now, though.

### DIFF
--- a/lua/neo-tree/command/init.lua
+++ b/lua/neo-tree/command/init.lua
@@ -100,7 +100,9 @@ M.execute = function(args)
   -- Handle setting directory if requested
   local path_changed = false
   if utils.truthy(args.dir) then
-    if #args.dir > 1 and args.dir:sub(-1) == utils.path_separator then
+    -- Root paths on Windows have 3 characters ("C:\")
+    local root_len = vim.fn.has("win32") == 1 and 3 or 1
+    if #args.dir > root_len and args.dir:sub(-1) == utils.path_separator then
       args.dir = args.dir:sub(1, -2)
     end
     path_changed = state.path ~= args.dir


miversen33:
This works on my windows machine. It took me a bit to wrap my head around _why_ it worked lol but ya it works :) @cseickel  this should be good to merge.

miversen33 (approved):

cseickel (approved):

cseickel:
> This works on my windows machine. It took me a bit to wrap my head around _why_ it worked lol but ya it works :) @cseickel this should be good to merge.

Thanks for checking @miversen33!

Issue #973, opened 2023-06-07
Provide default value for prompt "File not in cwd. Change cwd to"


BruceChanJianLe:
Hello there, I went through the neo-tree.txt but I couldn't find the answer.

I am wondering is it possible to provide a default value for the prompt when running ''Neotree focus reveal" command. Current, we can only force it to yes if "reveal_force_cwd" is used.

Is it possible to set it to no? If the file cannot be found on the current tree, then just toggle open the tree without revealing it?

Thank you.

wsdjeg:
I just open an PR about this, #992

BruceChanJianLe:
Hi @wsdjeg,

Thanks for looking into this matter! I shall close this issue when the PR is successful :D

cseickel:
This is really unnecessary. You can get the behavior you want by setting `follow_current_file = true`. With that option, every time you open the tree it will reveal the current file if it is within the cwd, but it won't prompt you if it is not.

In that case you don't need to use the `reveal` option in the command.

The only downside is that it will always follow the current file, not just when you first open the tree. If that's nmot what you want, then someone will need to create a new PR to add the new variation of reveal. Maybe `reveal_without_change_cwd`?

BruceChanJianLe:
That's what I'm looking for! Sorry for not understanding the use of follow_current_file at the first place. Now that you have explained, it makes perfect sense.

I am happy with it following the file for now. If others would like a `reveal_without_change_cwd` perhaps they could open a new PR.

Thank you both @wsdjeg and @cseickel for your time!

jakobkhansen:
Does using `follow_current_file` to suppress the prompt still work? I cannot seem to get it working on my end, it always prompts for y/n to change cwd.

cseickel:
@jakobkhansen if you don't want to be prompted, don't use a `reveal` command. 

If you want it to change cwd without prompting, then use the `reveal_force_cwd` flag instead. If you want it to follow the current file but never change cwd, then just use the `follow_current_file` option without specifying the `reveal` flag.

BruceChanJianLe:
> follow_current_file

My configuration still seems to work for me. I do `Neotree focus` with `follow_current_file` as true.

PR #992, opened 2023-06-15
feat(config): add force_change_cwd opt

wsdjeg:


cseickel:
This looks like a duplication of: https://github.com/nvim-neo-tree/neo-tree.nvim#reveal_force_cwd

Do we really need two ways to do the same thing?

wsdjeg:
thanks, does `[reveal_force_cwd](https://github.com/nvim-neo-tree/neo-tree.nvim#reveal_force_cwd)` option fix #973?

cseickel:
> thanks, does `reveal_force_cwd` option fix #973?

No, it does not. That was a request for the opposite actually, `reveal_without_change_cwd`, which is really unnecessary. I'll respond to that issue with more details.

PR #1007, opened 2023-06-21
fix(renderer): delete content of buffer before rendering

### LINKED ISSUE #1000: `:Neotree current` creates "ghost" file tree after opening a file
`:Neotree current` creates "ghost" file tree after opening a file.
But it disappears after you press `q` to exit and reopen with `:Neotree current`.

Recording 2023-06-20 at 05 36 51

pysan3:
After thinking of it for a while I think this is the best way to fix the issue.

With the old code, the content, or even the entire buffer was deleted with `close_all`, so this change would not create much overhead theoretically.

I've tested it with edge-cases I could think of but I'd be glad if you could test the change.

closes #1000
details https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1000#issuecomment-1601258265

Best,
pysan3

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -1086,6 +1086,7 @@ draw = function(nodes, state, parent_id)
 
   -- Create the tree if it doesn't exist.
   if not parent_id and not M.window_exists(state) then
+    M.close(state)
     create_window(state)
     create_tree(state)
   end


cseickel:
I have another fix for this... I'll ping you on the new PR.

pysan3:
Thanks!!

I'm not on a PC right now but I'll test it when I get back. 

Thanks again for all the debugging and fix.

Issue #928, opened 2023-05-13
[feature request] remember width


gmankab:
`:Neotree`
`:vertical resize 15`
`:Neotree toggle`
`:Neotree toggle`

it forgets the width of buffer

it will be cool if it remembers the width after toggles and after restarting nvim

nhat-vo:
I am planning to do some refactoring with the state manager, which will enable consistent width on `toggle`. 

However, implementing this for neovim restart is very hard with only neovim’s builtin feature. I would suggest using some session management plugins with which you could create hooks to remember and restore the width.

paulwyszynski:
I was about to ask the same question. Is this is still in progress?

zbyju:
This would be a great feature!

PR #1008, opened 2023-06-22
[WIP] feat(manager): allow window-specific state 

nhat-vo:
Following up on the conversation in #752, the goal is to prevent the flickering caused by the closing and reopening of the `Neotree` window when we change the source. Here are my current thoughts:
- Add a table to `manager` to keep track of the states by position. Whenever the source is created, the manager will use the old (if exists) window. This avoids the flickering window and also allow window-specific settings such as width (in #928)
- I am also thinking about the idea of having multiple state at the same time, instead of just one. This allows more flexible layouts, for example, `filesystem` on the left, `document_symbol` on the right, and `diagnostics` at the bottom. This might, however, come with some complexity especially when two of the same source coexist (e.g. the user uses the source selector to cycle through the sources).

I want to get some suggestions before doing the implementation, as this might requires many changes. What do you think @cseickel, @pysan3?

PS: I set the `v2.x` as the base branch as it does not allow me to set `main` since there is no commits yet. I will switch back afterwards.

### DIFF
--- a/lua/neo-tree/utils.lua
+++ b/lua/neo-tree/utils.lua
@@ -551,13 +551,14 @@ end
 ---@param bufnr number|nil The buffer number to open
 M.open_file = function(state, path, open_cmd, bufnr)
   open_cmd = open_cmd or "edit"
-  if open_cmd == "edit" or open_cmd == "e" then
+  local cmd_for_buf = { edit = "b", e = "b", split = "sb", sp = "sb" }
+  if open_cmd[cmd_for_buf] ~= nil then
     -- If the file is already open, switch to it.
     bufnr = bufnr or M.find_buffer_by_name(path)
     if bufnr <= 0 then
       bufnr = nil
     else
-      open_cmd = "b"
+      open_cmd = cmd_for_buf[open_cmd]
     end
   end
 
--- a/lua/neo-tree/utils.lua
+++ b/lua/neo-tree/utils.lua
@@ -551,14 +551,17 @@ end
 ---@param bufnr number|nil The buffer number to open
 M.open_file = function(state, path, open_cmd, bufnr)
   open_cmd = open_cmd or "edit"
-  local cmd_for_buf = { edit = "b", e = "b", split = "sb", sp = "sb" }
-  if open_cmd[cmd_for_buf] ~= nil then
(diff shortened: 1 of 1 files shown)


cseickel:
> * Add a table to manager to keep track of the states by position. Whenever the source is created, the manager will use the old (if exists) window. This avoids the flickering window and also allow window-specific settings such as width (in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/928)

Nothing really needs to be done here from the manager point of view, it will give you the state you request no problem. The issue is that the command module will close all windows in the target position before proceeding to open the new one. It does this to avoid the extra work that comes with properly transitioning a window from one source to another. The fix will be to stop that close command, and deal with all the bugs that will result from using another source's window.

> * I am also thinking about the idea of having multiple state at the same time, instead of just one. This allows more flexible layouts, for example, `filesystem` on the left, `document_symbol` on the right, and `diagnostics` at the bottom. This might, however, come with some complexity especially when two of the same source coexist (e.g. the user uses the source selector to cycle through the sources).

It already works that way. There is one state per source per tab for all positions other than current, and one state per source per window for position = current.

nhat-vo:
> It already works that way. There is one state per source per tab for all positions other than current, and one state per source per window for position = current.

I am not sure what you mean by this part. I understand that in that case, the user would need to manually create the 3 different windows, and then populate each window using `position = "current"`, is that correct?

cseickel:
> > It already works that way. There is one state per source per tab for all positions other than current, and one state per source per window for position = current.
> 
> I am not sure what you mean by this part. I understand that in that case, the user would need to manually create the 3 different windows, and then populate each window using `position = "current"`, is that correct?

No, not at all, try it out:
```
:Neotree diagnostics bottom
:Neotree filesystem left
:Neotree document_symbols right
```

You are restricted to one of each source per tab, but nothing is stopping you from showing a different source on each edge.

cseickel:
This was closed automatically by github on the last release because it contained the commits in main. I'll leave it closed though because I think this was on the wrong track.

I have some free time and I'd like to tackle this problem right now. I'm not 100% sure yet, but I think I am going to add a new service that keeps track of windows across sources.

PR #1011, opened 2023-06-24
fix: reuse windows and buffers when possible, prevents flashing when switching

### LINKED ISSUE #713: Changing source from Any source to Filesystem closes and reopens neo-tree instead of switching source

cseickel:
fixes #713

This adds logic to reuse Neotree windows between sources instead of always closing the window when opening a new source. It also improves upon the concept of reusing the buffer when possible instead of recreating it from scratch each time. These are primarily optimizations for quickly switching between sources when they are shown as sidebars.

### DIFF
--- a/lua/neo-tree/command/init.lua
+++ b/lua/neo-tree/command/init.lua
@@ -126,9 +126,9 @@ M.execute = function(args)
 
   -- All set, now show or focus the window
   local force_navigate = path_changed or do_reveal or git_base_changed or state.dirty
-  if position_changed and args.position ~= "current" and current_position ~= "current" then
-    manager.close(args.source)
-  end
+  --if position_changed and args.position ~= "current" and current_position ~= "current" then
+  --  manager.close(args.source)
+  --end
   if do_reveal then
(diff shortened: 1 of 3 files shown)


cseickel:
@nhat-vo I think this fixes the flashing window problem when switching sources.

andrevmatos:
I think this may have broken Neotree on different tabs.
Now they aren't independent anymore, and if I leave a tree open on tab 1, switch to tab 2 and try to open Neotree from there, it'll focus tab 1 and its Neotree, instead of opening a new, independent one in tab 2.

nhat-vo:
@cseickel This looks great, thanks a lot! Sorry, I haven't got much time in the last few weeks. I will have a look at the issue with the different tabs

PR #1015, opened 2023-06-26
feat(filesystem): show window immediately without waiting for async scan

### LINKED ISSUE #744: Render the window right away, even when async

cseickel:
closes #744


cseickel:
The problem with this is that it breaks all of the tests... I need to figure out a way to indicate to the tests that the buffer/tree is not ready for testing yet when they get created before the async scan is finished.

PR #1016, opened 2023-06-27
implemented changes intially created by @daog1 and addresses the comm…

thomascrha:
This PR continues where #938 left off - I really want this feature - but it seems that @daog1 has abandoned the change.

This PR is a carbon copy of his changes with the addition of the suggestions made in the comments.


lua/neo-tree/defaults.lua
```diff
@@ -367,6 +367,8 @@ local config = {
       ["p"] = "paste_from_clipboard",
       ["c"] = "copy", -- takes text input for destination, also accepts the config.show_path and config.insert_as options
       ["m"] = "move", -- takes text input for destination, also accepts the config.show_path and config.insert_as options
+      ["Y"] = "copy_path",
+      ["gy"] = "copy_abspath",
```
cseickel, line 371: The only problem here is that the commands are defined at the global level so that it applies to all sources, but it is not really a command that will work universally. It should be mapped within the `filesystem` and `buffers` commands instead.

It also makes sense to use it in the `git_status` source.

lua/neo-tree/sources/buffers/commands.lua
```diff
@@ -52,6 +52,14 @@ M.copy = function(state)
   cc.copy(state, redraw)
 end
 
+M.copy_path = function(state)
```
cseickel, line 55: I don't think this is needed at all because there is a
```
cc._add_common_commands(M)
```
call at the end of the file, which will add this command anyhow. This pattern of wrapping common commands is only needed when we need to alter the common command in some way, such as by adding a `redraw` callback.

lua/neo-tree/sources/common/commands.lua
```diff
@@ -481,6 +481,34 @@ M.copy = function(state, callback)
...
 
+---Copies a node relative path to clipboard.
+---@param state table The state of the source
+M.copy_path = function(state)
+  local node = state.tree:get_node()
+  if node.type == "message" then
+    return
+  end
+  local pwdpath = state.path
+  local content = node.path
```
cseickel, line 492: Since this is a common command, there's no guarantee that either of these exist for all sources.
```suggestion
  local pwdpath = state.path or ""
  local content = node.path or node:get_id()
```

lua/neo-tree/sources/filesystem/commands.lua
```diff
@@ -32,6 +32,14 @@ M.copy = function(state)
...
 end
 
+M.copy_path = function(state)
+  cc.copy_path(state)
+end
+
+M.copy_abspath = function(state)
+  cc.copy_abspath(state)
+end
+
```
cseickel, line 42: These can be deleted as well.

lua/neo-tree/defaults.lua
qubaitian: Forgive my ignorance, and would you explain it in more detail,thx alot

thomascrha:
Sorry - I stopped using neo-tree soon after this PR: I discovered telescope and harpoon and never looked back.

Issue #999, opened 2023-06-19
Howto keep current tree when switch to last buffer?


tmpm697:
I have this map:
```
map("n", "<leader>;", [[:e#<cr>]])
```
so when i press `leader` and `;`, neovim will switch to last buffer, this works but neo-tree will also close all other expanded folders, but this happens only on the first time i press `leader` and `;`, the second it's not happened, idk why.

steps:
```
tree folders:
fder1
  file1
  file2
fder2
  file3
  file4
1. select file1 and then select file3
2. select file4
3. press leader and ; to switch last buffer
4. observe now fder1 is collapsed and only fder2 is expaneded and buffer switched to file3
5. now select fder1 and select file1
6. select file2
7. switch buffer
8. observe that the tree structure is kept (fder1 and fder2 both expaneded), buffer switched to file1
```

so is there a way to always keep current tree structure in neo-tree despite of switch whatever buffer? especially using above mapping to switch last buffer?

i doubt that the switch buffer function from neovim causes this.

pysan3:
May I have the list of your current neo-tree configuration.

I bet this is caused by some specific combination of options such as `follow_current_file`, and it'll be easier to diagnose.

tmpm697:
> May I have the list of your current neo-tree configuration.
> 
> I bet this is caused by some specific combination of options such as `follow_current_file`, and it'll be easier to diagnose.

sure thing, this is my neotree config:
```
return {
	"nvim-neo-tree/neo-tree.nvim",
	dependencies = {
		"nvim-lua/plenary.nvim",
		"MunifTanjim/nui.nvim",
	},
	cmd = "Neotree",
	config = function()
		vim.cmd([[ let g:neo_tree_remove_legacy_commands = 1 ]])
		require("neo-tree").setup({
			renderers = {
				message = {
					{ "indent", with_markers = true },
					{ "name", highlight = "NeoTreeMessage" },
				},
			},
			use_default_mappings = false, -- no default mappings enable, define keybidings yourself
			close_if_last_window = false, -- Close Neo-tree if it is the last window left in the tab
			popup_border_style = "rounded",
			enable_git_status = true,
			enable_diagnostics = true,
			sort_case_insensitive = false, -- used when sorting files and directories in the tree
			sort_function = nil, -- use a custom function for sorting files and directories in the tree
			default_component_configs = {
				container = {
					enable_character_fade = true,
				},
				indent = {
					indent_size = 2,
[...]

Issue #903, opened 2023-04-27
[FEATURE] follow_current_file focuses on a file but closes other expanded folders


rootex-:
Hi, thanks for your effort,

Your plugin is great, yet there are some features that stop my migration from nvim-tree to your plugin. 

One of them:
    follow_current_file=true successfully focuses on a file but closes other expanded folders.
 
Many find it convenient to work with expanded filesystem in the project. 

Focusing is a great feature but it would be great to preserve previously expanded stuff.

pysan3:
Scrolling through the endless list of issues I found that we had the exact same issue here lol. 

Duplicate of #999. 

@rootex- closing this issue since we are having discussion over on #999. Sorry that I didn’t notice this for quite a while.

rootex-:
Nevertheless, I have already migrated to your plugin and feel happier than ever  🙌 the best file tree for me so far ❤

Issue #717, opened 2023-01-25
git_status expands all nodes whenever it is interacted with


braxtons12:
I haven't been able to find a past issue/discussion about this, but this seems like something that should have been solved already, so please point me to the configuration setting I'm missing if that's the case.

As is, whenever the git_status is interacted with, it automatically expands all nodes in it, regardless of the previous state they were in. For example, if I git_add_file a file, the entire tree re-expands. This can be really jarring and annoying when you have things like generated docs in your project filling your git_status view, and them constantly being re-expanded every time you try to perform a git action.

pysan3:
This is actually a feature hardcoded for example here. https://github.com/nvim-neo-tree/neo-tree.nvim/blob/70d3daa22bf24de9e3e284ac659506b1374e3ae2/lua/neo-tree/sources/buffers/lib/items.lua#L100-L102.

This behavior is actually the default for all sources except `filesystem`.
I discovered this while resolving #999.

I will add code so that it respects the `<source-name>.follow_current_file.leave_dirs_open` for all sources proposed here https://github.com/nvim-neo-tree/neo-tree.nvim/issues/999#issuecomment-1601887559, which will satisfy this feature request as a side product.

cseickel:
The problem here is that all sources except filesystem are designed to explicitly open all folders automatically. This just males the most sense for 99% of situations.

> I will add code so that it respects the <source-name>.follow_current_file.leave_dirs_open for all sources proposed here https://github.com/nvim-neo-tree/neo-tree.nvim/issues/999#issuecomment-1601887559, which will satisfy this feature request as a side product.

That doesn't really fit here at all because this feature has no relation to `follow_current_file`. This would be a new feature for the `git_status` source explicitly which would probably be called `auto_expand_folders` or something similar and default to true. In this case, if it existed then @braxtons12 could set that to false and then it could behave like the filesystem source and start off collapsed but keep track of what was actually explicitly opened.

Alternatively, `git_status` could have a new collection called `explicitly_closed` folders where it keeps track of the opposite action and ensures that only those folders are forced closed after each refresh. That's probably more what is needed here and would not require a new option.

I don't think we should consider making any changes to any other sources, as it does not make sense at all to have closed folders for the `buffers` source.

pysan3:
> This just makes the most sense for 99% of situations.

[...]

PR #1017, opened 2023-06-27
feat(sources): add follow_current_file.leave_dirs_open option

pysan3:
## Details

https://github.com/nvim-neo-tree/neo-tree.nvim/issues/999#issuecomment-1601887559

Closes: #999
Related: #903

## Changes

- `filesystem.follow_current_file` -> `filesystem.follow_current_file.enabled`
  - Added in `deprecations.lua`

## Notes

- `filesystem.follow_current_file.leave_dirs_open` defaults to `false`, which does not change the default behavior.
- Although `:Neotree reveal` automatically opens dirs to focus the current file, `filesystem.follow_current_file.enabled == false` && `:Neotree reveal` => does not respect `leave_dirs_open` nor remember dirs opened by `reveal`. This option only works when `enabled == true` is explicitly set.
- Changed `buffers.follow_current_file.enabled` as well for compatibility, but `buffers.follow_current_file.leave_dirs_open` is never referenced since in buffers source, all dirs are expanded by default. https://github.com/nvim-neo-tree/neo-tree.nvim/blob/70d3daa22bf24de9e3e284ac659506b1374e3ae2/lua/neo-tree/sources/buffers/lib/items.lua#L100-L102

### DIFF
--- a/lua/neo-tree/sources/filesystem/init.lua
+++ b/lua/neo-tree/sources/filesystem/init.lua
@@ -34,10 +34,10 @@ end
 local follow_internal = function(callback, force_show, async)
   log.trace("follow called")
   if vim.bo.filetype == "neo-tree" or vim.bo.filetype == "neo-tree-popup" then
-    return
+    return false
   end
   local path_to_reveal = manager.get_path_to_reveal()
-  if not utils.truthy(path_to_reveal) then
+  if not path_to_reveal or not utils.truthy(path_to_reveal) then
     return false
   end
 
--- a/lua/neo-tree/sources/filesystem/init.lua
+++ b/lua/neo-tree/sources/filesystem/init.lua
@@ -70,22 +70,19 @@ local follow_internal = function(callback, force_show, async)
 
   log.debug("follow file: ", path_to_reveal)
   local show_only_explicitly_opened = function()
-    local eod = state.explicitly_opened_directories or {}
+    state.explicitly_opened_directories = state.explicitly_opened_directories or {}
     local expanded_nodes = renderer.get_expanded_nodes(state.tree)
     local state_changed = false
     for _, id in ipairs(expanded_nodes) do
-      local is_explicit = eod[id]
-      if not is_explicit then
-        local is_in_path = path_to_reveal:sub(1, #id) == id
-        if is_in_path then
-          is_explicit = true
-        end
-      end
-      if not is_explicit then
-        local node = state.tree:get_node(id)
-        if node then
-          node:collapse()
-          state_changed = true
+      if not state.explicitly_opened_directories[id] then
+        if path_to_reveal:sub(1, #id) == id then
+          state.explicitly_opened_directories[id] = state.follow_current_file.leave_dirs_open
+        else
+          local node = state.tree:get_node(id)
+          if node then
+            node:collapse()
+            state_changed = true
+          end
         end
       end
       if state_changed then
@@ -146,7 +143,7 @@ M._navigate_internal = function(state, path, path_to_reveal, callback, async)
     fs_scan.get_items(state, nil, path_to_reveal, callback)
   else
     local is_current = state.current_position == "current"
-    local follow_file = state.follow_current_file
+    local follow_file = state.follow_current_file.enabled
       and not is_search
       and not is_current
       and manager.get_path_to_reveal()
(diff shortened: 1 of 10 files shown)


pysan3:
As I've stated here: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/717#issuecomment-1609378192, I would like to have some kind of `keep_dirs_expansion` option throughout all sources to make neo-tree never automatically `node:expand()` or `node:collapse()`.

However, for the sake of this PR, I will not be doing that and just focus on extending `explicitly_opened_directories` to change the behavior of `filesystem` source.

I've deleted the reference to #717 in the above comment, and will only be closing #999.

### Testing

@tmpm697 could you test my PR if it works for you as well?

```lua
return {
  "pysan3/neo-tree.nvim",
  branch = "leave-dirs-open",
  version = false,
  dependencies = {
    { "MunifTanjim/nui.nvim" },
    { "nvim-tree/nvim-web-devicons" },
  },
  cmd = { "Neotree" },
  keys = {
    { "<Leader>e", "<Cmd>Neotree toggle<CR>", remap = false, silent = true, desc = "<Cmd>Neotree toggle<CR>" },
  },
  opts = {
    filesystem = {
      follow_current_file = {
        enabled = true,
        leave_dirs_open = true,
      },
    },
  },
}

### Notes

@cseickel I don't mean to rush you but I would love to hear your opinion here: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/717#issuecomment-1609378192.

File: lua/neo-tree/sources/filesystem/init.lua [hunk 1]
```diff
@@ -34,10 +34,10 @@ end
 local follow_internal = function(callback, force_show, async)
   log.trace("follow called")
   if vim.bo.filetype == "neo-tree" or vim.bo.filetype == "neo-tree-popup" then
-    return
+    return false
   end
   local path_to_reveal = manager.get_path_to_reveal()
-  if not utils.truthy(path_to_reveal) then
+  if not path_to_reveal or not utils.truthy(path_to_reveal) then
```
cseickel, line 40: Shouldn't `utils.truthy` cover whatever is being handled by `not path_to_reveal`?

lua/neo-tree/utils.lua
```diff
@@ -349,13 +349,17 @@ end
 ---@param sourceObject table The table to set a value in.
 ---@param valuePath string The path to the value to set.
 ---@param value any The value to set.
-M.set_value = function(sourceObject, valuePath, value)
+---@param force_tabulize boolean? Overwrite existing non-table parents with a table along the way.
```
cseickel, line 352: I don't think that the `force_tabulize` needs to exist because there is no reason for it to ever be false.

lua/neo-tree/sources/filesystem/init.lua
pysan3: Indeed, but it shows diagnostic error so I simply added it. And theoretically it's faster because now we might skip one function call.

lua/neo-tree/utils.lua
pysan3: I thought someone might want to call it with `pcall` with `force_tabularize = false` to bail out if type is different, and so that it doesn't overwrite the old value.

cseickel: Following that logic, I would make it `allow_type_change` to cover all cases where setting a value may change a type and not just specifically handle the table case only. I don't see this being needed at all though so I would remove the option because it just adds unnecessary complication.

pysan3: Got it! I'll just remove it and make it the default behavior :)

lua/neo-tree/sources/filesystem/init.lua
pysan3: Well, it's a bit more difficult than that.

```lua
local x = function_return_0_or_nil()
-- type(x) == integer or nil
if not utils.truthy(x) then
  return
end
-- type(x) == integer or nil (still thinks x can be nil)
local y = x + 1 -- RED ERROR

--------
local x = function_return_0_or_nil()
-- type(x) == integer or nil
if not x or not utils.truthy(x) then
  return
end
-- type(x) == integer (x is always integer here)
local y = x + 1 -- HAPPY LSP ;)
```

cseickel: I don't have any issue from diagnostics with this code:
```lua
  local path_to_reveal = manager.get_path_to_reveal()
  if not utils.truthy(path_to_reveal) then
```

`path_to_reveal` is `string|nil`, `utils.truthy` accepts `any`. What error do you get?

cseickel (approved):

Issue #1018, opened 2023-06-28
BUG: Reproduction Configuration missing Plenary dependency


miversen33:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

NA

### Operating System / Version

NA

### Describe the Bug

When you try to use the reproduction config provided with this template, Neo-tree complains that it can't find plenary. Which is correct as plenary is not listed as a dependency in the config. PR incoming to correct this

### Screenshots, Traceback

```lua
Failed to run `config` for neo-tree.nvim
.../lua/neo-tree/sources/filesystem/lib/filter_external.lua:3: module 'plenary.job' not found:
^Ino field package.preload['plenary.job']
cache_loader: module plenary.job not found
cache_loader_lib: module plenary.job not found
^Ino file './plenary/job.lua'
^Ino file '/__w/neovim/neovim/.deps/usr/share/luajit-2.1.0-beta3/plenary/job.lua'
^Ino file '/usr/local/share/lua/5.1/plenary/job.lua'
^Ino file '/usr/local/share/lua/5.1/plenary/job/init.lua'
^Ino file '/__w/neovim/neovim/.deps/usr/share/lua/5.1/plenary/job.lua'
^Ino file '/__w/neovim/neovim/.deps/usr/share/lua/5.1/plenary/job/init.lua'
^Ino file './plenary/job.so'
^Ino file '/usr/local/lib/lua/5.1/plenary/job.so'
^Ino file '/__w/neovim/neovim/.deps/usr/lib/lua/5.1/plenary/job.so'
^Ino file '/usr/local/lib/lua/5.1/loadall.so'
^Ino file './plenary.so'
^Ino file '/usr/local/lib/lua/5.1/plenary.so'
^Ino file '/__w/neovim/neovim/.deps/usr/lib/lua/5.1/plenary.so'
^Ino file '/usr/local/lib/lua/5.1/loadall.so'
# stacktrace:
  - /neo-tree.nvim/lua/neo-tree/sources/filesystem/lib/filter_external.lua:3
  - /neo-tree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:6
  - /neo-tree.nvim/lua/neo-tree/sources/manager.lua:6
  - /neo-tree.nvim/lua/neo-tree.lua:5
  ```

### Steps to Reproduce

[...]

PR #1019, opened 2023-06-28
1018 add plenary to issue template

miversen33:
Addresses #1018 

Pretty simple change but still probably worth a quick review to make sure it doesn't just "works on my machine" :)


cseickel (approved):
The updated config works for me!

Issue #889, opened 2023-04-21
Paths with parentheses on Windows cannot be opened


SilverMira:
Hi, I noticed this bug while using Neotree in a Sveltekit project, where routes can be grouped in a folder with parentheses in its name. This bug only happens on Windows presumably due to the backslash path separator is escaping the opening parentheses somehow.

![image](https://user-images.githubusercontent.com/66930495/233666722-7d796401-b3b0-42d7-b9f2-74d346bfdf4c.png)

Notice how Neotree opened a buffer at `outer(inner)\file.txt` instead of `outer\(inner)\file.txt`. Additionally, I discovered that `:e outer\(inner)\file.txt` in a `nvim --clean`  instance is also unable to open the file, whereas `:e outer\\(inner)\\file.txt` can, perhaps a fix can be implemented by escaping all backslashes from the path on Windows before opening the file.

miversen33:
This is a plenary issue. As an example, I have the following directory structure on a windows machine
```
C:\Users\miver\Application Data
C:\Users\miver\Cookies
C:\Users\miver\dummy(file).txt
C:\Users\miver\Local Settings
C:\Users\miver\My Documents
C:\Users\miver\OneDrive
C:\Users\miver\Recent
C:\Users\miver\SendTo
C:\Users\miver\Start Menu
C:\Users\miver\Templates
```

And running the following lua code
```lua
local Path = require("plenary.path")
local scandir = require("plenary.scandir").scan_dir
print(scandir(vim.loop.cwd(), { depth = 1}))
```

I get the following output

```lua
{ "C:\\Users\\miver\\Application Data", "C:\\Users\\miver\\Cookies", "C:\\Users\\miver\\dummy(file).txt", "C:\\Users\\miver\\Local Settings", "C:\\Users\\miver\\My Documents", C:\\Users\\miver\\OneDrive", C:\\Users\\miver\\Recent", "C:\\Users\\miver\\SendTo", "C:\\Users\\miver\\Start Menu", "C:\\Users\\miver\\Templates" }
```

Notice that `dummy(file).txt` is not escaped? I will open an issue with plenary. This is not something we should fix (within Neo-tree anyway).

SilverMira:
With the directory structure that you have, I believe Neo-tree can open `dummy(file).txt` perfectly fine. 

In my testing, the bug only occurs if the path contains a `(` directly after the path separator `\` no matter if the segment that starts with `(` is a file or directory.

[...]

PR #1023, opened 2023-07-02
889 windows path escape

miversen33:
Addresses #889 
Tagging @SilverMira @treturner for testing

Tagging @cseickel and @pysan3 for review. The change is pretty small, thanks to @SilverMira for the little code snippet :) Basically just checking to see if we are on windows and if we are, doing a glob substitute on the path to ensure that anything that isn't escaped **is** escaped now.


treturner (commented):
That code snippet worked for me. Can open the file from neotree and treesitter. Thanks @miversen33 and @SilverMira

pysan3:
Hi @miversen33 

This is just my preference but we have `windowize_path` function around L900 which is exactly the reverse of `escape_path` so I think it should be written next to it?

Also, we already have `M.is_windows` defined at around L690 and I think we should use that instead of `has(win32)`? I'm not sure tho.

miversen33:
> Hi @miversen33 
> 
> This is just my preference but we have `windowize_path` function around L900 which is exactly the reverse of `escape_path` so I think it should be written next to it?
> 
> Also, we already have `M.is_windows` defined at around L690 and I think we should use that instead of `has(win32)`? I'm not sure tho. 

I have no issue in updating the position and using that function (I didn't even think to see if that existed lol). I'll update the PR with these requests

miversen33:
I've made the requested changes. Lemme know if there are any other requests :)

SilverMira (approved):
Yup, that does fix it for neo-tree

cseickel (approved):

PR #1024, opened 2023-07-03
style: update some nerdfont symbols since they were deprecated

Alan-Kuan:
Hi,

As some Nerd Font symbols used in this plugin have been updated and moved to new codepoints since v3.0.0 release of Material Design Icons, I did my best to find every deprecated symbols and manually updated them.

Please tell me if there's anything I missed.

Thank You!

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ use {
       vim.fn.sign_define("DiagnosticSignInfo",
         {text = " ", texthl = "DiagnosticSignInfo"})
       vim.fn.sign_define("DiagnosticSignHint",
-        {text = "", texthl = "DiagnosticSignHint"})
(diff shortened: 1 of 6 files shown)


miversen33:
Is this different than #909?

Alan-Kuan:
> Is this different than #909?

Oh! I didn't notice there was a PR working on this issue. Sorry about that.

Besides, Both changes are almost the same while I missed the symbols in `tests/neo-tree/ui` and the symbol for "String" lsp kind in `lua/neo-tree/defaults.lua`.

#909 did the job well. However, the look of the symbol for "Buffers" was changed. I found a better replacement (`󰈚`) which looks the same as the old one. I think we can just make this small change to #909.

Anyway, I'll just closed this PR.

PR #1035, opened 2023-07-06
docs: fix outdated diagnostics signs info

jensgreen:
README has the correct info, but the help file is not up to date.
 
For reference, the README with accurate info and the relevant code:

- https://github.com/nvim-neo-tree/neo-tree.nvim/blob/ee0c058f3fa36675c8fb93cb44b94ed678db6cfa/README.md?plain=1#L109-L119
- https://github.com/nvim-neo-tree/neo-tree.nvim/blob/ee0c058f3fa36675c8fb93cb44b94ed678db6cfa/lua/neo-tree/sources/common/components.lua#L89-L100


cseickel (approved):
Thanks!

PR #1043, opened 2023-07-09
feat(ui/inputs): enable normal mode in rename popup

LoneExile:
This PR addresses the issue discussed in [#1039](https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/1039), where users were unable to switch to normal mode in the NeoTree rename popup by pressing the escape key.

**Benefits:**

This change significantly improves the user experience by allowing users to switch to normal mode without closing the popup, and provides a way to close the popup from normal mode.

Moreover, it offers additional benefits when dealing with long file names. Users can now leverage normal mode commands such as `dw` (delete word), `dt`. (delete until dot), or `dd` (delete line) to edit file names more efficiently. This makes the renaming process faster and more intuitive, especially for those who are accustomed to Vim's normal mode operations.

Please let me know if you want me to change anything or if you would like this to be an option in the plugin.

### DIFF
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -22,7 +22,7 @@ local config = {
   enable_modified_markers = true, -- Show markers for files with unsaved changes.
   enable_opened_markers = true,   -- Enable tracking of opened files. Required for `components.name.highlight_opened_files`
   enable_refresh_on_write = true, -- Refresh the tree when a file is written. Only used if `use_libuv_file_watcher` is false.
-  enable_normal_mode_for_inputs = true, -- Enable normal mode for input dialogs when <esc> is pressed.
+  enable_normal_mode_for_inputs = true, -- Enable normal mode for input dialogs.
   git_status_async = true,
   -- These options are for people with VERY large git repos
   git_status_async_options = {
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -22,6 +22,7 @@ local config = {
   enable_modified_markers = true, -- Show markers for files with unsaved changes.
   enable_opened_markers = true,   -- Enable tracking of opened files. Required for `components.name.highlight_opened_files`
   enable_refresh_on_write = true, -- Refresh the tree when a file is written. Only used if `use_libuv_file_watcher` is false.
+  enable_normal_mode_for_inputs = true, -- Enable normal mode for input dialogs when <esc> is pressed.
   git_status_async = true,
   -- These options are for people with VERY large git repos
   git_status_async_options = {
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -22,7 +22,7 @@ local config = {
   enable_modified_markers = true, -- Show markers for files with unsaved changes.
   enable_opened_markers = true,   -- Enable tracking of opened files. Required for `components.name.highlight_opened_files`
   enable_refresh_on_write = true, -- Refresh the tree when a file is written. Only used if `use_libuv_file_watcher` is false.
-  enable_normal_mode_for_inputs = true, -- Enable normal mode for input dialogs.
+  enable_normal_mode_for_inputs = false, -- Enable normal mode for input dialogs.
   git_status_async = true,
   -- These options are for people with VERY large git repos
   git_status_async_options = {
--- a/README.md
+++ b/README.md
@@ -123,6 +123,7 @@ use {
         popup_border_style = "rounded",
         enable_git_status = true,
         enable_diagnostics = true,
+        enable_normal_mode_for_inputs = false, -- Enable normal mode for input dialogs.
         open_files_do_not_replace_types = { "terminal", "trouble", "qf" }, -- when opening files, do not use windows containing these filetypes or buftypes
         sort_case_insensitive = false, -- used when sorting files and directories in the tree
         sort_function = nil , -- use a custom function for sorting files and directories in the tree 
--- a/lua/neo-tree/ui/inputs.lua
+++ b/lua/neo-tree/ui/inputs.lua
@@ -14,6 +14,12 @@ M.show_input = function(input, callback)
  local config = require("neo-tree").config
   input:mount()
 
+  if config.enable_normal_mode_for_inputs then
+    vim.schedule(function()
+      vim.cmd("stopinsert")
+    end)
+  end
+
   input:map("i", "<esc>", function()
     vim.cmd("stopinsert")
     if not config.enable_normal_mode_for_inputs then
--- a/lua/neo-tree/ui/inputs.lua
+++ b/lua/neo-tree/ui/inputs.lua
@@ -11,10 +11,14 @@ local should_use_popup_input = function()
 end
 
 M.show_input = function(input, callback)
+ local config = require("neo-tree").config
   input:mount()
 
   input:map("i", "<esc>", function()
     vim.cmd("stopinsert")
+    if not config.enable_normal_mode_for_inputs then
+      input:unmount()
+    end
   end, { noremap = true })
 
   input:map("n", "<esc>", function()
@@ -33,7 +37,7 @@ M.show_input = function(input, callback)
     if callback then
       callback()
     end
-  end, { once = true })
+  end, { once = true }) 
 end
 
 M.input = function(message, default_value, callback, options, completion)
--- a/lua/neo-tree/ui/inputs.lua
+++ b/lua/neo-tree/ui/inputs.lua
@@ -14,7 +14,7 @@ M.show_input = function(input, callback)
  local config = require("neo-tree").config
   input:mount()
 
-  if config.enable_normal_mode_for_inputs then
+  if config.enable_normal_mode_for_inputs and input.prompt_type ~= "confirm" then
     vim.schedule(function()
       vim.cmd("stopinsert")
     end)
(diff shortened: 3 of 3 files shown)


cseickel:
I think I like this change. It is now 2x ESC to cancel which is not that big of a change and is pretty intuitive even if someone had no idea this change existed. I think we can make this the new "normal" (no pun intended).


I will leave this open for a day just to see if anyone else wants to comment.

sgoudham:
Hey :wave:, I was the one that raised the initial discussion. Thanks @LoneExile for raising this PR! 

I'm not sure if making it the new default is the best for all existing users that are used to the current behaviour. I'm sure we can all agree that users in the neovim ecosystem _love_ having choice so it might be best having it as a configuration option? Something like `close_popup_on_escape` which is default `true` ? 

Either way, happy to see it approved and merged in!

LoneExile:
Since this change has a global scope and affects all `file`, `buffer`, and `git` operations, I propose adding the option in the main setup configuration like so:

```lua
require("neo-tree").setup({
  -- existing configuration options...
  
  close_popup_on_escape = true,  -- Close the popup if the user presses escape
  
  -- more existing configuration options...
})
```

With this new option, the function would look something like this:

```lua
M.show_input = function(input, callback)
  local config = require("neo-tree").config -- << added loaded config
  input:mount()

  input:map("i", "<esc>", function()
    vim.cmd("stopinsert")
    if config.close_popup_on_escape then -- << added this conditional
      input:unmount()
    end
  end, { noremap = true })

  input:map("n", "<esc>", function()
    input:unmount()
  end, { noremap = true })

  input:map("n", "q", function()
    input:unmount()
  end, { noremap = true })

  input:map("i", "<C-w>", "<C-S-w>", { noremap = true })

  local event = require("nui.utils.autocmd").event
  input:on({ event.BufLeave, event.BufDelete }, function()
    input:unmount()
    if callback then
      callback()
    end
  end, { once = true })
end
```

What do you think about this approach?

bew:
> What do you think about this approach?

On the config naming, `close_popup_on_escape` doesn't really says that is only affects insert' escape, and that popups are still closed on _normal_ escape when set to `false`

cseickel:
> On the config naming, `close_popup_on_escape` doesn't really says that is only affects insert' escape, and that popups are still closed on _normal_ escape when set to `false`

That's true.

**IF** there was an option, I think `enable_normal_mode_for_inputs` may be the most descriptive. I'm definitely leaning toward no option at all though, because a double escape is not enough of an inconvenience to warrant another option and I think more people would want normal mode as an option.

LoneExile:
@bew. I agree that the configuration option name should be more descriptive.

However, I've been asking myself, **'Does this really make sense?'** I mean, going straight to insert mode after pressing `r` doesn't feel quite right to me. It seems a bit inconvenient to have to use the `arrow keys` for navigation or hold down `backspace` for a while to delete text. imho, it would make more sense to default to normal mode.

Here's an alternative solution that defaults to normal mode:
```lua
M.show_input = function(input, callback)
  input:mount()

  vim.schedule(function() -- << wait for input to be mounted
    vim.cmd("stopinsert")
  end)

  input:map("i", "<esc>", function()
    vim.cmd("stopinsert")
  end, { noremap = true })

  input:map("n", "<esc>", function()
    input:unmount()
  end, { noremap = true })

  input:map("n", "q", function()
    input:unmount()
  end, { noremap = true })

  input:map("i", "<C-w>", "<C-S-w>", { noremap = true })

  local event = require("nui.utils.autocmd").event
  input:on({ event.BufLeave, event.BufDelete }, function()
    input:unmount()
    if callback then
      callback()
    end
  end, { once = true })
end
```

Please let me know which solution you would prefer for me to implement. @cseickel

cseickel:
@LoneExile I think it makes sense to have an option that determines whether inputs start in insert mode or not. We could also use this to decide whether `<esc>` in insert mode cancels the input or just leaves insert mode. That way, if normal mode is not enabled, the behavior is unchanged. 

If normal mode is enabled, and is controlled by an option, should `<esc>` still be used to cancel? Maybe yes but I'm not sure only because I tend to spam extra escapes just to make sure when I am in that mental mode.

LoneExile:
I've updated the code to reflect the new behavior. Here's the flow:

Pressing `r` will now take you to insert mode. Pressing `esc` in insert mode will switch to normal mode. Pressing `esc` again in normal mode will exit the input.

cseickel:
I think that sounds correct to me. @bew and @sgoudham any comments?

CC: @miversen33 @nhat-vo @pysan3

miversen33:
My personal opinion is that we're splitting hairs here. A pop-up window is meant to be treated as a single line input. Yes technically it's a buffer but you shouldn't be acting like it is (and thus is generally intuitive to assume that an input box is automatically in insert mode).

That said, I don't really see any harm in adding a configuration option to exit insert mode without closing the input window immediately (double pressing seems intuitive for me). I also agree this new exit method should be the default, as it is still intuitive (in the vim sense of exit insert mode first, then exit buffer). That said, if we allow an input to be used in normal mode, we will need to verify that the input behaves" correctly" when it receives commands such as `:q`, `:w`, etc. I don't know what correctly means exactly here, but if we are going to allow the input line to be treated as a buffer, I suspect we will have users that will instinctively use vim commands to interact with it's contents.

Just my 2 cents, do or don't, I'm fine either way :)

lua/neo-tree/defaults.lua
```diff
@@ -22,6 +22,7 @@ local config = {
   enable_modified_markers = true, -- Show markers for files with unsaved changes.
   enable_opened_markers = true,   -- Enable tracking of opened files. Required for `components.name.highlight_opened_files`
   enable_refresh_on_write = true, -- Refresh the tree when a file is written. Only used if `use_libuv_file_watcher` is false.
+  enable_normal_mode_for_inputs = true, -- Enable normal mode for input dialogs.
```
cseickel, line 25: I think the default should be `false` for this option.

We should also include this option in the quickstart of the README.

LoneExile: I've made the requested changes, but I have another concern about the input box for the `delete file` operation. I believe it's not necessary to revert to normal mode since it only requires a `y` or `n` input. 

Here's my suggested code update for the confirm input to always retain the old behavior:

```lua
  if config.enable_normal_mode_for_inputs and input.prompt_type ~= "confirm" then
    vim.schedule(function()
      vim.cmd("stopinsert")
    end)
  end
  
  input:map("i", "<esc>", function()
    vim.cmd("stopinsert")
    if not config.enable_normal_mode_for_inputs or input.prompt_type == "confirm" then
      input:unmount()
    end
  end, { noremap = true })

  ```
  
  By adding `input.prompt_type = "confirm"` to the M.confirm function below before calling `M.show_input(input)`, we can handle this. Alternatively, we can check `input._.prompt._content == " y/n: "` without adding a value to the input.

If you think the current implementation is fine, I'm okay with that as well.

**Comment by pysan3** [comment]
Thanks @LoneExile , I'm really looking forward to this functionality. 

I'm sorry I'm currently not on a computer and cannot test it myself but here's some concerns and ideas that came to my mind. 

1. After going into normal mode, does that mean we can add a new line in the prompt with `o` or `O`? If yes, what happens to the file name? And if no, what happens with that keybind? And I guess there's more keybinds that'll cause some issues. 

2. I thought it would be nice if we could configure the keybind to go to normal mode. In normal editing, I usually use `jk` to exit insert mode so if I can map that in the prompt, it'll be intuitive for me. (Just a random idea tho)

3. Another random idea, can we place the cursor right before the extension by default? I mean that the cursor is set at `foo/bar|.lua`.  (This might be unrelated to this PR tho)

**Comment by LoneExile** [comment]
@pysan3 , Thank you for taking the time to review my PR.

> After going into normal mode, does that mean we can add a new line in the prompt with `o` or `O`? If yes, what happens to the file name? And if no, what happens with that keybind? And I guess there's more keybinds that'll cause some issues.

The buffer type for the input is set to `prompt`. This should ensure the correct behavior. You can refer to `:h prompt-buffer` for more details.

> I thought it would be nice if we could configure the keybind to go to normal mode. In normal editing, I usually use `jk` to exit insert mode so if I can map that in the prompt, it'll be intuitive for me. (Just a random idea tho)

Yes, if you have set `jk` to act as `esc` in `insert mode`, that should work in the prompt as well.

> Another random idea, can we place the cursor right before the extension by default? I mean that the cursor is set at `foo/bar|.lua`. (This might be unrelated to this PR tho)

I believe we can achieve the desired cursor positioning by using `vim.cmd('norm $F.h')`. This should place the cursor right before the extension, as you suggested.

File: lua/neo-tree/defaults.lua (hunk 1 above)
Comment (cseickel) [reply lua/neo-tree/defaults.lua]: I agree that y/n prompts should be in insert mode regardless of the `enable_normal_mode_for_inputs` setting. I also think adding `input.prompt_type = "confirm"` is the best way to accomplish that.

**Comment by miversen33** [comment]
> At that point, I would probably be in favor of adding a neo_tree_popup_enter event that people can use to do whatever customization they want, like placing the cursor, visual selection of the filename, add mappings, etc.

Moar events == gud

**Comment by cseickel** [comment]
This looks good. Thanks @LoneExile for doing this and taking the time to work through the details. Also thanks to everyone who provided input.

**Comment by zmunk** [comment]
This behavior has been deprecated in commit https://github.com/nvim-neo-tree/neo-tree.nvim/commit/403a9c5014624dc517b850a360f345340139307c


The following in the docs explains the new way to enable the behavior:
```
    {
      event = "neo_tree_popup_input_ready",
      ---@param args { bufnr: integer, winid: integer }
      handler = function(args)
        -- map <esc> to enter normal mode (by default closes prompt)
        -- don't forget `opts.buffer` to specify the buffer of the popup.
        vim.keymap.set("i", "<esc>", vim.cmd.stopinsert, { noremap = true, buffer = args.bufnr })
      end,
    }
```

https://github.com/nvim-neo-tree/neo-tree.nvim/wiki/Recipes#events

Issue #993, opened 2023-06-16
Neotree displays no files when in WSL2 root


jamesonBradfield:
Hi, this might be my misunderstanding of how WSL's file system works, But when I am in the root of my WSL2 filesystem and I open Neovim and open Neotree only one file is shown and its the Root parent folder \\wsl$\Ubuntu\root. No other files are shown nor can I navigate the filesystem. Anyways, Thanks for your time. Here is my Neotree Config written in lazy.
```
-- Unless you are still migrating, remove the deprecated commands from v1.x
vim.cmd([[ let g:neo_tree_remove_legacy_commands = 1 ]])

return {
  "nvim-neo-tree/neo-tree.nvim",
  version = "*",
  dependencies = {
    "nvim-lua/plenary.nvim",
    "nvim-tree/nvim-web-devicons", -- not strictly required, but recommended
    "MunifTanjim/nui.nvim",
  },
  keys = {
    { "<leader>ft", "<cmd>Neotree toggle<cr>", desc = "NeoTree" },
  },
  config = function ()
    require('neo-tree').setup {}
  end,
}
```
![Screenshot (23)](https://github.com/nvim-neo-tree/neo-tree.nvim/assets/113628216/cd1ac84c-9ad6-4153-a383-a7292b30a733)

miversen33:
This is odd. I created a dummy config (wrapping your Neo-tree config) that uses lazy to lazy load Neotree. I cannot recreate this (using this config) in my wsl instance.
![image](https://github.com/nvim-neo-tree/neo-tree.nvim/assets/2640668/debc8c27-319a-41fb-962f-8bf209e525dd)

Config. Don't be worried about your current setup, this will create an entirely new directory for lazy to dump into, and clean up said directory after it closes out

Let me know if you are seeing the same results with the below config. If not, I suspect you have an issue higher than Neo-tree. If so... well I don't know but we will cross that road if/when we get there :)

```lua
-- mini.lua
-- config
-- Use this via the `--clean` and `-u` flags in neovim
-- EG: nvim --clean -u mini.lua

-- Set this to false if you don't want this to wipe your plugin dir each
-- installation
vim.cmd([[ let g:neo_tree_remove_legacy_commands = 1 ]])
local sep = vim.loop.os_uname().sysname:lower():match('windows') and '\\' or '/' -- \ for windows, mac and linux both use \
local do_clean = true
[...]

PR #1051, opened 2023-07-14
fix: normalize file paths (#993)

miversen33:
image

Pretty simple fix but appears to resolve this. Tagging @emilgedda for testing and @cseickel / @pysan3  for review :)


cseickel (approved):
Nice and simple!

cseickel:
This is so straightforward I think we can just merge it to `main` and let it be tested there.

PR #1054, opened 2023-07-14
Fix: replace the usage of `get_icon` function to support the `overrid…

yebt:
…e_by_extension`

To support the `override_by_extension` items inside the icons, and complex extensions of files, try to use the `get_icon` function without force the extension

### DIFF
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -281,7 +281,7 @@ M.icon = function(config, node, state)
     local success, web_devicons = pcall(require, "nvim-web-devicons")
     if success then
       local ext = node.ext and node.ext:lower() or nil
-      local devicon, hl = web_devicons.get_icon(node.name, ext)
+      local devicon, hl = web_devicons.get_icon(node.name)
       icon = devicon or icon
       highlight = hl or highlight
     end


yebt:
It happens to me that when the extension is sent, it forces the assigned icon to fit the calculated extension and not the file, so complex extensions may be skipped.
An example of this is with the "app.component.spec.ts" file, which is interpreted as a file with a "ts" extension; or with the file "view.blade.php" which is interpreted as a "php" file.

When trying to assign specific icons to these files via "nvim-web-devicons" in your setup:

```lua
...
      override_by_extension = {
        ["log"] = {
          icon = "",
          color = "#81e043",
          name = "Log",
        },
        ["ts"] = {
          icon = "",
          color = "#519aba",
          cterm_color = "74",
          name = "Ts",
        },
        ["spec.ts"] = {
          icon = "",
          color = "#bf672d",
          cterm_color = "74",
          name = "SpecTs",
        },
        ["blade.php"] ={
          icon = "󰫐",
          color = "#f13d30",
          name = "BladePhp",
        }
      },
...
```
however, the modified icons are not recognized.



With the suggested change, the file extension to display is recognized and associated with the specific icons:

cseickel:
This was already merged to `main` in #1044 and will be part of v3.0 when it is released. If you want this functionality now, you can use the `main` branch but beware that there may be breaking changes from v2.x.

PR #1060, opened 2023-07-16
docs: Fix lazy.nvim instructions

tradiff:
Lazy uses `dependencies` instead of `requires` 😄

Source: [Lazy.nvim migration guide](https://github.com/folke/lazy.nvim#packernvim)


cseickel (approved):
oops... Thanks!

PR #1061, opened 2023-07-16
docs: fix help tags for custom commands and mappings

tradiff:
Sorry for the PR spam, I noticed this other typo immediately after you merged the previous PR 😂


cseickel:
I'm going to add onto this a little, because it turns out that the `neo-tree-custom-commands-global` tag does not work, and when I looked into it I realized that the table of contents for this section doesn't really make sense.

cseickel (approved):

PR #1081, opened 2023-08-04
feat(command): add source `last` refers to the last executed source

ofseed:
Implement #775 , add a source option `last` to allow users to open the last source used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,6 +482,7 @@ What to show. Can be one of:
 | filesystem | Show a file browser. DEFAULT |
 | buffers    | Show a list of currently open buffers. |
 | git_status | Show the output of `git status` in a tree layout. |
+| last       | Equivalent to the last source used |
 
 #### `position`
 Where to show it, can be one of:
--- a/lua/neo-tree/command/init.lua
+++ b/lua/neo-tree/command/init.lua
@@ -11,6 +11,9 @@ local M = {
   complete_args = completion.complete_args,
 }
 
+-- Store the last source used for `M.execute`
+M._last_source = nil
+
 ---Executes a Neo-tree action from outside of a Neo-tree window,
 ---such as show, hide, navigate, etc.
 ---@param args table The action to execute. The table can have the following keys:
@@ -60,6 +63,12 @@ M.execute = function(args)
 
   -- The rest of the actions require a source
   args.source = args.source or nt.config.default_source
+
+  -- Restore the last source used if requested
+  if args.source == "last" then
+    args.source = M._last_source or nt.config.default_source
+  end
+  M._last_source = args.source
 
   -- If position=current was requested, but we are currently in a neo-tree window,
   -- then we need to override that.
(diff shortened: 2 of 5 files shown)


ofseed:
Furthermore, considering the `:Neotree` command without a specific source is heavily used when interacting with users and most users don't truly trust the default when they need to open a specific source, I personally recommend setting the`default_source` to `last` in `neo-tree.defaults`.

cseickel (commented):
This is a good idea, but it's not quite complete because it ignores the position that the `last` source was opened in. I don't always use the same position so it's not enough for me to just set a default for every source, and when I use `last` I expect it to be the last **source** and the last **position**.

cseickel:
> Furthermore, considering the `:Neotree` command without a specific source is heavily used when interacting with users and most users don't truly trust the default when they need to open a specific source, I personally recommend setting the`default_source` to `last` in `neo-tree.defaults`.

I don't think you have any way of knowing that, you are just speaking about your own usage here and projecting that on others. 90% of the time I use a mapping and not the command. When I just use `:Neotree` without specifying a source, I expect it to be the "filesystem" source because that is the one I use more often. Also, based on my interactions with users to date I would say that the vast majority of users primarily use the "filesystem" source, so it makes sense that this would be the default.

ofseed:
Of course, this is merely a minor suggestion for the design. Since it could disrupt existing user configurations, I'm not implementing these thoughts in the current PR.

But honestly, I think this design is a little weird. Considering neo-tree could switch between different sources in the same window, some users might be very eager to leverage this feature, and there is a situation:

I opened a neo-tree filesystem window by `Neotree toggle` (no matter with a keymap or enter the command directly), and after a while, I switch to another source like the buffer source. Finally, I want to close the existing neo-tree window by `Neotree toggle` because I think it would do that, but the result is I switch the current window to the filesystem window rather than close it.

What I want to demonstrate is when you toggle, you'd better know what you are toggle, rather than take advantage of the default value. If you assign a key to the Neotree toggle filesystem, it becomes clear that this key is solely for toggling the filesystem, regardless of the current source.

In fact, this PR also solves the previous problem. Because switching to another source would call `execute` again, when you use `Neotree toggle last`, no matter what source you switched after that, `Neotree toggle last` will close this window.

But as I mentioned before, I would not implement it in this PR, it is just a suggestion.

ofseed:
I discovered that implementing this feature might not require extra work. because when I execute Neotree toggle <source> without specifying an explicit position, it defaults to the last position where this source was opened.

So, in essence, Neotree toggle is equivalent to Neotree toggle position=last. The only requirement is to implement source=last, which has been accomplished in this PR.

cseickel:
> I discovered that implementing this feature might not require extra work. because when I execute Neotree toggle without specifying an explicit position, it defaults to the last position where this source was opened.

It does do that most of the time, but not always. It does not work for `position="current"`. If you use the `current` position, it will toggle the default position of the sidebar version instead.

ofseed:
Completed, the position will be last if it is not specified now.

cseickel (approved):
Works perfectly!

PR #1088, opened 2023-08-08
feat: check explicitly opened directories in get_folder_node

nnez140814:
Fix issue when adding/pasting file with following config to an empty folder.

```lua
['a'] = { 'add', config = { insert_as = 'sibling' } },
['p'] = { 'paste_from_clipboard', config = { insert_as = 'sibling' } },
```

1. Expand folder `A`
2. press `a` and add file `a`
3. file `a` will be a sibling of `A` instead of child

image

### DIFF
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -31,7 +31,12 @@ local function get_folder_node(state)
       use_parent = insert_as_global == "sibling"
     end
 
-    local is_open_dir = node.type == "directory" and (node:is_expanded() or node.empty_expanded)
+    local is_open_dir = node.type == "directory"
+      and (
+        node:is_expanded()
+        or node.empty_expanded
+        or state.explicitly_opened_directories[node:get_id()]
+      )
     if use_parent and not is_open_dir then
       return tree:get_node(node:get_parent_id())
     end


miversen33:
While I understand the problem you are trying to address, is this really something that _needs_ to be addressed? What is preventing you from simply navigating up to the parent and creating a new node there (how you would in a traditional file explorer)?

nnez140814:
>  (how you would in a traditional file explorer)?

I preferred adding a file to a folder by right-clicking at everywhere in this opened folder. i.e. If I want to add a file to a folder, I will open the folder first.

From my understanding, `insert_as = 'sibling'` behaves like this:

* Adding a file to a closed node(including normal file and collapsed folder) will cause the new node to become a sibling of the closed node.
* Adding a file to an opened node will cause the new node to become a child of the opened node.

It works in most cases now except an **opened empty folder**. Adding a file to an opened empty folder will cause the new node to become a sibling of the empty folder now. I guess this is a misbehavior.

cseickel:
> It works in most cases now except an **opened empty folder**. Adding a file to an opened empty folder will cause the new node to become a sibling of the empty folder now. I guess this is a misbehavior.

I agree that the behavior you described would be an error. At the same time, it works fine for me without your fix. Also, I'm not sure why you made the change you did. If `node.empty_expanded` was working correctly then the existing code would work. If it is not set correctly, then that would be the thing to fix.

nnez140814:
Oh I misunderstand that. I thought `folder_empty` icon indicates the folder is opened.

 I'm using `filesystem.scan_mode = "shallow"` and empty folder will not empty_expand. Is this behavior expected?

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/38293fe690981aba6cfef5e440f26d8b956d463e/lua/neo-tree/sources/filesystem/init.lua#L426-L427

cseickel:
> Oh I misunderstand that. I thought `folder_empty` icon indicates the folder is opened.
> 
> I'm using `filesystem.scan_mode = "shallow"` and empty folder will not empty_expand. Is this behavior expected?

That is correct, with shallow it will be closed until you open it. A side effect of "deep" is that all empty folders will end up "empty open", although the primary purpose of this is to automatically follow through a set of empty folders until it gets to the end or a none-empty one.

One problem I have is that I don't show a different icon for "empty_open" vs "empty_closed" so it's hard to tell the difference. That's a bug that should be fixed.

nnez140814:
> That is correct, with shallow it will be closed until you open it.

But with shallow I can't open an empty folder with 'toggle_node'. Maybe that's an issue

PR #1098, opened 2023-08-12
fix(manager): reveal_current_file attemp to call nil field

nnez140814:
`reveal_current_file` still tries calling `require("neo-tree").close_all_except(source_name)` which is removed by 78d26168e0b4c7c06a0c59b26c2cd305e64f3d12 and causes this error.

image

Here's my usage:

```
mappings = {
  ['gc'] = function()
    local manager = require('neo-tree.sources.manager')
    manager.reveal_current_file('filesystem')
  end,
},
```

From my understanding, it needs to close all neo-tree window to bring focus to 'current file' and after that `M.get_path_to_reveal()` can find the path.

I tried using `M.close_all_except(source_name)` but since the neo-tree window is not closed `M.get_path_to_reveal()` can't find the path to reveal. So I think `M.close_all()` is the proper choice.

### DIFF
--- a/lua/neo-tree/sources/manager.lua
+++ b/lua/neo-tree/sources/manager.lua
@@ -527,7 +527,7 @@ M.reveal_current_file = function(source_name, callback, force_cwd)
   -- (potentially) different position/node
   state.position.is.restorable = false
 
-  require("neo-tree").close_all_except(source_name)
+  M.close_all()
   local path = M.get_path_to_reveal()
   if not path then
     M.focus(source_name)


cseickel:
Honestly, the method you are trying to fix should have been deprecated and removed as well, I just missed it. It was definitely not meant to be used the way you are using it. i

It will not work as a mapping from a neo-tree window because if you are in the neo-tree window, then there is no "current file" to reveal. What you have done here with 'close_all' is create a hack that closes the window which will focus some other window and make the reveal logic work in that way. I suppose that is predictable if you only ever use a single window layout with no splits, but if that's the case you can also accomplish a similar thing without closing neo-tree with this:
```
vim.cmd("wincmd l")
vim.cmd("Neotree reveal")
```

The key here is that "revealing" the current file can only work if you are executing that action from the window that contains the file you want to reveal. That is why you use:
```
:wincmd l
```
to focus the window to the right of the current one, then:
```
:Neotree reveal
```
to reveal it in neo-tree.

-----

Since this function does not even work right now I don't see much harm in just deleting it before someone else tries to use it. I will go ahead and do that.

PR #1107, opened 2023-08-18
feat!: show file stats in columns, add sort commands

cseickel:
image

image

This was inspired by this discussion: https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/1104

This makes several major changes:

1. It collects stats for every node, IF the new columns are visible or a sort that requires them was activated.
2. It adds new default columns for size, type, modified, and created.
4. It adds a new config for components in a container, called `required_width` that will only render a component when the window width is greater than the `required_width`. **The new columns take advantage of this so that they only show on wide windows.**
5. It adds a more advanced sort system that supports sorting on all visible attributes.
6. It adds a new sub-menu feature to the help menu, for use with multi-key mappings like the new "order_by" commands.

The big question is how many people would be pissed off by adding this stuff and enabling it by default. Collecting the stats may have a performance impact on a very large directory, and displaying them may take up valuable space for people with really long filenames or just look like clutter.


miversen33:
I am unable to test this WIP. Pressing `i` does dump the node stat to the messages which is good, but I cannot see anything additional on the filesystem source.

That said, I would be hesitant to set this as the default for the filesystem source as, as you said, its a pretty big UI change. If I can get it visually to display, I would be curious how much screen real estate it makes the filesystem source take. 

All that said, I really love the idea of us incorporating the stat data into the nodes in some way. I just don't know the best way to visualize it in a "small compact" way as the filesystem source currently does.

In any case, great work :)

cseickel:
> I am unable to test this WIP. Pressing `i` does dump the node stat to the messages which is good, but I cannot see anything additional on the filesystem source.

It will only show if the window is wide enough, wider than a standard sidebar. Did you try making the tree window really wide or showing it with `position="current"`? It's really meant for the current position when there is already a lot of extra space.

miversen33:
Sure didn't, I just assumed because I didn't see any trailing characters that it wasn't working. Expanding the window worked great, good call!

This afternoon I am going to try to slap together a dockerfile that we can use to reliably test your PR against a massive (but empty) filesystem. I would love to see how long it takes to open say 5000 nodes with the extra stat call vs current state.

cseickel:
> This afternoon I am going to try to slap together a dockerfile that we can use to reliably test your PR against a massive (but empty) filesystem. I would love to see how long it takes to open say 5000 nodes with the extra stat call vs current state.

That's awesome, thanks for the assist!

cseickel:
> This afternoon I am going to try to slap together a dockerfile that we can use to reliably test your PR against a massive (but empty) filesystem. I would love to see how long it takes to open say 5000 nodes with the extra stat call vs current state.

FYI - I changed the stat lookup to a lazy load, so it won't be loaded unless those columns are visible.

miversen33:
@cseickel  Here is a dockerfile that you can use to test with a massive amount of items in the filesystem. I don't see any performance issues but I didn't try too much

```dockerfile
# Can be found here as well
# https://pastebin.com/raw/SNWuw4tM
FROM debian:latest

RUN \
    apt-get update -yq 2>&1 > /dev/null ;\
    apt-get install -yq bash git wget 2>&1 > /dev/null ;\
    mkdir -p /tmp/testdir ;\
    cd /tmp/testdir ;\
    for i in $(seq 10) ;\
    do \
      for j in $(seq 1000);\
        do \
          mkdir -p "parent_$i";\
          echo "$j" > "parent_$i/$j.txt";\
      done;\
    done;\
    echo -n '\n\
local root = vim.fn.fnamemodify("./.repro", ":p")\n\
for _, name in ipairs({ "config", "data", "state", "cache" }) do\n\
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name\n\
end\n\
local lazypath = root .. "/plugins/lazy.nvim"\n\
if not vim.loop.fs_stat(lazypath) then\n\
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })\n\
end\n\
vim.opt.runtimepath:prepend(lazypath)\n\
local plugins = {\n\
  "folke/tokyonight.nvim",\n\
}\n\
local neotree_config = {\n\
  "nvim-neo-tree/neo-tree.nvim",\n\
  branch = "add-more-stats",\n\
  dependencies = { "MunifTanjim/nui.nvim", "nvim-tree/nvim-web-devicons", "nvim-lua/plenary.nvim" },\n\
  cmd = { "Neotree" },\n\
  keys = {\n\
    { "<Leader>e", "<Cmd>Neotree<CR>" }, -- change or remove this line if relevant.\n\
  },\n\
  opts = {\n\
  },\n\
}\n\
table.insert(plugins, neotree_config)\n\
require("lazy").setup(plugins, {\n\
  root = root .. "/plugins",\n\
})\n\
vim.cmd.colorscheme("tokyonight")\n\
vim.api.nvim_create_autocmd({ "VimEnter" }, {\n\
	pattern = "*",\n\
	command = ":Neotree show reveal",\n\
})\n\
vim.opt.relativenumber = true\n\
' > /opt/1106-conf.lua;\
    wget -qO /tmp/nvim.tar.gz https://github.com/neovim/neovim/releases/download/stable/nvim-linux64.tar.gz;\
    tar -C /opt/ -xzf /tmp/nvim.tar.gz

WORKDIR /tmp/testdir

ENTRYPOINT /opt/nvim-linux64/bin/nvim --clean -u /opt/1106-conf.lua
```

I don't know how good you are with docker so here is how you use it.

```shell
$ docker build -t neo-tree-1107 https://pastebin.com/raw/SNWuw4tM
$ docker run --rm -it neo-tree-1107
```

cseickel:
Thanks @miversen33 that was helpful. I could tell the difference between opening a folder in sidebar mode vs current when the stats lookup would have been triggered. It was barely perceptible though with 1,000 files so I think that is reasonable.

cseickel:
To disable the newly added columns, add this to your config:
``` lua
require 'neo-tree'.setup {
  default_component_configs = {
    file_size = { enabled = false },
    type = { enabled = false },
    last_modified = { enabled = false },
    created = { enabled = false },
  }
}
```

If the columns are not used, and you don't sort by one of the fields that require it, there will not be any extra file stat lookups.

Issue #910, opened 2023-05-02
Add `empty open`/`symlink (open)` folder icon field


hungvx-dev:
example: 
<img width="208" alt="image" src="https://user-images.githubusercontent.com/85088954/235582398-607aec82-0afc-4032-917c-40b82eb5eb25.png">

<img width="183" alt="image" src="https://user-images.githubusercontent.com/85088954/235582468-cb0268f8-8402-475a-8976-6546a8c947ad.png">

zenoli:
This was introduced intoduced in #600 .
The options are called `folder_empty` and `folder_empty_open`.

PR #1119, opened 2023-08-25
feat(icon): file & folder symlink icon (related to #910)

jackyhevey:


miversen33 (approved):
Not really sure why the styling was changed on several of the functions but outside that very minor complain (which I assume gets fixed on merge by a bot), this looks good :)

lua/neo-tree/defaults.lua
```diff
@@ -200,6 +200,8 @@ local config = {
       folder_open = "",
       folder_empty = "󰉖",
       folder_empty_open = "󰷏",
+      folder_symlink = "",
+      file_symlink = "",
```
cseickel, line 204: This link is almost impossible for me to distinguish from the standard "unknown file" icon. Can we find one that's more obvious?

lua/neo-tree/sources/common/components.lua
```diff
@@ -275,7 +275,11 @@ M.icon = function(config, node, state)
     elseif node:is_expanded() then
       icon = config.folder_open or "-"
```
cseickel, line 276: I realize this problem pre-existed, but now I see that it doesn't make sense that there is only a symlink icon for closed dirs and not open ones. This should be fixed.

lua/neo-tree/sources/common/components.lua
```diff
@@ -275,7 +275,11 @@ M.icon = function(config, node, state)
...
     else
-      icon = config.folder_closed or "+"
+      if node.is_link then
+        icon = config.folder_symlink
```
cseickel, line 279:
```suggestion
        icon = config.folder_symlink or config.folder_closed or "+"
```

This needs fallbacks.

lua/neo-tree/sources/common/components.lua
```diff
@@ -284,6 +288,9 @@ M.icon = function(config, node, state)
       icon = devicon or icon
       highlight = hl or highlight
     end
+    if node.is_link then
```
cseickel, line 291:
```suggestion
    if node.is_link and config.file_symlink then
```

What if someone sets this to `nil` because they don't want to show a symlink icon?

lua/neo-tree/defaults.lua
cseickel: Also, I'm not entirely sure I even want a default icon set for this. Some people might want to see the target's icon and not a generic symlink icon.

Issue #746, opened 2023-02-14
highlighting/showing symlinks differently


figsoda:
Something like this `from-file -> to-file`

cseickel:
That exists as a component already, it's just not enabled by default:
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/245cf1e68840defcc75a16297740f6203f5a045d/lua/neo-tree/defaults.lua#L261-L265

figsoda:
Would I have to copy this whole part to enable it? https://github.com/nvim-neo-tree/neo-tree.nvim/blob/245cf1e68840defcc75a16297740f6203f5a045d/lua/neo-tree/defaults.lua#L252-L271

cseickel:
> Thanks! Would I have to copy this whole part to enable it?

Yes. 

I think I implemented the ability to disable components so that they can be in the default renderer as disabled and then enabled with `default_component_configs` instead of having to define an entire custom renderer, but unfortunately I haven't actually put that ability to use yet.

PR #1122, opened 2023-08-27
refactor!: setup the `symlink_target` component using default config

true2fik:
Currently the only ways to enable the `symlink_target` component is copying the entire the directory and file renderer configs or defining a custom config that includes them. This can be annoying to maintain if the default rendered configuration changes.

This PR uncomments the component in the default config and disables it by default as mentioned in [this comment](https://github.com/nvim-neo-tree/neo-tree.nvim/issues/746#issuecomment-1429090271).

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,9 @@ use {
             enabled = true,
             required_width = 110, -- min width of window required to show this column
           },
+          symlink_target = {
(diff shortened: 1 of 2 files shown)


cseickel (approved):
Excellent idea!

PR #1137, opened 2023-09-05
docs: change README link to misspelled wiki recipes

nurseiit:
Ran into this link that leads nowhere, but the Wiki Has a "Recipies" page which is misspelled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -700,7 +700,7 @@ Details on how to configure everything is in the help file at `:h
 neo-tree-configuration` or online at
 [neo-tree.txt](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/main/doc/neo-tree.txt)
 
-Recipes for customizations can be found on the [wiki](https://github.com/nvim-neo-tree/neo-tree.nvim/wiki/Recipes). Recipes include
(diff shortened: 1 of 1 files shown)


miversen33:
That link doesn't lead "nowhere". Clicking it leads directly to the recipes wiki page. Ironically, your correct actually leads to the mispelled version.

I assume the mispelled version still exists strictly because it was originally mispelled and so older documentation may still point to it.

But the readme update you are proposing is invalid

cseickel:
Oh sorry, when I read this PR I actually went to the wiki and fixed the misspelled page name and I meant to come back here and close the PR with an explanation but I must have gotten distracted and I forgot about it.

Now that the wiki page misspelling has been fixed, this PR is no longer needed. Thanks for finding the issue!

nurseiit:
> That link doesn't lead "nowhere". Clicking it leads directly to the recipes wiki page. Ironically, your correct actually leads to the mispelled version.
> 
> I assume the mispelled version still exists strictly because it was originally mispelled and so older documentation may still point to it.
> 
> But the readme update you are proposing is invalid

😮‍💨 okay, looks like it was updated, I am just trying to help here, thanks!

PR #1148, opened 2023-09-12
Ignore disabled diagnostics

sidequestboy:
If you disable a diagnostic with `vim.diagnostic.disable`, don't count towards the diagnostic count.

### DIFF
--- a/lua/neo-tree/utils/init.lua
+++ b/lua/neo-tree/utils/init.lua
@@ -34,6 +34,10 @@ local diag_severity_to_string = function(severity)
   end
 end
 
+local diagnostic_filter = function(diag)
+  return diag.source == "Lua Diagnostics." and diag.message == "Undefined global `vim`."
+end
+
 local tracked_functions = {}
 M.debounce_strategy = {
   CALL_FIRST_AND_LAST = 0,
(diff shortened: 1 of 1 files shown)


cseickel:
I am hesitating because I wonder if we could instead filter them when pulling the diagnostics to begin with, rather than calling a function to check each one later. Have you looked into that?

sidequestboy:
I agree it doesn't seem great to check the namespace on every diagnostic. iirc, what we can do is filter by namespace to the `vim.diagnostic.get` and do the check only once per namespace using`vim.diagnostic.get_namespaces` - I think we would need one call to get for each namespace for each open buffer. (still probably better than checking on each diagnostic) There didn't seem to be a built in way to filter the diagnostics by disabled status though.

I can rewrite this patch later today!

sidequestboy:
WIP - I changed the shape of the table returned from `get_diagnostics_count` without good reason I think.

sidequestboy:
ok, the most recent force-pushed branch is good to review. Let me know what you think.

lua/neo-tree/utils/init.lua
```diff
@@ -214,39 +218,78 @@ M.human_size = function (size)
...
+    for _, bufnr in ipairs(vim.api.nvim_list_bufs()) do
+      local success, file_name = pcall(vim.api.nvim_buf_get_name, bufnr)
+      if success and not vim.diagnostic.is_disabled(bufnr, ns) then
+        for severity, _ in ipairs(vim.diagnostic.severity) do
+          local diagnostics = vim.diagnostic.get(bufnr, { namespace = ns, severity = severity })
+
+          -- It would be nice to not do this manual filtering below
+          -- Note that lua diagnostics can be suppressed using neodev
+          -- or lua_ls settings instead.
+          -- https://github.com/folke/neodev.nvim
```
cseickel, line 247: I totally agree with this **and** I think it's time to just remove it. This was added before neodev and before I learned about the lua_ls setting.

It's a minor breaking change, but it has such an obvious solution and will remove a lot of wasted work so I think it is worth it.

cseickel (commented):
I think this looks great. Let's just go ahead and remove the legacy lua `vim` global filter though.

cseickel (approved):
Looks good, thanks!

wookayin:
Is this a breaking change or a regression? After this commit I keep getting

```
2023-09-19T01:56:50 Neo-tree  ERROR Error in seed function for vim_diagnostic_changed: .../neo-tree.nvim/lua/neo-tree/utils/init.lua:236: attempt to call field 'is_disabled' (a nil value)
2023-09-19T01:56:55 Neo-tree  ERROR Error in event handler for event state_created[buffers.state_created]: ...neo-tree.nvim/lua/neo-tree/utils/init.lua:236: attempt to call field 'is_disabled' (a nil value)
2023-09-19T01:56:55 Neo-tree  ERROR Error in event handler for event state_created[git_status.state_created]: .../neo-tree.nvim/lua/neo-tree/utils/init.lua:236: attempt to call field 'is_disabled' (a nil value)
```

for neovim 0.8.x.

PR #1155, opened 2023-09-15
feat: Fix paste_default_config

ein-shved:
After moving utils.lua to utils/init.lua the paste_default_config function fails to locate its source root.


cseickel (approved):
Oops... thanks for the fix!

PR #1159, opened 2023-09-19
Add a nil check to vim.diagnostic.is_disabled

sidequestboy:
#1148 accidentally broke 0.8 and earlier - here we add a nil check

I made sure to test and it works 0.8.3


cseickel (approved):

wookayin:
Thanks for the fast follow-up!

PR #1186, opened 2023-10-17
feat: custom text for mappings in help window

coffebar:
With this feature, you can add labels for custom mappings (lua functions).

### Example mapping with text:

```lua
["<c-f>"] = {
  function(state)
     require("telescope.builtin").live_grep({ cwd = context_dir(state) })
  end,
  config = { text = "live grep" }, -- text here
},
```

### Demo

Before

before

After

after

### DIFF
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -406,7 +406,14 @@ For example:
    require("neo-tree").setup({
      window = {
        mappings = {
-         ["A"] = "command_a"
+         ["A"] = "command_a",
+         ["i"] = {
+           function(state)
+             local node = state.tree:get_node()
+             print(node.path)
+           end,
+           label = "print path",
+         },
        }
      },
      filesystem = {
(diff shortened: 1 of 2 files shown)


coffebar:
Changed `label` over `text`. I thought so, too.


```lua
["<c-f>"] = {
  function(state)
     require("telescope.builtin").live_grep({ cwd = context_dir(state) })
  end,
  label = "live grep",
},
```

Will work now

cseickel (commented):
Looks good, but I just realized there are no docs included so no one will ever even know this is possible! Can you add some docs?

coffebar:
Sure, it's been added to the documentation. I believe it should also be added to the 'wiki/Recipes,' but only after this feature has been merged. Not to mislead users.

cseickel (approved):

coffebar:
> Ah, I wish I caught this earlier, would `desc` be a better key for this to align with the `keymap.set` API? Could add it as an also supported key to avoid a breaking change

I don't mind replacing with a breaking change. It's unlikely that anyone but me is already using it

mehalter:
Thanks @coffebar ! Just opened a PR. Let me know if I need to adjust anything. It might also not be considered a breaking change if it hasn't made it into a release yet.

cseickel:
> It might also not be considered a breaking change if it hasn't made it into a release yet.

That's how I see it.

Issue #1170, opened 2023-10-04
Disable winbar when opening neo-tree by command


serhez:
### Did you check the docs?

- [X] I have read all the docs.

### Is your feature request related to a problem? Please describe.

At the moment, the source selector being enabled/disabled is a global option that applied to all windows/commands.

### Describe the solution you'd like.

It would be great to have the option of disabling the winbar source selector (and perhaps also the statusline, for consistency, or simply the source selector alltogether) when calling the neo-tree command. E.g.: `Neotree source=document_symbols position=right winbar=false`. This would allow having different usages for neo-tree, e.g., having a command to open the file tree + git tree on the left position with the source selector enabled, as well as having another command to open the document symbols of the left position with the source selector disabled.

pysan3:
Related: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/895

Issue #895, opened 2023-04-24
[Feature request] Allow hiding the source selector depending on different positions


akinsho:
The main use case for this feature request is that when running neo-tree in a float or more commonly for me the current position if I navigate into a directory, I don't need or want to see the source selector since the only source I want in this case is the filesystem, In this case I want neo tree to be as simple as possible and find the selectors confusing to see in this context. I think it would be nice to optionally not display it in certain circumstances.

miversen33:
IMO the best way to approach this would be to not show the source selector if the user manually specifies a source.

EG

-- No source selector
> :Neotree filesystem

-- Source Selector
> :Neotree
> :Neotree toggle

@cseickel @nhat-vo thoughts?

cseickel:
@miversen33 I don't think that would work for everyone that wants to use the source selector. My first solution would be to see if this special case can be handled with an event handler like `neo_tree_window_before_open` or `neo_tree_buffer_enter`. Maybe the source selector config for the state can be enabled/disabled in one of those events before it even gets applied.

If enough people want this specific condition, then we could add an option to the source selector config to specify what positions it is enabled for.

akinsho:
I think having to manually disable it in a callback of some kind might be complicated for the user to implement since they have to write the logic to determine if they've just entered a directory (I'm fine to do that but think it might tough for anyone new to neovim who wanted the same thing). Unless there was something like:
```lua
 neo_tree_directory_entered = function(state)
   state.source_selector = false
 end
```
Although I think in terms of sane default, seems like people generally wouldn't have a need for the selector when navigating into a directory and rather than there being any hoops, the selector just doesn't show in the specific context of opening neovim in the current window. I think no other (n)vim explorer plugins do anything like this i.e. showing a winbar in this context so presumably generally people won't expect to have one. 


As a side note, it doesn't also actually scale well to being open at this size (each tab is gigantic, although this isn't really my main motivation)

<img width="965" alt="image" src="https://user-images.githubusercontent.com/22454918/234022912-990be4e8-de33-4fdd-b157-2c42bff253cd.png">

cseickel:
> I think having to manually disable it in a callback of some kind might be complicated for the user to implement since they have to write the logic to determine if they've just entered a directory (I'm fine to do that but think it might tough for anyone new to neovim who wanted the same thing)

[...]

PR #1187, opened 2023-10-18
feat(selector): add command option to toggle source selector

pysan3:
Closes: #1170, #895

Add command option `:Neotree selector=true / false` to toggle source selector on command.

Note: @cseickel 
This adds a new field `state.enable_source_selector` which I'm not sure if you are a fan of.

If you don't want to introduce a new field to `state`, I believe it would be pretty much impossible to implement this feature tho.

If you are OK with this PR, I'll add some documentation and change this from _draft_ to a PR.

### DIFF
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -129,10 +129,10 @@ automatically change the directory without prompting. This option implies
 "reveal", so you do not need to specify both.
 
 selector~
-This is a boolean flag. When you specifically set this to false (|selector=false|)
+This is a boolean flag. When you specifically set this to false (`selector=false`)
 neo-tree will disable the |neo-tree-source-selector| for that neo-tree
 instance. Otherwise, the source selector will depend on what you specified in
-the configuration (|config.source_selector.{winbar,statusline}|).
+the configuration (`config.source_selector.{winbar,statusline}`).
 
 CALLING_FROM_LUA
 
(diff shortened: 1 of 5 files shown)


cseickel:
I think that's a nice clean solution. Go ahead and document it.

pysan3:
> I think that's a nice clean solution. Go ahead and document it.

Thanks @cseickel. Updated the docs and is ready to be merged!

doc/neo-tree.txt
```diff
@@ -128,6 +128,12 @@ want to change the current working directory. If you include this flag, it will
 automatically change the directory without prompting. This option implies
 "reveal", so you do not need to specify both.
 
+selector~
+This is a boolean flag. When you specifically set this to false (|selector=false|)
```
cseickel, line 132: The `|word|` syntax is for creating a link to a tag. Tags are created from words in this format:`*word*`. There is no `*selector=false*` tag so this is just a dead link. You probably want to use backticks here instead:

```suggestion
This is a boolean flag. When you specifically set this to false (`selector=false`)
```

doc/neo-tree.txt
```diff
@@ -128,6 +128,12 @@ want to change the current working directory. If you include this flag, it will
...
+neo-tree will disable the |neo-tree-source-selector| for that neo-tree
+instance. Otherwise, the source selector will depend on what you specified in
+the configuration (|config.source_selector.{winbar,statusline}|).
```
cseickel, line 135: Same here, this should probably be backticks:
```suggestion
the configuration (`config.source_selector.{winbar,statusline}`).
```

doc/neo-tree.txt
pysan3: Thanks. As you might have guessed, I'm very new to vim docs lol.
Fixed ;)

cseickel (approved):

Issue #1153, opened 2023-09-15
FEATURE: Add Lua api / mention vim.fn.fnameescape for vim.cmd calls in examples


artfulrobot:
### Did you check the docs?

- [X] I have read all the docs.

### Is your feature request related to a problem? Please describe.

From the docs, it seems the way to programmatically open 'reveal' neotree from lua is something like:

```
 vim.cmd("Neotree reveal_force_cwd " .. path)
```

But this causes crashes as soon as `path` has a space in its filename.

### Describe the solution you'd like.

My feature request is one or more of the following:

1. Document that a more reliable way to do this is
   ```
   vim.cmd("Neotree reveal_force_cwd " .. vim.fn.fnameescape(path)) 
   ```
   
   You could say "oh well, *of course* you should escape the argument!" but it would be far more helpful to include an example like this in the docs for those of us who are not vimscript/lua experts.

2. Offer/document some sort of lua api so we're not going lua → vimscript → lua, e.g. something like
   ```
   require('neotree').dostuff({ reveal_force_cwd = path, toggle = true })
   ```

(2) might already exist but I don't know how to discover it if so.

### Additional Context

FYI this is the code I run, to make the `-` key open a useful neotree:

```
keys = {
  {'-', function()                                               
    local d = vim.fn.getreg('%')                                 
    if (d == '') then                                            
      d = vim.fn.getcwd();                                       
    end                                                          
    vim.cmd("Neotree reveal_force_cwd " .. vim.fn.fnameescape(d))
  end, desc = "NeoTree"},                                        
}                                                                
```

pysan3:
> Offer/document some sort of lua api so we're not going lua → vimscript → lua, e.g. something like

You're pretty close. It's `require('neotree.command').execute({ reveal_force_cwd = path, toggle = true })`.
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/63ebe879ad4798b66d29c0b2c8d04942389d438e/lua/neo-tree/command/init.lua#L46

> Document that a more reliable way to do this is

A PR is welcome!

artfulrobot:
@pysan3 happy to do a pr, but to come back to your comment.

The [docs](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/main/README.md?plain=1#L554-L555) state `reveal_force_cwd` "This is a boolean flag." yet in your example you're passing `path` to this boolean?

artfulrobot:
[...]

PR #1188, opened 2023-10-18
Improve docs: add how to call neotree from Lua

artfulrobot:
This fixes
https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1153

There are two commits.

1. Adds minimal explanation
2. Adds example `keys` entry to do something useful.


doc/neo-tree.txt
```diff
@@ -148,6 +148,38 @@ commands by adding this option before the Neotree config is run:
...
+      else
+        local f = io.open(reveal_file, "r")
+        if (f) then
+          f.close(f)
+        else
+          reveal_file = vim.fn.getcwd()
+        end
+      end
+      require('neo-tree.command').execute({
+        reveal_file, reveal_force_cwd = true
```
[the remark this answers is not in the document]
artfulrobot: (by happenstance, it did work, but it's definitely clearer to do `reveal_file = reveal_file`.)

coffebar (approved):

doc/neo-tree.txt
```diff
@@ -148,6 +148,38 @@ commands by adding this option before the Neotree config is run:
...
+>lua
+  require('neo-tree.command').execute({ ... })
+
+For example, in the `keys` configuration, you could use the following to make
+the `-` key reveal the current file, or if in an unsaved file, the current
+working directory.
+
+>lua
+
+  keys = {
```
cseickel, line 164: What is `keys`? It doesn't have anything to do with Neo-tree and I don't think this code on its own would do anything. I could be missing out on something though.

doc/neo-tree.txt
```diff
@@ -148,6 +148,38 @@ commands by adding this option before the Neotree config is run:
...
+        local f = io.open(reveal_file, "r")
+        if (f) then
+          f.close(f)
+        else
+          reveal_file = vim.fn.getcwd()
+        end
+      end
+      require('neo-tree.command').execute({
+        reveal_file = reveal_file, reveal_force_cwd = true
+      })
```
cseickel, line 179: If we are providing an example as documentation I would change two things here:

1. Put each option on its own line.
2. Provide each optional argument as well as an example.

```suggestion
      require('neo-tree.command').execute({
        action = "focus", -- OPTIONAL, this is the default value
        source = "filesystem", -- OPTIONAL, this is the default value
        position = "left", -- OPTIONAL, this is the default value
        reveal_file = reveal_file, -- path to file or folder to reveal
        reveal_force_cwd = true, -- change cwd without asking if needed
      })
```

doc/neo-tree.txt
```diff
@@ -148,6 +148,38 @@ commands by adding this option before the Neotree config is run:
 >vim
   let g:neo_tree_remove_legacy_commands = 1
 <
+CALLING_FROM_LUA
+
+You can call Neotree commands from your Lua scripts as follows:
+
+>lua
+  require('neo-tree.command').execute({ ... })
+
```
cseickel, line 157: Shouldn't this code block have a closing tag?

```suggestion
<

```

doc/neo-tree.txt
```diff
@@ -148,6 +148,38 @@ commands by adding this option before the Neotree config is run:
...
+        else
+          reveal_file = vim.fn.getcwd()
+        end
+      end
+      require('neo-tree.command').execute({
+        reveal_file = reveal_file, reveal_force_cwd = true
+      })
+    }
+  }
+
```
cseickel, line 182: This is also missing the closing tag.

```suggestion
<

```

doc/neo-tree.txt
```diff
@@ -148,6 +148,38 @@ commands by adding this option before the Neotree config is run:
 >vim
   let g:neo_tree_remove_legacy_commands = 1
 <
```
cseickel, line 150: It's a little confusing to have this come after the deprecated commands and not immediately following the documentation for this particular function. The deprecated commands have actually been removed anyway and they should have been deleted from the docs. Might as well do that now.

```suggestion
```

cseickel (commented):
I wish  more people would submit documentation PRs! I do have some comments and suggestions.

doc/neo-tree.txt
artfulrobot: @cseickel ah. TBH I'm always a bit mystified about this stuff, too. I think Lazy.nvim must interpret it to create key maps. If it's not neo-tree native I'll remove this and just use the normal keymap call instaead. - todo:me.

doc/neo-tree.txt
artfulrobot: yes, good idea.

doc/neo-tree.txt
artfulrobot: First vim help doc I've ever edited, and was guessing how to do it. I'll go with what you say!

doc/neo-tree.txt
artfulrobot: done cf4fd5a

cseickel (approved):
Looks good, thanks!

Issue #1172, opened 2023-10-07
FEATURE: Allow short msg when 'use_popups_for_input = false' #1167


jonahfang:
### Did you check the docs?

- [X] I have read all the docs.

### Is your feature request related to a problem? Please describe.

The msg before new filename is too long:

<img width="892" alt="image" src="https://github.com/nvim-neo-tree/neo-tree.nvim/assets/16301509/5b8ef5fc-ed86-4b74-9ebc-7d23d1c149cb">

### Describe the solution you'd like.

Provide option to setup short msg content.

pysan3:
What do you think would be a good message? @jonahfang

jonahfang:
I like the NERDTree ui:

<img width="620" alt="image" src="https://github.com/nvim-neo-tree/neo-tree.nvim/assets/16301509/fff5b88f-60a9-44e1-b201-9e4a24c91958">

PR #1189, opened 2023-10-18
feat(input): add newline when `use_popups_for_input = false`

pysan3:
Closes: #1172

A very simple PR; inserts a `'\n'` between the popup message and default value instead of a space.

### DIFF
--- a/lua/neo-tree/ui/inputs.lua
+++ b/lua/neo-tree/ui/inputs.lua
@@ -11,7 +11,7 @@ local should_use_popup_input = function()
 end
 
 M.show_input = function(input, callback)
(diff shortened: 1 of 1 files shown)


cseickel (approved):

jonahfang:
When I set `opt.cmdheight=0`, the prompt disappears!

pysan3:
@jonahfang At least in my environment I do not have any problems with `cmdheight=0`.

Could you elaborate your situation and create a new issue?

I would like to hear your neovim version, OS and a screenshot would be nice as well.

Please mention me (@pysan3) in the issue as well and I'll look into it whenever possible.

PR #1190, opened 2023-10-19
feat!: rename `label` to `desc` for mapping descriptions

mehalter:
This aligns the mapping API closer with the core `keymap.set` lua API in Neovim

This is a continuation on #1186

### DIFF
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -412,7 +412,7 @@ For example:
              local node = state.tree:get_node()
              print(node.path)
            end,
-           label = "print path",
+           desc = "print path",
          },
        }
      },
@@ -428,7 +428,7 @@ For example:
 The above config will map `A` to command_a for all sources except for
 filesystem, which will use command_b instead.
 
-The label is used to display in the help popup.
+The desc is used to display in the help popup.
 
 If you don't want to use *any* default mappings, you can set
 `use_default_mappings = false` in your config.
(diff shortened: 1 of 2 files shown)


cseickel (approved):

NGPONG:
Sorry for disturb you all, but when this PR is merged, is there any other way we can set the `desc` attribute of a single key?

coffebar:
> Sorry for disturb you all, but when this PR is merged, is there any other way we can set the `desc` attribute of a single key?

please explain the purpose of what you want to do, why do you need another way

NGPONG:
Hi @coffebar, Thank you for your response. In my workflow I often rely on `which-key` to do the key prompting for me, and the `which-key` plugin in turn relies on the `desc` property of `key` (specified by the `opts` parameter of vim.keymap.set)

Prior to this pr merge, I would have configured the above workflow in this way

```lua
require('neo-tree').setup({
    window = {
      mapping_options = { 
          silent = true,
          noremap = true,
          nowait = true,
      },
      mappings = {
          ['a'] = { command = 'noop', desc = 'which_key_ignore' },
          ['fi'] = { command = 'foo', desc = 'Show file details.' },
      },
    }
})
```
But after this pr merge, the `desc` attribute that was previously set won't work, because it will be filtered out here

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/dcef81d96e0678b01cca5445cffb9a3ae0e9ce4f/lua/neo-tree/ui/renderer.lua#L827

This caused me to rework the previous init.lua framework to accommodate the change.

mehalter:
That should be considered a bug on my part. Sorry! That should not be getting filtered out 😓

mehalter:
I opeened #1206 to resolve this. It also does a step further and makes sure `desc` is set when using functions based on string.

Issue #128, opened 2022-02-06
Restore tree state in session restore


cseickel:
### Discussed in https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/126

<div type='discussions-op-text'>

<sup>Originally posted by **bennypowers** February  6, 2022</sup>
Hello!

I'm using neo-tree with `auto-session` and `bufferline.nvim`, and notice the following:

- WHEN neo-tree is open and I quit vim with buffers open in session via `quitall`
- THEN when I reopen the session from the shell prompt, 
  - The left pane (where the tree should be) is blank
  - A new, empty buffer appears in the buffer line labeled `neo-tree filesystem [1]`
  - the bufferline, normally offset via bufferline config, is flush to the left of the screen

<img width="1014" alt="Screen Shot 2022-02-06 at 11 06 34" src="https://user-images.githubusercontent.com/1466420/152674368-7d6b3567-66de-4e05-969a-f08da25e8ff4.png">
<img width="1014" alt="Screen Shot 2022-02-06 at 11 06 40" src="https://user-images.githubusercontent.com/1466420/152674370-ab26ee75-0594-49ed-b4c6-855a336536af.png">

I'd like instead for the session to restore as is, i.e.

- WHEN neo-tree is open and I quit vim with buffers open in session via `quitall`
- THEN when I reopen the session from the shell prompt, 
  - The left pane (where the tree should be) contains a neo-tree with the same state as at quit
  - the bufferline offsets from the tree as it did before quit
  - the state of the bufferline persists

Should I hook into auto-session to prevent saving buffers with type `neo-tree` and to focus the active buffer on startup? </div>

cseickel:
@bennypowers I am trying out `auto-session` and I really like it. Unfortunately I can get this or my old session manager to restore the tree windows at all, not even as empty buffers. It is just completely committed from the session restore. 

Is it just me or does that happen to you as well with the latest updates?

aiecee:
The way I got around this with `auto-session` was to use the command hooks.
`pre_save_cmds` to do `tabdo NeoTreeClose`
`post_save_cmds` to do `tabdo NeoTreeReveal`
`post_restore_cmds` to do `tabdo NeoTreeReveal`

bennypowers:
This is my current config:

```lua
local function close_neo_tree()
  require 'neo-tree.sources.manager'.close_all()
end

local function open_neo_tree()
[...]

PR #1191, opened 2023-10-21
Update state.explicitly_opened_directories when collapsing by action close_all

coffebar:
This purpose of this change is to keep `state.explicitly_opened_directories` up to date not only when you open Nodes, but also when you close them. 
Specifically, this applies to actions `close_all_nodes` and `close_all_subnodes`.

Not sure if this change is fixing some bugs in the neo-tree itself. The reason why I need this: one of my plugins uses `state.explicitly_opened_directories` to save and restore Neo-tree when switching projects, and in above cases, 'explicitly_opened_directories' variable does not correspond to the real state. This change will fix [the bug.](https://github.com/coffebar/neovim-project/issues/12)

### DIFF
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -153,6 +153,12 @@ M.close_node = function(state, callback)
     target_node:collapse()
     renderer.redraw(state)
     renderer.focus_node(state, target_node:get_id())
+    if
+      state.explicitly_opened_directories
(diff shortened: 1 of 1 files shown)


cseickel (approved):

cseickel:
These are good fixes, but I should point out that using `explicitly_opened_directories` to restore a session is not the best way to go about that. See this comment for a good example:

https://github.com/nvim-neo-tree/neo-tree.nvim/issues/128#issuecomment-1742645902

PR #1203, opened 2023-10-26
feat(preview): use image.nvim to load images into preview window

### LINKED ISSUE #1196: FEATURE: Image Preview inside Preview Window
### Did you check the docs?

### Is your feature request related to a problem? Please describe.

We had some discussions here: https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/831#discussioncomment-7291028 and thought of introducing image preview into neo-tree.

As I stated here: https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/831#discussioncomment-7352011, the author of [image.nvim](https://github.com/3rd/image.nvim) added a feature to preview images inside a buffer using the kitty graphics protocol.

This is a good building block for image preview inside the preview window, and I thought of start working on it.

I already got a proof-of-concept working and will be submitting a PR once I'm able to test it thoroughly.

https://user-images.githubusercontent.com/41065736/277180763-b7152637-f310-43a5-b8c3-4bcba135629d.mp4

I made this issue to log my progress and maybe guide people who googled for image preview support in neovim here.

Any suggestion or advice is also appreciated.

### Describe the solution you'd like.

I'll be working on this as soon as I got some spare time.

### Describe alternatives you've considered.

_No response_

### Additional Context

_No response_

pysan3:
Closes #1196

- [x] Added docs in README and help page as well.
- [x] Checking `vim.bo[bufnr].filetype == "image_nvim"` is enough to decide if we need 3rd/image.nvim.

Tagging some people that might be relevant. (Please ignore if you are not interested)
@SlimShadyIAm @cseickel

### DIFF
--- a/lua/neo-tree/sources/common/preview.lua
+++ b/lua/neo-tree/sources/common/preview.lua
@@ -272,7 +272,7 @@ end
 
 ---@param winid number
 ---@param bufnr number
-local function load_image_nvim_buf(winid, bufnr)
+local function try_load_image_nvim_buf(winid, bufnr)
   if vim.bo[bufnr].filetype ~= "image_nvim" then
     return false
   end
@@ -292,7 +292,7 @@ function Preview:setBuffer(bufnr)
   vim.opt.eventignore:append("BufEnter,BufWinEnter")
   vim.api.nvim_win_set_buf(self.winid, bufnr)
   if self.config.use_image_nvim then
-    load_image_nvim_buf(self.winid, bufnr)
+    try_load_image_nvim_buf(self.winid, bufnr)
   end
   if self.config.use_float then
     -- I'm not sufe why float windows won;t show numbers without this
--- a/lua/neo-tree/sources/common/preview.lua
+++ b/lua/neo-tree/sources/common/preview.lua
@@ -270,12 +270,30 @@ function Preview:activate()
   vim.api.nvim_win_set_var(self.winid, "neo_tree_preview", 1)
 end
 
+---@param winid number
+---@param bufnr number
+local function load_image_nvim_buf(winid, bufnr)
+  if vim.bo[bufnr].filetype ~= "image_nvim" then
+    return false
+  end
+  local success, mod = pcall(require, "image")
+  if not success or not mod.hijack_buffer then
+    local image_nvim_url = "https://github.com/3rd/image.nvim"
+    log.debug("You'll need to install image.nvim to use this command: " .. image_nvim_url)
+    return false
+  end
+  return mod.hijack_buffer(vim.api.nvim_buf_get_name(bufnr), winid, bufnr)
+end
+
 ---Set the buffer in the preview window without executing BufEnter or BufWinEnter autocommands.
 --@param bufnr number The buffer number of the buffer to set.
 function Preview:setBuffer(bufnr)
   local eventignore = vim.opt.eventignore
   vim.opt.eventignore:append("BufEnter,BufWinEnter")
   vim.api.nvim_win_set_buf(self.winid, bufnr)
+  if self.config.use_image_nvim then
+    load_image_nvim_buf(self.winid, bufnr)
+  end
   if self.config.use_float then
     -- I'm not sufe why float windows won;t show numbers without this
     vim.api.nvim_win_set_option(self.winid, "number", true)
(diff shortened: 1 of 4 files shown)


miversen33:
I don't know how I feel about enabling this by default. There is a line that logs out (at debug level) that you need `image.nvim` to preview images ([here](https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1203/files#diff-c4d524498332fcd546caf1d6a1460113a6ddd2db2acde65b664f6c13a493cc25R282)). I don't remember, does the `log` module print as well or just dump to the log file?

If there is no pain for an end user that doesn't care about this (or actively doesn't want it), I don't see an issue with it. My quick glance just saw that log line and makes me worry that users may get told they need something that they may not actually want (or can't use, if you're a windows user for example).

cseickel:
> does the `log` module print as well or just dump to the log file?

No, only INFO, WARN, and ERROR are shown as messages. DEBUG and TRACE only go to the log file.

lua/neo-tree/sources/common/preview.lua
```diff
@@ -270,12 +270,30 @@ function Preview:activate()
   vim.api.nvim_win_set_var(self.winid, "neo_tree_preview", 1)
 end
 
+---@param winid number
+---@param bufnr number
+local function load_image_nvim_buf(winid, bufnr)
```
cseickel, line 275: I think this should be called `try_load_image_nvim_buf` or `load_image_nvim_buf_if_image` or something similar.

cseickel (commented):
Just a few comments.

pysan3:
> There is a line that logs out (at debug level) that you need image.nvim to preview images

Thanks for the review @miversen33 . Actually, if you don't have image.nvim installed, you will never have `vim.bo[bufnr].filetype` set to `image_nvim` (the `if` statement before checking `pcall`) so you don't have to worry about that (unless you manually set it for some obscure reason). 

@cseickel I will change it tomorrow. Thanks.

wookayin:
I'm also not sure if this can be a default behavior -- should be an opt-in feature upon configuration. Out of curiosity, what happens if one uses non-compliant terminals (e.g. iTerm or wezterm) or tmux?

pysan3:
Thanks for ur comment @wookayin.

As I said in above comments, if the plugin is not installed, buffer filetype will never be set to `image_nvim` which is what is checked first (`vim.bo[bufnr].filetype == "image_nvim"`), even before `pcall(require, "image")`.

> Out of curiosity, what happens if one uses non-compliant terminals (e.g. iTerm or wezterm) or tmux?

Wezterm does support kitty graphics protocol so you can use it OOB iirc (dunno about iTerm sorry). Tmux works just fine if the underlying terminal works (for more info plz read the [README](https://github.com/3rd/image.nvim)). For other terminals (such as `st` or windows terminal), there's no reason to install `image.nvim`. 

If you do install image.nvim on a terminal which does not support kitty graphics protocol, it should delete the content (binary junk) of the buffer and silently fail to display the image so you'll get an empty buffer.

I am more towards opt-out (since it is just one extra `if-check` when `image.nvim` is not installed) but I would like to ask @cseickel for the final decision as he also mentioned to make the default to `true` here: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1196#issuecomment-1775584636

lua/neo-tree/sources/common/preview.lua
pysan3: Changed to `try_load_image_nvim_buf`

cseickel:
> I am more towards opt-out (since it is just one extra `if-check` when `image.nvim` is not installed) but I would like to ask @cseickel for the final decision as he also mentioned to make the default to `true` here: [#1196 (comment)](https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1196#issuecomment-1775584636)

On the one hand, if the user has already installed image.nvim, that is a very intentional action and they wouldn't do so if their terminal didn't support it. If they haven;t taken they extra step of installing image.nvim, nothing should happen.

On the other hand, it's always a safe bet to add new features as opt-in at first, and then later on change the default to opt-out when it's had more testing.

Safe is always better, let's go ahead and default `use_image_nvim` to false at first.

pysan3:
> On the other hand, it's always a safe bet to add new features as opt-in at first, and then later on change the default to opt-out when it's had more testing.

I agree on that. OK, fixed to `false`.

Dang it, mistake on the commit message 🤦...

cseickel (approved):

PR #1206, opened 2023-10-28
Fix `desc` being filtered out of mappings

mehalter:
This also adds the `desc` when using a function by name

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -824,7 +824,7 @@ local set_buffer_mappings = function(state)
         local desc
         if type(func) == "table" then
           for key, value in pairs(func) do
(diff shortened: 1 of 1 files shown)


lua/neo-tree/ui/renderer.lua
```diff
@@ -834,6 +834,7 @@ local set_buffer_mappings = function(state)
         end
         if type(func) == "string" then
           resolved_mappings[cmd] = { text = func }
+          map_options.desc = func
```
cseickel, line 837: I think this should preserve the provided `desc` in case it was set. Someone might want to have a description that's different from the function name.
```suggestion
          map_options.desc = map_options.desc or func
```

cseickel (approved):

PR #1213, opened 2023-10-31
fix: set max height for popup help window

intxff:
When the list of keymaps is too long, the borders of the popup help window are not displayed properly.

Before
before

After
after


cseickel (approved):
Looks good, thanks!

Issue #1165, opened 2023-09-26
FEATURE: Add some additional movement commands


kmoschcau:
### Did you check the docs?

- [X] I have read all the docs.

### Is your feature request related to a problem? Please describe.

In some constellations of opened child nodes it is somewhat hard to get to the nodes I want to quickly. For example I'd like to go to the first or last child of an opened node or skip an opened node entirely.

### Describe the solution you'd like.

Add commands for the following:
- next/previous sibling (skips child nodes)
- first/last sibling (jumps to the first or last sibling on the same nesting level)

dannywrayuk:
I'd also like to see this. Though, I think it might be possible using a custom command. I'll reply if I figure it out.

miki725:
Another one I found missing from nvim-tree was to go to the parent node. I have something like:

```lua
parent = function(state)
    local node = state.tree:get_node()
    local parent_path, _ = utils.split_path(node:get_id())
    renderer.focus_node(state, parent_path)
end,
```

dannywrayuk:
I've been using this: 

```lua
local goToParent = function(state)
    local node = state.tree:get_node()
    renderer.focus_node(state, node:get_parent_id())
end

local goToParentAndClose = function(state)
    goToParent(state)
    local node = state.tree:get_node()
    if node:is_expanded() then
        fsc.toggle_node(state)
    end
end
```

pysan3:
I think your snippets are worth a PR ;)

If you have any interests, take a look at https://github.com/nvim-neo-tree/neo-tree.nvim/blob/main/lua/neo-tree/sources/common/commands.lua and https://github.com/nvim-neo-tree/neo-tree.nvim/blob/main/lua/neo-tree/sources/filesystem/commands.lua.

kmoschcau:
Since we already have this discussion going and I didn't want to create a new issue just for this: Is there a way to get the open commands to NOT automatically expand nested files and instead just open the top level file?

pysan3:
> instead just open the top level file?

Do you mean folder?

kmoschcau:
I mean to have one command that does the following:

1. when run on a closed folder/directory -> expands it
2. when run on a file with other files nested under it -> opens the file without expanding it
3. when run on any other file -> opens the file

Number 2 is not possible right now as far as I know, because the built-in open command will always first expand files with nesting and then open it when you run it again on the same file.

pysan3:
Ok, got it.

[...]

PR #1218, opened 2023-11-07
feat(commands): add option to open top file even when nested

pysan3:
This feature was asked here: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1165#issuecomment-1797935289

Basically to add a keybind to open the top level file of nested files without expanding it.

Could you review this as well? @@Teddytrombone 

---

I mean to have one command that does the following:

1. when run on a closed folder/directory -> expands it
2. when run on a file with other files nested under it -> opens the file without expanding it
3. when run on any other file -> opens the file

Number 2 is not possible right now as far as I know, because the built-in open command will always first expand files with nesting and then open it when you run it again on the same file.

_Originally posted by @kmoschcau in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1165#issuecomment-1797935289_

### DIFF
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -365,6 +365,7 @@ local config = {
       },
       ["<2-LeftMouse>"] = "open",
       ["<cr>"] = "open",
+      -- ["<cr>"] = { "open", config = { no_expand_file = true } }, -- open top file instead of expanding when nested
       ["<esc>"] = "cancel", -- close preview or floating neo-tree window
       ["P"] = { "toggle_preview", config = { use_float = true, use_image_nvim = false } },
       ["l"] = "focus_preview",
--- a/lua/neo-tree/sources/common/commands.lua
+++ b/lua/neo-tree/sources/common/commands.lua
@@ -713,7 +713,8 @@ local open_with_cmd = function(state, open_cmd, toggle_directory, open_file)
     if toggle_directory and node.type == "directory" then
       toggle_directory(node)
     elseif node:has_children() then
-      if node:is_expanded() and node.type ~= "directory" then
+      local no_expand_file = (state.config or {}).no_expand_file
+      if node.type ~= "directory" and (no_expand_file or node:is_expanded()) then
         return open()
       end
 


kmoschcau:
Nice turnaround! 😀
I'm not a maintainer, but this seems good to me. If anything it could use some more documentation than just one comment in the example config, but other than that it's good.

cseickel (approved):

pysan3:
Thanks for your feedback @kmoschcau 

Unfortunately I haven't been able to follow closely with the development of Neotree recently but is there a place where this file nesting is documented? That would be a good place to add it. I'd be glad if you could add it somewhere @kmoschcau   

Or we could go to the extreme and actually add a keybind for open with no expand file as default. (Maybe <c-cr> or <m-cr>?) or even make it the default behavior?? how'd you think? @cseickel

cseickel:
I don't think it's necessary to add a new keybind for a variation of the open command @pysan3, I really expect that most people should be customizing their mappings anyhow and this is a pretty minor thing to change.

It probably should be the default behavior though, and I think it used to be. It probably just got broken along the way. I don't use nesting so I might not have approved the change and not even noticed that a bug was added.

pysan3:
> It probably should be the default behavior though, and I think it used to be.

Oh, I see. I'll make a new PR then. 

But my fix is not convenient and if you make it the default, there'll be no way to expand the folded files by default. Is that ok?

Also, There are split open, vsplit open etc, and you mean that all should be changed right?

cseickel:
> But my fix is not convenient and if you make it the default, there'll be no way to expand the folded files by default. Is that ok?

I think that this is the exact reason why we have a separate action for expand, so that you can expand/close nested files without opening them.

> Also, There are split open, vsplit open etc, and you mean that all should be changed right?

All of the commands that use `open_with_cmd` already naturally support the option so it should be documented as such.

I do think that the default should be the `no_expand_file=true` you just added, but since it is the default you should NOT have to specify the option. I would actually rename the option to `expand_nested_files` which defaults to `nil`/`false` which is the opposite meaning of `no_expand_file`. For backwards compatibility (because I already released it 🤦🏼) we should support both options but only mention the new `expand_nested_files` option in the docs.

Teddytrombone:
> Could you review this as well? @@Teddytrombone

Sorry, I was quite busy the last days. At a first glance it looks good to me :-D

pysan3:
Thanks mate ;)

My bad It was moved to https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1220 and it should be good to go.

PR #1219, opened 2023-11-08
fix(commands)!: open nested files default to open instead of expand

pysan3:
- use `{ "open", config = { expand_nested_files = true } }` for old behavior
- `no_expand_file` options is deprecated, move to `expand_nested_files` (OPPOSITE)

### DIFF
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -365,7 +365,7 @@ local config = {
       },
       ["<2-LeftMouse>"] = "open",
       ["<cr>"] = "open",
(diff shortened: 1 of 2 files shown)


pysan3:
Wait is this really related to this PR?

I really cannot find the cause.

Would you mind rerunning the test action @cseickel ?

cseickel:
I reran it twice and it has the same result @pysan3 .

pysan3:
> I reran it twice and it has the same result @pysan3 .

Thanks, ok so it should be on me.

I really can't stop anything wrong now so let me come back to this later on.

Issue #1214, opened 2023-11-03
FEATURE: auto select new added file (created by copy)


jonahfang:
### Did you check the docs?

- [X] I have read all the docs.

### Is your feature request related to a problem? Please describe.

not related to any problem.

### Describe the solution you'd like.

I am adding new files using `use_popups_for_input=false` option and I want to automatically select newly added files after adding or copying. When I do a copy operation, I always mistake the original file for the new file because the currently selected file has not changed. Newly created files can be edited immediately if they are automatically selected.

pysan3:
Yes, there is clearly a bug in the codebase due to some racing conditions of debounces.

At least the plugin is trying to do this by default. It's just not working lol.

I'll try to look into it ;)

pysan3:
Maybe I found the issue.

@jonahfang 
Would you check if this gets fixed when you set `hijack_netrw_behavior = "disabled"`?

jonahfang:
> Maybe I found the issue.
> 
> @jonahfang Would you check if this gets fixed when you set `hijack_netrw_behavior = "disabled"`?

I try it, and when I copy to create a new file, the cursor still remain at the original file.

pysan3:
Thanks, OK first guess wrong...

I'm now starting to think it's caused by `use_libuv_file_watcher` but not sure.
I'll try to get back to you when I figured out the reason

pysan3:
Now, the race condition is still there, but the cause is that `utils.debounce_strategy.CALL_FIRST_AND_LAST == 0` meaning that we were waiting 0ms before.

I'm not sure if extending that to 100ms will solve this issue for all devices in the world, but at least it is working on my 6 yo laptop (kekw) so I guess that margin is pretty enough?

jonahfang:
Is there anything I can do to help?

pysan3:
Thanks, that would help a lot!

Would you install the fork and see if it fixes the problem?

If you are using lazy, it will be something like 

```lua
return {
  "pysan3/neo-tree.nvim",
  branch = "debounce-correctly",
  -- ...
}
```

jonahfang:
I  installed the fork and it does not fixe the problem!

pysan3:
Dang it...

Would you mind sharing you config?

Also could you send me the trace log?

```lua
{
  "pysan3/neo-tree.nvim",
...
  opts = {
    log_level = "trace",
    log_to_file = true,
  },
}
```

[...]

PR #1227, opened 2023-11-12
fix(filesystem): fixed call to debounce with wrong arguments

pysan3:
closes: #1214

Please read [my comments](https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1214#issuecomment-1807104855) for the explanation.

But ultimately, this is a failure of an untyped scripting language lol.

We definitely need to introduce type notations at some point. I'm too sick of these kind of errors.

### DIFF
--- a/lua/neo-tree/utils/init.lua
+++ b/lua/neo-tree/utils/init.lua
@@ -35,20 +35,25 @@ local diag_severity_to_string = function(severity)
 end
 
 local tracked_functions = {}
+---@enum NeotreeDebounceStrategy
 M.debounce_strategy = {
   CALL_FIRST_AND_LAST = 0,
   CALL_LAST_ONLY = 1,
 }
 
+---@enum NeotreeDebounceAction
 M.debounce_action = {
   START_NORMAL = 0,
   START_ASYNC_JOB = 1,
   COMPLETE_ASYNC_JOB = 2,
 }
 
-local defer_function
--- Part of debounce. Moved out of the function to eliminate memory leaks.
-defer_function = function(id, frequency_in_ms, strategy, action)
+---Part of debounce. Moved out of the function to eliminate memory leaks.
+---@param id string Identifier for the debounce group, such as the function name.
+---@param frequency_in_ms number Miniumum amount of time between invocations of fn.
+---@param strategy NeotreeDebounceStrategy The debounce_strategy to use, determines which calls to fn are not dropped.
+---@param action NeotreeDebounceAction? The debounce_action to use, determines how the function is invoked
+local function defer_function(id, frequency_in_ms, strategy, action)
   tracked_functions[id].in_debounce_period = true
   vim.defer_fn(function()
     local current_data = tracked_functions[id]
@@ -70,9 +75,10 @@ end
 
 ---Call fn, but not more than once every x milliseconds.
 ---@param id string Identifier for the debounce group, such as the function name.
----@param fn function Function to be executed.
+---@param fn function? Function to be executed.
 ---@param frequency_in_ms number Miniumum amount of time between invocations of fn.
----@param strategy number The debounce_strategy to use, determines which calls to fn are not dropped.
+---@param strategy NeotreeDebounceStrategy The debounce_strategy to use, determines which calls to fn are not dropped.
+---@param action NeotreeDebounceAction? The debounce_action to use, determines how the function is invoked
 M.debounce = function(id, fn, frequency_in_ms, strategy, action)
   local fn_data = tracked_functions[id]
 
@@ -119,8 +125,8 @@ M.debounce = function(id, fn, frequency_in_ms, strategy, action)
   if type(fn) == "function" then
     success, result = pcall(fn)
   end
-  fn_data.fn = nil
   fn = nil
+  fn_data.fn = fn
 
   if not success then
     log.error("debounce ", id, " error: ", result)
@@ -208,7 +214,7 @@ end
 ---Converts a filesize from libuv.stats into a human readable string with appropriate units.
 ---@param size any
 ---@return string
-M.human_size = function (size)
+M.human_size = function(size)
   local human = filesize(size, { output = "string" })
   ---@cast human string
   return human
@@ -355,12 +361,12 @@ M.get_inner_win_width = function(winid)
 end
 
 local stat_providers = {
-  default = function (node)
+  default = function(node)
     return vim.loop.fs_stat(node.path)
   end,
 }
 
---- Gets the statics for a node in the file system. The `stat` object will be cached 
+--- Gets the statics for a node in the file system. The `stat` object will be cached
 --- for the lifetime of the node.
 ---
 ---@param node table The Nui TreeNode node to get the stats for.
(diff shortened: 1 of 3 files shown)


cseickel:
Actually, there are type annotations already:

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/0f6e7acfd86b052acf78baccba04d5c61dcbbc0d/lua/neo-tree/utils/init.lua#L71-L76

The problem here is that both of the args are actually integers. We could add a validation on the value passed in for strategy and make sure that it does not overlap with valid values for frequency.

pysan3:
There's a type specifically dedicated to this kind of problem. https://github.com/LuaLS/lua-language-server/wiki/Annotations#enum

I was kinda talking about type notation for the whole plugin. It is very easy to loose track of `state` as well as I hear a lot of questions of how to use `events` in the config.

We would want to define a `NeotreeState` [class](https://github.com/LuaLS/lua-language-server/wiki/Annotations#class) and also one for config would be nice.
Just like this: https://github.com/3rd/image.nvim/blob/master/lua/types.lua.

cseickel:
> There's a type specifically dedicated to this kind of problem. https://github.com/LuaLS/lua-language-server/wiki/Annotations#enum

I didn't realize that was an option, can you apply that in this PR?

> Running a type check test for all PRs will solve most of the problems.
Like https://github.com/mrcjkb/lua-typecheck-action should do the job (haven't used myself tho).

Hell yeah! I'm adding it.


> I was kinda talking about type notation for the whole plugin. It is very easy to loose track of `state` as well as I hear a lot of questions of how to use `events` in the config.
> 
> We would want to define a `NeotreeState` [class](https://github.com/LuaLS/lua-language-server/wiki/Annotations#class) and also one for config would be nice. Just like this: https://github.com/3rd/image.nvim/blob/master/lua/types.lua.

That is an amazing idea, it would be great to have. I don't quite have that quantity of time any more but I would happily merge it if you (or anyone else) could do the work.

pysan3:
I'll change it to the enum notation but I'd like to make a naming convention for the types we are going to be defining.

How about `NeotreePascalCase` for types that users might use outside (defined in `./neo-tree/types.lua` (Eg `NeotreeState`, `NeotreeConfig`, ...) and `neotree_snake_case` for types used inside? Eg, this enum will be named like `---@enum neotree_debounce_strategy`.

Unfortunately I'm quite busy now but I'll be able to work on it on February next year.

I don't think it's a good idea to add the action now since it will raise tons of errors all over the place.

If you are ok with waiting until next year, I can do the work then. It would be nice if we could do an entire feature freeze while I'm working on that since I think I'll need to change quite a lot of code.

On top of that, if that's ever possible I'm also thinking of reinventing a path module so that it works just right on all OS. As well as git on windows seems to be broken right now when there's a non-ascii char in the path.

pysan3:
Here's a poc for enum. As you can see, diagnose is working correctly here ;)

image

cseickel:
I think the Neotree prefix and using pascal case both sound good. I don't think it's worthwhile to have a different convention for internally used types though, I would just use Pascal case for all types.

You are right about the type checker. I opened a PR but there are a lot of errors and warnings to deal with so I don't know if I will be able to follow through or not.

I totally understand your time constraints, and if you are still willing and able to work on it next year then that will be great. I would not mind a feature freeze while you do that because I think that this work would be way more important than any new features.

regarding a path module, you're right about that. It might be a good idea to check on Plenary's Path module and see if it handles these issues or not.

pysan3:
> I think the Neotree prefix and using pascal case both sound good. I don't think it's worthwhile to have a different convention for internally used types though, I would just use Pascal case for all types.

Alright, will go with that.

> You are right about the type checker. I opened a PR but there are a lot of errors and warnings to deal with so I don't know if I will be able to follow through or not.

I know how to fix the type issues so I can handle it if I've got time ;) (We have the final option `---@diagnostic disable-line` up our sleeves lol)

> I totally understand your time constraints, and if you are still willing and able to work on it next year then that will be great. I would not mind a feature freeze while you do that because I think that this work would be way more important than any new features.
> 

Thanks, I'll ping you when I can start working on it.

> regarding a path module, you're right about that. It might be a good idea to check on Plenary's Path module and see if it handles these issues or not.

Yah, I'm aware of that thing but I don't think it's meant for a heavy usage like in our case imo. It's lacking features and we would eventually have to implement many things with barebone luv anyways.

pysan3:
@cseickel Do you know what you were trying to do here?
image

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/0f6e7acfd86b052acf78baccba04d5c61dcbbc0d/lua/neo-tree/git/status.lua#L240

pysan3:
> > I think the Neotree prefix and using pascal case both sound good. I don't think it's worthwhile to have a different convention for internally used types though, I would just use Pascal case for all types.
> 
> Alright, will go with that.

Well, we can distinguish between what would be considered as a _breaking change_ if we have some kind of difference between private types and public types.

Using private types from outside will be definitely valid but we can break it at any time, but for the pascal case ones, they should be announced as a breaking change or something like that.

cseickel:
> @cseickel Do you know what you were trying to do here? image

That's just nonsense that slipped by me in #637. You can delete that line as its not doing anything.

pysan3:
> That's just nonsense that slipped by me in #637. You can delete that line as its not doing anything.

Done.

I guess it's ready to go (tho the tests are failing...)

BTW may I ask your opinion on this? https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1227#issuecomment-1807194997

cseickel:
> Well, we can distinguish between what would be considered as a _breaking change_ if we have some kind of difference between private types and public types.
> 
> Using private types from outside will be definitely valid but we can break it at any time, but for the pascal case ones, they should be announced as a breaking change or something like that.

I just don't see types as something that are easily defined as internal or external. Methods and modules are what is truly internal/external. Besides, I'm not sure we actually have any types that would qualify as internal only:

https://github.com/nvim-neo-tree/neo-tree.nvim#what-is-a-breaking-change

lua/neo-tree/utils/init.lua
```diff
@@ -70,9 +75,10 @@ end
 
 ---Call fn, but not more than once every x milliseconds.
 ---@param id string Identifier for the debounce group, such as the function name.
----@param fn function Function to be executed.
+---@param fn function? Function to be executed.
```
cseickel, line 78: This is actually required.

pysan3: Sorry, fixed.

pysan3:
> > Well, we can distinguish between what would be considered as a _breaking change_ if we have some kind of difference between private types and public types.
> > Using private types from outside will be definitely valid but we can break it at any time, but for the pascal case ones, they should be announced as a breaking change or something like that.
> 
> I just don't see types as something that are easily defined as internal or external. Methods and modules are what is truly internal/external. Besides, I'm not sure we actually have any types that would qualify as internal only:
> 
> [nvim-neo-tree/neo-tree.nvim#what-is-a-breaking-change](https://github.com/nvim-neo-tree/neo-tree.nvim#what-is-a-breaking-change)

I see. I'll go all in with PascalCase :+1:

cseickel (approved):

pysan3:
Not gonna start yet but pasting this as a reminder. 

https://github.com/mrcjkb/lua-typecheck-action/issues/34

PR #1228, opened 2023-11-12
feat(filesystem): move to new dst after copy / move in filesystem

pysan3:
Requested [here](https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1214#issuecomment-1807130056), this PR changes the cursor position to the newly created file after `copy` command instead of focusing the src file.

https://github.com/nvim-neo-tree/neo-tree.nvim/assets/41065736/47daf22d-2e66-45e4-bce5-ccf5464b32cf

### DIFF
--- a/lua/neo-tree/sources/filesystem/commands.lua
+++ b/lua/neo-tree/sources/filesystem/commands.lua
@@ -29,7 +29,7 @@ M.clear_filter = function(state)
 end
 
 M.copy = function(state)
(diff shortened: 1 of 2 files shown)


cseickel:
This may seem nit-picky, but could you rename all instances of `dst` to `destination` or `dest`? I just have a hard time reading `dst` as destination.

cseickel (approved):

PR #1239, opened 2023-11-19
feat(popups): toggle title

sand4rt:
To disable all popup titles, catering to those seeking a minimalist setup. Users are likely acquainted with the popups and may find titles unnecessary.


lua/neo-tree/defaults.lua
```diff
@@ -13,6 +13,7 @@ local config = {
   auto_clean_after_session_restore = false, -- Automatically clean up broken neo-tree buffers saved in sessions
   close_if_last_window = false, -- Close Neo-tree if it is the last window left in the tab
   -- popup_border_style is for input and confirmation dialogs.
+  -- popup_title_enabled = false to disable the title of all popups
```
cseickel, line 16: This should just be set to `true` and not commented out. You can leave an additional comment to describe what it does.

PR #1252, opened 2023-12-02
fix: use getftype to check symlink

yehuohan:
This PR is simple fix to get correct 'link' file type for windows. Fot that it's seemed `fs_stat` can't work properly for windows' symlink.

### DIFF
--- a/lua/neo-tree/sources/common/file-items.lua
+++ b/lua/neo-tree/sources/common/file-items.lua
@@ -117,7 +117,7 @@ function create_item(context, path, _type, bufnr)
     path = path,
     type = _type,
   }
-  if item.type == "link" then
+  if vim.fn.getftype(path) == "link" then
     item.is_link = true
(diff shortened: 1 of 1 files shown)


lua/neo-tree/sources/common/file-items.lua
```diff
@@ -117,7 +117,7 @@ function create_item(context, path, _type, bufnr)
     path = path,
     type = _type,
   }
-  if item.type == "link" then
+  if vim.fn.getftype(path) == "link" then
```
cseickel, line 120: I worry that this will have some impact on performance on non-windows computers if we have this extra function call for every file. Can you change it to only do this extra check if the OS is windows?

```suggestion
  if utils.is_windows then
    if vim.fn.getftype(path) == "link" then
      item.type = "link"
    end
  end
  if item.type == "link" then
```

yehuohan: Commited. Thank your review.

PR #1258, opened 2023-12-10
fix(filesystem): Mixed path separators in Windows

### LINKED ISSUE #1256: BUG: Path normalization inconsistent with NeoVim's, breaking forward slashes in Windows

### Neovim Version (nvim -v)

0.9.4

Windows 11 22631.2715

### Describe the Bug

Note: I'm using PowerShell 7.4, AKA "PowerShell Core", the cross platform PowerShell. I'm new to Windows, so bare with me if I missed a detail.

When using `:Neotree reveal` with a C# file (Omnisharp LSP attached), I get the "File not in cwd" prompt. But the y/n offer to change the cwd looked odd to me.

- Actual file: `C:\Users\example\src\project\foo.cs`
- Actual cwd: `C:\Users\example\src\project`
- Proposed "change cwd to" in prompt: `C:\Users` <-- huh?

So I did some digging. In [utils.normalize_path](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/9298c1b7a52e63ed4fd977fd792a0d4afe9260d5/lua/neo-tree/utils/init.lua#L760), which is used by [handle_reveal](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/9298c1b7a52e63ed4fd977fd792a0d4afe9260d5/lua/neo-tree/command/init.lua#L208), it doesn't take into account path separators. Inspecting the values of the CWD and buffer(name|path), it's strange:

- Actual value of expanding `%:p`: `C:\Users/example/src/project/foo.cs`

It honestly took me a while to even notice because I'm not used to dealing with Windowsisms, but the path is mixing forward and reverse slashes. This actually seems to be valid on Windows.

So what's happening is that [utils.is_subpath](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/9298c1b7a52e63ed4fd977fd792a0d4afe9260d5/lua/neo-tree/utils/init.lua#L780) always ends up returning false because of the mismatched path separators. Then, to create the proposed CWD for the prompt, [utils.split_path](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/9298c1b7a52e63ed4fd977fd792a0d4afe9260d5/lua/neo-tree/utils/init.lua#L890) splits on the backslash on Windows, which is how we end up with the strange suggestion in the prompt.

I looked into how NeoVim normalizes paths internally. Long story short, `vim.fs.normalize(path)` results in *only* forward slashes.

I did a quick hard-coding of `utils.normalize_path` to return `vim.fs.normalize(path)` and it seemed to have solved my particular issue with `:Neotree reveal`, but I didn't do any thorough testing, so I haven't made a PR. But upon further inspection, it seems like this *class* of bug might be in multiple places. I'm honestly pretty bad at Lua and don't know this code base, so I figured this detailed bug ticket would be better than a PR, from me at least.

I'd be happy to test any branches that attempt to fix this, knowing a lot of devs here don't use Windows.

Thanks!

### Screenshots, Traceback

_No response_

### Steps to Reproduce

1. Start a new PowerShell Core session
2. `cd ~/src/project`
3. `nvim .`
4. `:e example.cs`
5. `:Neotree reveal`

### Expected Behavior

Neotree UI displays itself and navigates to the current buffer's file path.

### Your Configuration

```Lua
-- Yes, I tested this and it does reproduce.

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  -- add any other plugins here
}

local neotree_config = {
  dependencies = { "MunifTanjim/nui.nvim", "nvim-tree/nvim-web-devicons", "nvim-lua/plenary.nvim" },
  cmd = { "Neotree" },
  keys = {
  },
  opts = {
    -- Your config here
    -- ...
  },
}

  root = root .. "/plugins",
})

-- add anything else here
```

gplusplus314:
Fixes #1256 . One thing to note, NeoVim normalizes paths differently on Windows than a lot of plugins do, so I did this in a way that keeps it consistent with Neo-tree's code, which is to use backslashes on Windows.

### DIFF
--- a/lua/neo-tree/utils/init.lua
+++ b/lua/neo-tree/utils/init.lua
@@ -761,6 +761,11 @@ M.normalize_path = function(path)
   if M.is_windows then
     -- normalize the drive letter to uppercase
     path = path:sub(1, 1):upper() .. path:sub(2)
+    -- Turn mixed forward and back slashes into all forward slashes
+    -- using NeoVim's logic
+    path = vim.fs.normalize(path)
+    -- Now use backslashes, as expected by the rest of Neo-Tree's code
+    path = path:gsub("/", M.path_separator)
   end
   return path
 end


cseickel (approved):
Thanks!

Issue #170, opened 2022-03-03
[Suggestion] Use default highlight group for diagnostic icons


AndreAugustoAAQ:
The plugin can use custom diagnostic icons
By default `nvim` already has the `DiagnosticSign*` highlight group
But currently the plugin uses `LspDiagnosticsSign`, as seen [here][0]
It would be interesting to change the origin of the icons to the `nvim` default

A plugin that would be useful for this is [ray-x/navigator.lua][1], which sets icons using the default highlight group
This would make it easier to define icons, and also more practical to have uniform icons in `nvim`
If the person already uses [ray-x/navigator.lua][1], the icons will be taken from there, or if not, you can set the icons as it is currently done, just changing the name of the highlight used.

The [function where the severity is obtained][severity] is also necessary to change, changing Warning to Warn and Information to Info

<details>
<summary>Severity function code</summary>

Current code:
```lua
local diag_severity_to_string = function(severity)
  if severity == vim.diagnostic.severity.ERROR then
    return "Error"
  elseif severity == vim.diagnostic.severity.WARN then
    return "Warning"
  elseif severity == vim.diagnostic.severity.INFO then
    return "Information"
  elseif severity == vim.diagnostic.severity.HINT then
    return "Hint"
  else
    return nil
  end
end
```

Modified code:
```lua
local diag_severity_to_string = function(severity)
  if severity == vim.diagnostic.severity.ERROR then
    return "Error"
  elseif severity == vim.diagnostic.severity.WARN then
    return "Warn"
  elseif severity == vim.diagnostic.severity.INFO then
    return "Info"
  elseif severity == vim.diagnostic.severity.HINT then
    return "Hint"
  else
    return nil
  end
end
```

</details>

If the suggestion is accepted, I can make these changes

[0]: https://github.com/nvim-neo-tree/neo-tree.nvim/blob/0fd720115887f006fe4867e65c68ab24d95b51dd/lua/neo-tree/sources/common/components.lua#L60
[severity]: https://github.com/nvim-neo-tree/neo-tree.nvim/blob/0fd720115887f006fe4867e65c68ab24d95b51dd/lua/neo-tree/utils.lua#L11-L23
[1]: https://github.com/ray-x/navigator.lua "ray-x/navigator.lua"

cseickel:
[...]

PR #1265, opened 2023-12-17
fix: compatibility with 0.10 extmark sign

ofseed:
After https://github.com/neovim/neovim/pull/26193 , the internal implementation of the diagnostic signs changed to the new extmark-based sign, `sign_getdefine` will no longer work.

### DIFF
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -74,19 +74,10 @@ M.current_filter = function(config, node, state)
   }
 end
 
-M.diagnostics = function(config, node, state)
-  local diag = state.diagnostics_lookup or {}
-  local diag_state = diag[node:get_id()]
-  if config.hide_when_expanded and node.type == "directory" and node:is_expanded() then
-    return {}
-  end
-  if not diag_state then
-    return {}
-  end
-  if config.errors_only and diag_state.severity_number > 1 then
-    return {}
-  end
-  local severity = diag_state.severity_string
+---`sign_getdefined` based wrapper with compatibility
+---@param severity string
+---@return vim.fn.sign_getdefined.ret.item
+local function get_defined_sign(severity)
   local defined = vim.fn.sign_getdefined("DiagnosticSign" .. severity)
   if not defined then
     -- backwards compatibility...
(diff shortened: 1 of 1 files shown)


lua/neo-tree/sources/common/components.lua
```diff
@@ -74,6 +74,45 @@ M.current_filter = function(config, node, state)
   }
 end
 
+---`sign_getdefined` based wrapper with compatibility
+---@param severity string
+---@return vim.fn.sign_getdefined.ret.item
+local function get_defined_sign(severity)
+  local defined
+
+  if vim.fn.has("nvim-0.10") then
```
wookayin, line 83:
```suggestion
  if vim.fn.has("nvim-0.10") > 0 then
```

lua/neo-tree/sources/common/components.lua
```diff
@@ -74,6 +74,45 @@ M.current_filter = function(config, node, state)
...
+        identifier = "N"
+      end
+      defined = {
+        text = signs_config.text[vim.diagnostic.severity[identifier]],
+        texthl = "DiagnosticSign" .. severity,
+      }
+    end
+  else -- before 0.10
+    defined = vim.fn.sign_getdefined("DiagnosticSign" .. severity)
+    if not defined then
```
wookayin, line 97: Same; In lua, an empty table `{}` evaluates to true, so this will be never executed.

wookayin (changes requested):
This would need be tested more comprehensively in both versions

lua/neo-tree/sources/common/components.lua
ofseed: This branch is extracted from the original, it's maybe an undiscovered bug.

ofseed:
Note: I fixed the unreachable branch in https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1265/commits/374e00db5ac7e4862c408752d59021266c052e7c as @wookayin mentioned above, but I am not sure if this branch is truly needed because no one reported it since https://github.com/nvim-neo-tree/neo-tree.nvim/issues/170 .

cseickel (approved):
Sorry for the delay here. I held off because a test was failing, but I realize that it is just a flaky test that needs to be fixed so I'll push this through.

PR #1278, opened 2023-12-31
doc: correct table for git symbols to be within git_status table

aedificatio:
For my system the symbols-table needs to be within a git_status-table. 
My system has the versions below installed:
"lazy.nvim": { "branch": "main", "commit": "96584866b9c5e998cbae300594d0ccfd0c464627" } 
"neo-tree.nvim": { "branch": "v3.x", "commit": "77d9f484b88fd380386b46ed9206e5374d69d9d8" }

### DIFF
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -1040,18 +1040,20 @@ the following properties:
 >lua
     require("neo-tree").setup({
       default_component_configs = {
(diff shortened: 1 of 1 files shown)


cseickel (approved):
I'm surprised no one pointed this out before. Thanks for the fix!

PR #1280, opened 2023-12-31
fix(renderer): after focusing a node, such node was always focused

### LINKED ISSUE #1279: BUG: previously revealed file incorrectly focused on all future renders
> Sorry for mentioning that this pr seems breaking the `open` action in `Neotree reveal`.
> 
> Open one file and run `Neotree reveal`, then run `<cr>` to `open` the directory of that file.
> 
> After running `<CR>`, it jumps to the file location not toggle the directory.

That is true. Beyond that, any action you take will result in the previously revealed file being focused. That includes opening other files. Whatever the last revealed file is will be permanently stuck and refocused after every render.

I will make this a new issue.

_Originally posted by @cseickel in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1274#issuecomment-1872982362_

georgeguimaraes:
Sorry for the bug. This should fix it.

Do you prefer to have the code without the comment?

Fixes #1279

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -668,7 +668,11 @@ M.position = {
         end)
       end
       if state.position.node_id then
+        log.debug("Focusing on node_id: " .. state.position.node_id)
         M.focus_node(state, state.position.node_id, true)
+        -- After focusing a node, we clear it so that subsequent renderer.position.restore don't
+        -- focus on it anymore
+        state.position.node_id = nil
       end
     else
       log.debug("Position is not restorable")


cseickel (approved):
Thanks for the quick fix. No need to remove the comment, you can never have too many comments.

cseickel:
Darn. I got so used to bypassing one flaky test that I let this through without noticing that it broke more tests. I'm removing that flaky test now and I'll take a look at the newly broken tests:

`:Neotree show toggle` should retain the focused node on next show
`:Neotree focus toggle` should retain the focused node on next focus

georgeguimaraes:
I'll take a look too.

georgeguimaraes:
Yeah, when we call manager.close, I'm not saving the position in this new way. So the cursor is always at the first line.
I'm experimenting adding renderer.position.save on some other points of code, but still no success yet. I'm still trying though, just wanted to give a heads up.

georgeguimaraes:
PR with fix available at https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1280. Tests passing!

PR #1294, opened 2024-01-07
feat(filesystem): open in right below or left above split

Blackmorse:
This pull request addresses a common scenario where I often open files in a split, but I want control over whether the split appears on the left or right relative to the last buffer. With this update, I've introduced the option to open files using `rightabove`  or `leftbelow vs`. This enhancement provides more control over split placement, offering flexibility based on user preference. I believe this feature could be beneficial for users who seek a more tailored split behavior.


cseickel:
Since we have a vsplit, it make sense to have these other variations as well. For completeness, you should also have version of the new command available in the `git_status` and `buffers` sources.

There should also be documentation of the new commands in `:h neo-tree-navigation`.

Blackmorse:
@cseickel , thanks for you review

I've added documentation for the new command.

Regarding to support for `git_status` and `buffers` sources, I believe that new commands are already available for them, since they should be added by the call of `_add_common_commands` at `<source>/commands.lua` (the same as for `split` and `vsplit`)

cseickel:
> Regarding to support for `git_status` and `buffers` sources, I believe that new commands are already available for them, since they should be added by the call of `_add_common_commands` at `<source>/commands.lua` (the same as for `split` and `vsplit`)

You're absolutely correct, sorry about that.

cseickel (approved):

Issue #1290, opened 2024-01-04
BUG: attempt to index field 'text' (a nil value)


kawre:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

v0.10.0-dev-2022+ga064ed6229

### Operating System / Version

6.6.8-arch1-1

### Describe the Bug

[Neo-tree WARN] Error rendering component diagnostics: ...neo-tree.nvim/lua/neo-tree/sources/common/components.lua:91: attempt to index field 'text' (a nil value)
One fix i found is to set update_in_insert to true as i showcased in the video.
Also at 0;25 you can see that the filename changed to that error

### Screenshots, Traceback

https://github.com/nvim-neo-tree/neo-tree.nvim/assets/69250723/706a396d-51d6-47cf-ae47-21eed9dba4c5

### Steps to Reproduce

1. open `.lua` file
2. type anything that will trigger diagnotic errors once `lua_ls` loads
3. wait and stay in insert mode
4. once `lua_ls` loads it should give an warning

### Expected Behavior

not throw an error?

### Your Configuration

I don't think reproducing this is needed but if it is tell me 

lspconfig - https://github.com/kawre/nvim/tree/master/lua/plugins/lsp/config

```Lua
local M = {
    "nvim-neo-tree/neo-tree.nvim",
    dependencies = {
        "nvim-lua/plenary.nvim",
        "nvim-tree/nvim-web-devicons",
        "MunifTanjim/nui.nvim",
    },
    lazy = false,
    enabled = true,
}

M.keys = {
    { "<leader>e", "<cmd>Neotree left toggle<cr>", desc = "Toggle Neotree" },
}

M.init = function()
    vim.g.loaded_netrw = 1
    vim.g.loaded_netrwPlugin = 1

    vim.cmd([[hi link NeoTreeGitAdded String]])
end

M.opts = {
    window = {
        position = "left",
        width = 60,
    },
    default_component_configs = {
        indent = {
            indent_size = 3,
            highlight = "Comment",
        },
        icon = {
            folder_closed = "",
            folder_open = "",
            folder_empty = "󰜌",
            -- The next two settings are only a fallback, if you use nvim-web-devicons and configure default icons there
            -- then these will never be used.
            default = "*",
            highlight = "NeoTreeFileIcon",
        },
        modified = {
            symbol = "",
[...]

PR #1295, opened 2024-01-07
fix(diagnostic): `text` is `nil`

kawre:
#1290

### DIFF
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -88,7 +88,7 @@ local function get_defined_sign(severity)
         identifier = "N"
       end
       defined = {
-        text = signs_config.text[vim.diagnostic.severity[identifier]],
+        text = (signs_config.text or {})[vim.diagnostic.severity[identifier]],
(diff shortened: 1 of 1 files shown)


cseickel (commented):
Thanks for submitting this fix. 

One thing I would change would be that if the text lookup fails, I think it would be better to fall back to the first letter of the severity code as the icon instead of just letting it be blank.

kawre:
@cseickel Isn't this handled by this function? https://github.com/nvim-neo-tree/neo-tree.nvim/blob/396ff81e6bb384b0ef12ae427346cbf97d08f808/lua/neo-tree/sources/common/components.lua#L116C4-L116C4

cseickel:
> @cseickel Isn't this handled by this function? https://github.com/nvim-neo-tree/neo-tree.nvim/blob/396ff81e6bb384b0ef12ae427346cbf97d08f808/lua/neo-tree/sources/common/components.lua#L116C4-L116C4

Yes, it looks like it is. Sorry I didn't remember about that as I was reading the changes.

cseickel (approved):

PR #1327, opened 2024-01-29
fix(filesystem): fix filesystem path on windows platform

konosubakonoakua:
`\` and `/` both appears in the path, lead to string comparision problem on windows platform, user can not focus on the buffer file.

### DIFF
--- a/lua/neo-tree/sources/filesystem/init.lua
+++ b/lua/neo-tree/sources/filesystem/init.lua
@@ -41,8 +41,6 @@ local follow_internal = function(callback, force_show, async)
   end
   if not state.path then
     return false
-  else
-    state.path = utils.normalize_path(state.path)
   end
   local window_exists = renderer.window_exists(state)
   if window_exists then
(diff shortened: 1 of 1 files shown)


lua/neo-tree/sources/filesystem/init.lua
```diff
@@ -41,6 +41,8 @@ local follow_internal = function(callback, force_show, async)
   end
   if not state.path then
     return false
+  else
+    state.path = utils.normalize_path(state.path)
```
cseickel, line 45: This fix should be in `M._navigate_internal` before the path is first set rather than here. Probably at line 125 just before

``` lua
if path ~= state.path then
```

should be:

``` lua
path = utils.normalize_path(state.path)
if path ~= state.path then
```

konosubakonoakua: fixed, please review again. 😄

cseickel (approved):
Thanks for the fix!

cseickel:
This seems to have completely broken the plugin on linux.

konosubakonoakua:
> This seems to have completely broken the plugin on linux.

fixed 😃

lua/neo-tree/sources/filesystem/init.lua
```diff
@@ -120,6 +120,8 @@ M._navigate_internal = function(state, path, path_to_reveal, callback, async)
     log.debug("navigate_internal: path is nil, using cwd")
     path = manager.get_cwd(state)
   end
+  path = utils.normalize_path(path)
+  state.path = utils.normalize_path(state.path)
```
cseickel, line 124: I think the line above was the fix but this line is not needed:

```suggestion
```

cseickel (approved):

PR #1328, opened 2024-01-29
fix(prompt): confirm `cwd` changing using `vim.fn.comfirm`

konosubakonoakua:
It is kind of cumbersome to input "y or n", and press <kbd>enter</kbd> for the confirm process, why not just press <kbd>y</kbd> or <kbd>n</kbd> instead.

new prompt

### DIFF
--- a/lua/neo-tree/ui/inputs.lua
+++ b/lua/neo-tree/ui/inputs.lua
@@ -91,12 +91,7 @@ M.confirm = function(message, callback)
     input.prompt_type = "confirm"
     M.show_input(input)
   else
-    local opts = {
-      prompt = message .. " y/n: ",
-    }
-    vim.ui.input(opts, function(value)
-      callback(value == "y" or value == "Y")
-    end)
+    callback(vim.fn.confirm(message, "&y[Y]\n&n[N]"))
   end
 end
 
--- a/lua/neo-tree/ui/inputs.lua
+++ b/lua/neo-tree/ui/inputs.lua
@@ -91,7 +91,7 @@ M.confirm = function(message, callback)
     input.prompt_type = "confirm"
     M.show_input(input)
   else
-    callback(vim.fn.confirm(message, "&y[Y]\n&n[N]"))
+    callback(vim.fn.confirm(message, "&Yes\n&No") == 1)
   end
 end
 


cseickel:
I didn't even know that existed, do you know what version of nvim it was added in?

konosubakonoakua:
> I didn't even know that existed, do you know what version of nvim it was added in?

I found some thing here: https://github.com/neovim/neovim/blame/5e5b004da44c7075ed1e20ae9d05ab09c6f2ac58/runtime/doc/usr_41.txt#L974

Looks like it is a vim feature, introduced at least 10 years ago. 😄 

BTW, I'm using `noice.nvim` which takes over the `confirm` ui.
The prompt will show up just below the status bar without the plugin.
You can use `nvim -u NONE -c ':lua vim.fn.confirm("Ok or Not?", "&y\n&n")'` for a test.

konosubakonoakua:
BTW, shall we replace all the confirm prompt with `vim.fn.confirm`?

lua/neo-tree/command/init.lua
```diff
@@ -219,14 +219,13 @@ handle_reveal = function(args, state)
...
     cwd, _ = utils.split_path(path)
-    inputs.confirm("File not in cwd. Change cwd to " .. cwd .. "?", function(response)
-      if response == true then
-        args.dir = cwd
-      else
-        args.reveal_file = nil
-      end
-      do_show_or_focus(args, state, true)
-    end)
+    local response = vim.fn.confirm("File not in cwd. Change cwd to " .. (cwd or "nil") .. " ?", "&y[Y]\n&n[N]")
```
cseickel, line 222: This is entirely the wrong place to do this. This sort of change should go into the `inputs.confirm` function so that it applies to all confirmation prompts. It should be in the `else` branch of `use_popups_for_input()` check since this will not be a popup by default.

lua/neo-tree/ui/inputs.lua
```diff
@@ -91,12 +91,7 @@ M.confirm = function(message, callback)
     input.prompt_type = "confirm"
     M.show_input(input)
   else
-    local opts = {
-      prompt = message .. " y/n: ",
-    }
-    vim.ui.input(opts, function(value)
-      callback(value == "y" or value == "Y")
-    end)
+    callback(vim.fn.confirm(message, "&y[Y]\n&n[N]"))
```
cseickel, line 94: This does not work at all. It will confirm like "yes" for any choice. You need to test your code before you submit a PR.

Also, I think `"&Yes\n&No"` makes for a better set of choices.

I think this will work:
```suggestion
    callback(vim.fn.confirm(message, "&Yes\n&No") == 1)
```

konosubakonoakua: Sorry for the imprudence，and thanks for your patience.
It does need the `==` to handle for different inputs, as selecting `no` returns `2`, and <C-c> returns `0`.

cseickel (approved):

Issue #960, opened 2023-05-30
BUG: Plugin doesn't work in windows with folder with restrictive permissions


lnc3l0t:
When using the plugin in Windows within a folder that contains subfolders with restricted permissions, the plugin attempts to open the folders using `fs_opendir` function https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8c89efb93b8383666a6f989762c2f610f5f2c930/lua/neo-tree/sources/filesystem/lib/fs_scan.lua#L152. 
However, this operation fails, and when `print(vim.inspect(dir))` is executed, it displays "EPERM Operation not permitted" in addittion to return nil. Despite the error, success from pcall is still evaluated as true.

As a result of this issue, when the plugin tries to close the directory immediately after, https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8c89efb93b8383666a6f989762c2f610f5f2c930/lua/neo-tree/sources/filesystem/lib/fs_scan.lua#L163 it raises an error. This error can be one of the following, depending on whether you ran `nvim` or `nvim file`:
```bash
[Neo-tree ERROR] debounce  filesystem_navigate  error:  ...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:165: bad argument #1 to 'fs_closedir' (uv_dir expected, got nil)
```
or
```bash
Error executing luv callback:
...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:165: bad argument #1 to 'fs_closedir' (uv_dir expected, got nil)
stack traceback:
        [C]: in function 'fs_closedir'
        ...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:165: in function 'get_children_sync'
        ...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:208: in function 'callback'
        ...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:180: in function <...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:171>
```

pysan3:
Shall we close this? @cseickel

cseickel:
yup

PR #1343, opened 2024-02-05
fix(filesystem): Add error handler for `get_children_async()`

pynappo:
Similar to #960 but for `get_children_async`.

Sometimes, `get_children_async` may try to read from folders with restrictive permissions and libuv will error out. For example, I use `scan_mode = "deep"`, and opening neo-tree in the Windows $HOME results in a lot of errors because Windows puts a lot of folders with restrictive permissions there (for some reason). This PR will just callback with an empty list of children in that case.

On a tangential note, I think the fs_scan file could use a run of `stylua`, some code seems to have extra indentation and there's `function (...)` (note the space) instead of `function(...)`. I didn't want to make the diff too big for a small bugfix though.

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/fs_scan.lua
+++ b/lua/neo-tree/sources/filesystem/lib/fs_scan.lua
@@ -207,6 +207,13 @@ end
 
 local function get_children_async(path, callback)
   uv.fs_opendir(path, function(_, dir)
+    if err then
+      if dir then
+        uv.fs_closedir(dir)
+      end
+      callback({})
+      return
+    end
     uv.fs_readdir(dir, function(_, stats)
       local children = {}
       if stats then
--- a/lua/neo-tree/sources/filesystem/lib/fs_scan.lua
+++ b/lua/neo-tree/sources/filesystem/lib/fs_scan.lua
@@ -208,9 +208,6 @@ end
 local function get_children_async(path, callback)
   uv.fs_opendir(path, function(_, dir)
     if err then
-      if dir then
-        uv.fs_closedir(dir)
-      end
       callback({})
       return
     end


lua/neo-tree/sources/filesystem/lib/fs_scan.lua
```diff
@@ -207,6 +207,10 @@ end
 
 local function get_children_async(path, callback)
   uv.fs_opendir(path, function(_, dir)
+    if err then
```
cseickel, line 210: Where does this `err` variable come from?

pynappo: oh oops. i messed up a different branch and copied the main part of the fix over but not all the changes. it should come from the first arg of the opendir callback.

lua/neo-tree/sources/filesystem/lib/fs_scan.lua
```diff
@@ -206,7 +206,11 @@ local function get_children_sync(path)
 end
 
 local function get_children_async(path, callback)
-  uv.fs_opendir(path, function(_, dir)
+  uv.fs_opendir(path, function(err, dir)
+    if err then
```
cseickel, line 210: It's not correct to just silently disregard every error. It should be specific about what errors can be ignored. In all cases, the error should be logged at some level. Probably `debug` for ignored errors and let it crash for other errors.

pynappo: Understood, just pushed a few commits that log permission errors to `debug` and lets the other errors crash.

Also, I see that `get_children_sync` logs to `error` - would that be preferable for the other errors?

cseickel: Now that you mention it, they should be the same. How about they both check for and ignore permission based errors and log an error for anything else.

pynappo: Cool, I made them handle errors the same now and added a check for the Unix permission-based error. This PR now resolves any blocking error spam when encountering permission problems on my Linux and Windows machines in both sync and async modes.

I also removed the `pcall` surrounding the libuv calls, because those libuv functions return `nil, err, name` (the latter 2 are strings) when it errors anyways, so we can check for those instead of pcall's success value if needed.

Issue #1350, opened 2024-02-11
BUG: `close_if_last_window` invalid command to swap to modified buffer


bwpge:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.9.2

### Operating System / Version

Windows 11 22H2

### Describe the Bug

When the `close_if_last_window` option is enabled, the "swap to modified buffer" scheduled function in `init.lua` causes the error `E492: Not an editor command: ...`. This only happens for buffers of files on disk (not unnamed buffers), specifically on a Windows system. This does not occur on WSL for the same system.

**Possible root cause:**
Looking at the traceback, it looks like this occurs because the `:b[uffer]` command is concatenating with an absolute Windows path (e.g., starting with a partition letter) resulting in an "unknown" command such as `:bC:/...` (see line 310 below): https://github.com/nvim-neo-tree/neo-tree.nvim/blob/cfe1920c5dfb0524b3a13e827c35b6eb571143aa/lua/neo-tree/setup/init.lua#L306-L311

**Example `:b` test on Windows:**
- Open a file in Neovim with `nvim C:/foo.txt`
- Both `:b C:\foo.txt` and `:b C:/foo.txt` work as expected
- Both `:bC:\foo.txt` and `:bC:/foo.txt` causes the error `E492: Not an editor command: bC:...`

### Screenshots, Traceback
```
Error executing vim.schedule lua callback: vim/_editor.lua:341: nvim_exec2(): Vim:E492: Not an editor command: bC:/Users/user/AppData/Local/nvim/lua/user/plugins/neo-tree.lua
stack traceback:
        [C]: in function 'nvim_exec2'
        vim/_editor.lua:341: in function 'cmd'
        ...nvim-data/lazy/neo-tree.nvim/lua/neo-tree/setup/init.lua:310: in function <...nvim-data/lazy/neo-tree.nvim/lua/neo-tree/setup/init.lua:306>
```

### Steps to Reproduce

1. Use Neovim on a Windows system with this plugin, minimally configured with `lazy` per [README](https://github.com/nvim-neo-tree/neo-tree.nvim?tab=readme-ov-file#minimal-quickstart). Minimal `init.lua` provided below, but the following is the relevant plugin spec for neo-tree to create the behavior:
```lua
{
    "nvim-neo-tree/neo-tree.nvim",
[...]

PR #1351, opened 2024-02-11
fix: make buffer command compatible with Windows paths (#1350)

bwpge:
This is an incredibly boring PR. It adds a space to the `b[uffer]` command so that buffer names that use absolute Windows paths can be used without causing errors. This is in line with usage described in [`:h :buffer`](https://neovim.io/doc/user/windows.html#%3Ab) and [`:h {bufname}`](https://neovim.io/doc/user/windows.html#%7Bbufname%7D).

I've manually tested this change on Windows and WSL with the following types of buffers to ensure no regressions as best I can:
- No name buffers
- Buffer names with no spaces
- Buffer names with spaces in the path but not filename
- Buffer names with spaces in the filename but not path
- Buffer names with spaces in both path and filename

All cases result in the correct behavior: a warning is issued and the unsaved buffer is focused, without causing errors. Unfortunately I don't have a macOS machine handy to test but I imagine paths and buffer names behave the same or nearly the same as on WSL.

**PS:** The current implementation of swapping to a buffer by name may be problematic for buffer names that are numbers (see: [`:h :buffer`](https://neovim.io/doc/user/windows.html#%3Ab)). This is outside the scope of this PR, but for reference this comes from the `utils.get_opened_buffers` function: https://github.com/nvim-neo-tree/neo-tree.nvim/blob/cfe1920c5dfb0524b3a13e827c35b6eb571143aa/lua/neo-tree/utils/init.lua#L314-L331

### DIFF
--- a/lua/neo-tree/setup/init.lua
+++ b/lua/neo-tree/setup/init.lua
@@ -307,7 +307,7 @@ M.win_enter_event = function()
               log.warn(message)
               vim.cmd("rightbelow vertical split")
               vim.api.nvim_win_set_width(win_id, state.window.width or 40)
-              vim.cmd("b" .. buf_name)
+              vim.cmd("b " .. buf_name)
             end)
             return
           end


cseickel (approved):
> This is an incredibly boring PR.

I love boring. Boring is good. Thanks for the fix!

bwpge:
Happy to help, and sorry for goofing up the commit message style (thought I was supposed to add my issue number in parens).

Thanks for the quick merge!

Issue #1352, opened 2024-02-12
BUG: directories beginning or ending with symbol cause error in filesystem command `set_root`


bwpge:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.9.2

### Operating System / Version

Windows 11 22H2

### Describe the Bug

On Windows systems, Neo-tree throws an error when using the filesystem source `set_root` command to set the current working directory when a path segment begins or ends with a symbol (such as `(`, `&`, etc.). This type of path is common with web projects that use "dynamic" paths such as `foo/(bar)/baz`. This is almost the same issue reported in #889, and is part of a bigger issue of properly handling Windows paths.

The Neo-tree view is able to properly update itself to the correct root node, but Neovim will not have the current working directory updated (confirmed with `:pwd`).

**Root cause:**
This is a Windows-specific path problem. The `set_cwd` function uses `lcd`, `tcd`, or `cd` commands to set the working directory (see snippet below). On Windows the argument passed to these commands is having the path separator `\` treated as an escape character for a symbol such as `\(`. The path separator is lost and the final path is treated as `foo(bar)` instead of `foo\(bar)`.
https://github.com/nvim-neo-tree/neo-tree.nvim/blob/f3941c57ec85d7bdb44fa53fd858fd80f159018f/lua/neo-tree/sources/manager.lua#L353-L360

### Screenshots, Traceback

![image](https://github.com/nvim-neo-tree/neo-tree.nvim/assets/114827766/1fdfb8cc-2e1c-463a-8336-525e967ac96b)

```
E5108: Error executing lua: vim/_editor.lua:341: nvim_exec2(): Vim(tcd):E344: Can't find directory "X:\foo(bar)" in cdpath
stack traceback:
        [C]: in function 'nvim_exec2'
        vim/_editor.lua:341: in function 'cmd'
        ...data/lazy/neo-tree.nvim/lua/neo-tree/sources/manager.lua:357: in function 'set_cwd'
        ...y/neo-tree.nvim/lua/neo-tree/sources/filesystem/init.lua:160: in function '_navigate_internal'
        ...o-tree.nvim/lua/neo-tree/sources/filesystem/commands.lua:231: in function <...o-tree.nvim/lua/neo-tree/sources/filesystem/commands.lua:224>
```

### Steps to Reproduce

[...]

Issue #1264, opened 2023-12-17
BUG: modified status and diagnostic does not change for buffers that opened by neo-tree


mrbeardad:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.9.4

### Operating System / Version

Windows 11

### Describe the Bug

File modified status and diagnostic status does not change for the most file, only works for the file I opened with `nvim works_file`

### Screenshots, Traceback

https://github.com/nvim-neo-tree/neo-tree.nvim/assets/54128430/56721b41-3ad8-4b23-a04a-366519df9623

### Steps to Reproduce

1. `nvim README.md`
2. `:Neotree`
3. modify README.md, modified status show in neotree
4. modify CONTRIBUTING.md, modified status does not show
5. diagnostic status is similar to modified status

### Expected Behavior

modified status and diagnostic status should change for step 4 above.

### Your Configuration

```Lua
if vim.g.vscode then
  return
end

-- bootstrap plugins manager
local plugins_path = vim.fn.stdpath("data") .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(plugins_path) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    "--branch=stable",
    plugins_path,
  })
end
vim.opt.rtp:prepend(plugins_path)

-- load user configs before load plugins
--vim.cmd("runtime! lua/user/configs/*.lua")

require("lazy").setup({
  -- load plugin specifics in directories
  spec = {
    --{ import = "user/plugins" },
    --{ import = "user/langs" },
    { import = "user/debug" },
  },
  defaults = {
    -- do not lazy load plugins by default
    lazy = false,
    -- always use the latest git commit, set to "*" for latest stable version
    version = false,
    -- disable unnecessary plugins for vscode
    cond = function()
      return true
    end,
  },
  install = {
    -- automatically install missing plugins
    missing = true,
    -- try to load one of these colorschemes when starting an installation during startup
    colorscheme = { "tokyonight", "habamax" },
  },
  -- do not automatically check for plugin updates
  checker = { enabled = false },
  -- do not automatically check for config file changes and reload the ui
  change_detection = { enabled = false },
  performance = {
    rtp = {
      disabled_plugins = {
        "gzip",
[...]

Issue #1157, opened 2023-09-19
BUG: attempt to concatenate local 'cwd' (a nil value)


Serranya:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.9.2

### Operating System / Version

Windows 10 Pro 22H2

### Describe the Bug

I get the following error when trying to open neotree the second time in the same directory.

```
E5108: Error executing lua: ...im-data/lazy/neo-tree.nvim/lua/neo-tree/command/init.lua:219: attempt to concatenate local 'cwd' (a nil value)
stack traceback:
	...im-data/lazy/neo-tree.nvim/lua/neo-tree/command/init.lua:219: in function 'handle_reveal'
	...im-data/lazy/neo-tree.nvim/lua/neo-tree/command/init.lua:156: in function 'execute'
	...ers\peter.lamby\AppData\Local\nvim/lua/serra/neotree.lua:29: in function <...ers\peter.lamby\AppData\Local\nvim/lua/serra/neotree.lua:28>
```

Of note:
While `:Neotree current show reveal` fails the second time `:Neotree current show` works just fine.

### Steps to Reproduce

1. Start nvim
2. `:cd ~\dev\inner\de.serra.parent`
3. <leader>e and open a file in the current directory
4. `:vs`
5. <C-w><Right>
6. <leader>e to open another file. (The error occurs here before neotree shows in the buffer)

### Expected Behavior

When opening neotree for the second time it should succeed.

### Your Configuration

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  -- add any other plugins here
}

local neotree_config = {
[...]

PR #1353, opened 2024-02-13
fix: improve path escaping for commands on Windows

bwpge:
This PR changes the handling of paths on Windows systems to address a few open issues. This builds on the work originally introduced in #1023 (to address #889). Realistically, this is another attempt to deal with a (neo)vim issue as mentioned in https://github.com/neovim/neovim/issues/3912 (no real update since 2015).

I've added some doc comments to help with maintainability, but if it's too verbose I'm happy to edit.

I've run tests locally and all pass, but I would really love to add regression tests for path escaping so we don't run into more problems in the future. However, I'm having trouble figuring out how to deal with platform-specific behavior of `vim.fn.fnameescape` (e.g., if you run tests on WSL or macOS, you won't get Windows path escaping). If anyone has any ideas for adding tests, I'm definitely all ears. Ideally, we would test both Windows and Unix-style path escaping regardless of the platform running the tests.

This PR fixes the following issues I could find:

- #1352 (opened by me when testing this fix)
- #1264

This also fixes the following external issues:
- https://github.com/nvim-lualine/lualine.nvim/issues/1183

The following issues *may* be fixed, but I have not confirmed:

- #1157

I think there are definitely more issues that could be addressed by this, but I'm reaching my issue-searching capability limits :)

I feel like I have a good grasp of how paths are being used in this limited capacity (such as with `set_cwd` and `open_file`) but I know this change could have cascading effects because it's modifying the input to commands like `:cd`, `:b`, `:vs`, etc. If anyone sees anything problematic with more context of the full codebase, I'm happy to discuss.

### DIFF
--- a/lua/neo-tree/utils/init.lua
+++ b/lua/neo-tree/utils/init.lua
@@ -664,7 +664,7 @@ M.open_file = function(state, path, open_cmd, bufnr)
   end
 
   if M.truthy(path) then
-    local escaped_path = M.escape_path(path)
+    local escaped_path = M.escape_path_for_cmd(path)
     local bufnr_or_path = bufnr or escaped_path
     local events = require("neo-tree.events")
     local result = true
@@ -1000,10 +1000,29 @@ M.windowize_path = function(path)
   return path:gsub("/", "\\")
 end
 
-M.escape_path = function(path)
+---Escapes a path primarily relying on `vim.fn.fnameescape`. This function should
+---only be used when preparing a path to be used in a vim command, such as `:e`.
+---
+---For Windows systems, this function handles punctuation characters that will
+---be escaped, but may appear at the beginning of a path segment. For example,
+---the path `C:\foo\(bar)\baz.txt` (where foo, (bar), and baz.txt are segments)
+---will remain unchanged when escaped by `fnaemescape` on a Windows system.
+---However, if that string is used to edit a file with `:e`, `:b`, etc., the open
+---parenthesis will be treated as an escaped character and the path separator will
+---be lost.
+---
+---For more details, see issue #889 when this function was introduced, and further
+---discussions in #1264 and #1352.
+---@param path string
+---@return string
+M.escape_path_for_cmd = function(path)
   local escaped_path = vim.fn.fnameescape(path)
   if M.is_windows then
-    escaped_path = escaped_path:gsub("\\", "/"):gsub("/ ", " ")
+    -- on windows, any punctuation preceeded by a `\` needs to have a second `\`
+    -- added to preserve the path separator. this is a naive replacement and
+    -- definitely not bullet proof. if we start finding issues with opening files
+    -- or changing directories, look here first.
+    escaped_path = escaped_path:gsub("\\%p", "\\%1")
   end
   return escaped_path
 end
--- a/lua/neo-tree/utils/init.lua
+++ b/lua/neo-tree/utils/init.lua
@@ -1237,6 +1237,7 @@ M.index_by_path = function(tbl, key)
   end
 
   -- on windows, paths that differ only by case are considered equal
+  -- TODO: we should optimize this, see discussion in #1353
   if M.is_windows then
     local key_lower = key:lower()
     for k, v in pairs(tbl) do
--- a/lua/neo-tree/utils/init.lua
+++ b/lua/neo-tree/utils/init.lua
@@ -1220,4 +1220,33 @@ M.brace_expand = function(s)
   return result
 end
 
+---Indexes a table that uses paths as keys. Case-insensitive logic is used when
+---running on Windows.
+---
+---Consideration should be taken before using this function, because it is a
+---bit expensive on Windows. However, this function helps when trying to index
+---with absolute path keys, which can have inconsistent casing on Windows (such
+---as with drive letters).
+---@param tbl table
+---@param key string
+---@return unknown
+M.index_by_path = function(tbl, key)
+  local value = tbl[key]
+  if value ~= nil then
+    return value
+  end
+
+  -- on windows, paths that differ only by case are considered equal
+  if M.is_windows then
+    local key_lower = key:lower()
+    for k, v in pairs(tbl) do
+      if key_lower == k:lower() then
+        return v
+      end
+    end
+  end
+
+  return value
+end
+
 return M
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -115,7 +115,7 @@ end
 
 M.diagnostics = function(config, node, state)
   local diag = state.diagnostics_lookup or {}
-  local diag_state = diag[node:get_id()]
+  local diag_state = utils.index_by_path(diag, node:get_id())
   if config.hide_when_expanded and node.type == "directory" and node:is_expanded() then
     return {}
   end
@@ -322,7 +322,7 @@ end
 
 M.modified = function(config, node, state)
   local opened_buffers = state.opened_buffers or {}
-  local buf_info = opened_buffers[node.path]
+  local buf_info = utils.index_by_path(opened_buffers, node.path)
 
   if buf_info and buf_info.modified then
     return {
--- a/lua/neo-tree/sources/manager.lua
+++ b/lua/neo-tree/sources/manager.lua
@@ -351,12 +351,13 @@ M.set_cwd = function(state)
 
   local _, cwd = pcall(vim.fn.getcwd, winid, tabnr)
   if state.path ~= cwd then
+    local path = utils.escape_path_for_cmd(state.path)
     if winid > 0 then
-      vim.cmd("lcd " .. state.path)
+      vim.cmd("lcd " .. path)
     elseif tabnr > 0 then
-      vim.cmd("tcd " .. state.path)
+      vim.cmd("tcd " .. path)
     else
-      vim.cmd("cd " .. state.path)
+      vim.cmd("cd " .. path)
     end
   end
 end


cseickel:
@pysan3 Can you review this? I know we have planned to incorporate pathlib but this is probably a good patch while we wait for that. Besides, making that library makes you the resident expert on path issues, and it would also be good for you to know about this bug to make sure it is handled in pathlib.

pysan3:
@bwpge Could you possibly test it with `vim.fn.shellescape(string, true)` instead? Or maybe the second argument can be omitted?

Unfortunately I don't have a Windows machine at hand and I can't test it right away.

This might bring what you want, but we need to make sure that this doesn't start breaking things in other places.

If `shellescape` doesn't do the trick, I think this approach is good enough if it works. Sorry again, I don't have a way to test it locally, so I can't say for sure...

pysan3:
@bwpge Could you also test simply passing the path to `vim.cmd.cd` without constructing a cmd string in the first place?

- `vim.cmd.cd(state.path)`
- `vim.cmd.lcd(state.path)`
- `vim.cmd.tcd(state.path)`

bwpge:
@pysan3 sure thing, I am headed to bed but I will test these in the morning and post findings. And I totally understand the apprehension, this is a big change so we should take extra time to explore its effects.

pysan3:
Thank you very much @bwpge !

So after all my ideas are not working so doing `gsub("\\%p", ...` is the best we can do now. 

I'll remember this issue and will try to fix it in the rewrite but this is a good fix for now.

bwpge:
Just adding some extra notes someone can reference when tackling this in a rewrite:

- The main issue to fix is the mixing/matching of how paths are handled in Neo-tree on Windows (`\` vs `/`)
  - Both path separators are "valid" insofar as they can point to a file or directory
  - (Neo)vim clearly handles them inconsistently
  - Makes it difficult to reliably do things like checking if paths contain other paths, dealing with relative paths, etc.
  - Sounds like pathlib aims to deal with this problem which would be great
- Default behavior of Neovim on windows is to use `\` separators
  - Such as if you tab-complete a path with `:e foo/b<TAB>`, you will get `:e foo\bar.txt` filled in
  - Other plugins tend to expect this behavior, as an example with `lualine` trying to detect git branch from the current buffer name
- Setting the working directory with `/` separators does not seem to be an issue
  - The value returned by `vim.fn.getcwd()` seems to always use `\` separators, even if set with `vim.cmd.cd[[X:/foo/bar]]`
- Opening a file with `/` will have side-effects, most notably changing the buffer name
  - **NOTE:** This only seems to happen if an absolute path with `/` separators is used
  - Editing a file like `:e foo/bar.txt` will still name the buffer `X:\foo\bar.txt` (checking the name with `vim.api.nvim_buf_get_name(0)`)
  - Editing a file `:e X:/foo/bar/baz.txt` **does** change the buffer name: `X:/foo/bar/baz.txt`

bwpge:
@pysan3 I apologize to hit you with so many pings, but I just stumbled across another Windows-path issue that can be fixed relatively easily.

The way `vim.lsp.buf.definition()` opens a file vs. what Neotree is using to compare buffers for modified status: Windows paths have a mismatch on drive letters (upper vs. lowercase) which breaks the buffer status on Neotree. This is a one-line change in `utils.normalize_path`, change `upper()` to `lower()` to fix the problem:

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/f3941c57ec85d7bdb44fa53fd858fd80f159018f/lua/neo-tree/utils/init.lua#L763

Unfortunately this affects the rendered path at the top of Neotree -- drive letters will be lowercase, which is a bit ugly. I suppose that could be cleaned up in the UI/renderer (just transform the string before rendering without changing the node path), but would require more changes.

Would you want me to add to this PR, or move to another PR?

pysan3:
Could you instead capitalize the input from lsp?

That's what is done in `vim.fs.normalize` so it should be better to align all drive letters as capital. 

`vim.fs.normalize` has too many flaws and can't be used here but it's good to keep consistency with upstream.

bwpge:
I've been playing with a simple solution this morning/afternoon to avoid touching too many things, I think it's a "good enough" solution without touching too much of the plugin behavior. I've added https://github.com/bwpge/neo-tree.nvim/commit/8afd20977348fe30b4c0fce21067272d81ac6c44 to this PR so you can see what I've changed (the explanation in a comment was getting too long and hard to follow) -- if we don't like it I can revert.

Basically we just change how we try to index into `state.diagnostics_lookup` and `state.opened_buffers` for nodes. This way everything in the plugin behaves the same, and we only have to do string manipulation for Windows clients, on a much more limited number of items.

The benefit of this solution is it catches other case-sensitivity issues Windows people might see since Windows doesn't treat different casing as different files. Also Linux/macOS users wouldn't have impact from this change, the function would just fall through and basically only pay the cost of two `if` checks.

lua/neo-tree/utils/init.lua
```diff
@@ -1201,4 +1220,33 @@ M.brace_expand = function(s)
...
+---@param tbl table
+---@param key string
+---@return unknown
+M.index_by_path = function(tbl, key)
+  local value = tbl[key]
+  if value ~= nil then
+    return value
+  end
+
+  -- on windows, paths that differ only by case are considered equal
```
pysan3, line 1239: Could you add a TODO comment to remember to optimize this function someday?

pysan3:
If you add the todo comment, I also agree this is a good _enough_ workaround @cseickel ;)

I'll remember to revisit this after I manage to fully move to pathlib. (I'm not gonna say a due date here, as it might be forever lol).

> Done is better than perfect.

lua/neo-tree/utils/init.lua
bwpge: Done!

pysan3:
LGTM @cseickel

cseickel (approved):
Thanks @bwpge and @pysan3.

pysan3:
@bwpge Looks like we unfortunately hit an edge case: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1382

These are the chars what will be matched with `"%p"` but could you instead test all out which ones will not cooperate with vim's `:e`?

```lua
-- The entire list [[!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~]]

-- This is from a google search; What characters are forbidden in Windows and Linux directory names?
-- < (less than)
-- > (greater than)
-- : (colon - sometimes works, but is actually NTFS Alternate Data Streams)
-- " (double quote)
-- / (forward slash)
-- \ (backslash)
-- | (vertical bar or pipe)
-- ? (question mark)
-- * (asterisk)

-- I deleted those from the list. But `\` and `<Space>` must indeed be escaped. 
local puncts = [[!#$%&'()+,-;=@[]^_`{}~\ ]]

-- Please remove characters that don't need to be escaped from the list.

escaped_path = vim.fn.escape(escaped_path, puncts)
```

bwpge:
@pysan3 here's the raw output from `:ls!` with your test above (very clever!) -- just a heads up you need add another closing paren on the `vim.cmd.edit` line:

```
  8  h   ".\tests\!foo.txt"             line 1
  9  h   ".\tests#foo.txt"              line 1
 10  h   ".\tests\$foo.txt"             line 1
 11  h   ".\tests%foo.txt"              line 1
 12  h   ".\tests&foo.txt"              line 1
 13  h   ".\tests'foo.txt"              line 1
 14  h   ".\tests(foo.txt"              line 1
 15  h   ".\tests)foo.txt"              line 1
 16  h   ".\tests\+foo.txt"             line 1
 17  h   ".\tests\,foo.txt"             line 1
 18  h   ".\tests\-foo.txt"             line 1
 19  h   ".\tests;foo.txt"              line 1
 20  h   ".\tests\=foo.txt"             line 1
 21  h   ".\tests\@foo.txt"             line 1
 22  h   ".\tests\[foo.txt"             line 1
 23  h   ".\tests\]foo.txt"             line 1
 24  h   ".\tests^foo.txt"              line 1
 25  h   ".\tests\_foo.txt"             line 1
 26  h   ".\tests`foo.txt"              line 1
 27  h   ".\tests\{foo.txt"             line 1
 28  h   ".\tests\}foo.txt"             line 1                                                                                                                                                                                    29 #h   ".\tests\~foo.txt"             line 1                                                                                                                                                                                    30 %a   ".\tests foo.txt"              line 1
```

Looks like the following is our list:
| Status | Symbol |
|---|---|
| Not working | `!$+,-=@[]_{}~` |
| Working | ``#%&'();^` `` (last is `<Space>`) |

I'll work on fixing this today and do some testing to make sure we don't have any regressions.

pysan3:
@bwpge When you are done, could you change the regarding code using `vim.fn.escape` instead and test it and submit a PR. My bad I cannot easily reach a Windows device...

```lua
escaped_path = vim.fn.escape(escaped_path, puncts)
```

bwpge:
Just to sanity check, my intent would be to do something like this:

```lua
M.escape_path_for_cmd = function(path)
   -- this is working on unix paths, don't touch for now
  local escaped_path = vim.fn.fnameescape(path)

  if M.is_windows then
    local puncts = [[!$+,-=@[]_{}~]]
    -- do extra windows escapes
  end
  return escaped_path
end
```

pysan3:
Yes and `:h escape()` does exactly that as I mentioned above.

bwpge:
Sorry meant to send this earlier:

Regarding `:h escape()`, part of our problem is that we only need the extra escape character if the symbol is preceded by a `\`. So for example, `.\foo\(bar\baz.txt` needs an extra escape on `(`, but `.\foo\bar)\baz.txt` does not need it on `)`. So I think it will need a more targeted approach than `vim.fn.escape`.

Will post when I have more info to share.

pysan3:
Oh shoot. You are right. I guess we need some kind of a complex regex.

Issue #1310, opened 2024-01-16
BUG: jump to selected item when opening folder


rick-yao:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.9.5 Build type: Release LuaJIT 2.1.1693350652

### Operating System / Version

macos 14.2.1 

### Describe the Bug

some times when opening folder, cursor will jump to a random item rather than stay at the foler.
some times it jump to opened file ,sometimes it jump to item in the folder. i couldn't get exact clue.

### Screenshots, Traceback

[replicate video](https://1drv.ms/v/s!Aul7Mp-RSA34gfxgj6FKIE1ps0TmPg?e=IeXjS9)

### Steps to Reproduce

1. open the file with neotree
2. focus on neotree
3. selecte a folder , press l to open
4. the cursor jump to opened item , not stay at the folder

### Expected Behavior

the cursor should stay at the folder item.

### Your Configuration

```Lua
i am using AstroNvim 3.41.2 with my personal config
{
    "nvim-neo-tree/neo-tree.nvim",
    opts = {
      filesystem = {
        filtered_items = {
          visible = true,
          show_hidden_count = true,
          hide_dotfiles = false,
          hide_gitignored = true,
          hide_by_name = {
            -- '.git',
            -- '.DS_Store',
            -- 'thumbs.db',
          },
          never_show = {},
        },
      },
      window = {
        position = "right",
      },
    },
  }
```

InfiniteRain:
experiencing the same thing

InfiniteRain:
update: downgrading to 3.14 has fixed the issue for me:

```lua
use({ "nvim-neo-tree/neo-tree.nvim", tag = "3.14" })
```

georgeguimaraes:
Hi, I'm gonna take a look into this. Probably related to the changes I made.

georgeguimaraes:
Hello, I tried and couldn't reproduce it with neovim 0.9.5 and with neo-tree's repro.lua (as stated in the bug report template):

```
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
	vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
[...]

PR #1288, opened 2024-01-03
fix(renderer): fix Neotree focus/show toggle behaviour

georgeguimaraes:
Fixes bug introduced by https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1280.

PR #1354, opened 2024-02-13
fix(renderer): Makes position.save() from render_tree a lower priority

georgeguimaraes:
This fixes some issues when several events trigger render_tree, and you open/browse a lot of files and directories.

A race condition happens when a position is saved but never restored.

I'm not super happy with the addition of another conditional. Suggestions are always welcomed.

Fixes: #1310

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -642,8 +642,11 @@ end
 
 ---Functions to save and restore the focused node.
 M.position = {
-  save = function(state)
-    if state.position.topline and state.position.lnum then
+  save = function(state, overwritable)
+    overwritable = overwritable or false
+
+    -- Save from render_tree can be overwritten.
+    if state.position.topline and state.position.lnum and not state.position.is.overwritable then
       log.debug("There's already a position saved to be restored. Cannot save another.")
       return
     end
@@ -652,6 +655,7 @@ M.position = {
       local win_state = vim.api.nvim_win_call(state.winid, vim.fn.winsaveview)
       state.position.topline = win_state.topline
       state.position.lnum = win_state.lnum
+      state.position.is.overwritable = overwritable
       log.debug("Saved cursor position with lnum: " .. state.position.lnum)
       log.debug("Saved window position with topline: " .. state.position.topline)
       -- Only need to restore the cursor state once per save, comes
@@ -678,6 +682,7 @@ M.position = {
         -- Clear saved position, so that we can save another position later.
         state.position.topline = nil
         state.position.lnum = nil
+        state.position.is.overwritable = nil
       end
       if state.position.node_id then
         log.debug("Focusing on node_id: " .. state.position.node_id)
@@ -1144,7 +1149,7 @@ render_tree = function(state)
   local should_pre_render = should_auto_expand or state.current_position == "current"
 
   log.debug("render_tree: Saving position")
-  M.position.save(state)
+  M.position.save(state, true)
 
   if should_pre_render and M.tree_is_visible(state) then
     log.trace("pre-rendering tree")


cseickel:
Thanks for tackling this. I get the general concept but could use more details.

Could you explain what position save are overwritable vs what is not and what are the sequence of events that caused you to make these choices?

georgeguimaraes:
Sure,

1) We need to save and restore the position in render_tree because of the changes in https://github.com/neovim/neovim/pull/24824.

2) We also need to save the position when we close neo-tree's window in case the user runs `:Neotree show` (without follow_current_file). That's what my PR #1288 meant to fix. We can rely on the fact that because of 1), render_tree will restore the position, but we don't want the M.position.save() from inside render_tree to overwrite the M.position.save() from when we close the window, that's why the guard clause was added.

Those are the 2 happy paths I was dealing with.

However, there is a race condition that it's difficult for me to detect in my machine. It appears that two events (vim_buffer_enter and vim_buffer_added, but I'm not entirely sure) trigger render_tree concurrently, and one tries to save the save position but can't. The problem is that we don't restore it and thus, we don't reset the position with nil. And now, the next open/expand command in neo-tree window will restore the window to a stale position.

I've considered using a queue here, but out-of-order messages would cause the same problem.

Since saving the position in render_tree is actually a workaround for a neovim behavior and not a feature, we can overwrite it if there's another part of the code that wants to save a position.

**TL;DR:**
M.position.save() from render_tree is always overwritable. In other places, it shouldn't be.

Thoughts?

mehalter:
@georgeguimaraes I'm able to periodically trigger this race condition still while browsing around in neo-tree. I do think it's happening less frequently than before but still present

georgeguimaraes:
Back to the drawing board, then. I'll try another approach.
Thanks, @mehalter and @XavierChanth, for trying it.

XavierChanth:
> Back to the drawing board, then. I'll try another approach. Thanks, @mehalter and @XavierChanth, for trying it.

Thanks, feel free to ping me as a tester

mehalter:
> Back to the drawing board, then. I'll try another approach.
> Thanks, @mehalter and @XavierChanth, for trying it.

No problem at all! Thanks so much for spending time on this @georgeguimaraes 😊

pysan3:
@mehalter Could you give me some examples of this problem?

I'm doing a [big rewrite](https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/1340) and I might be able to deal with async updates of tree using [mutex from nvim-nio](https://github.com/nvim-neotest/nvim-nio/blob/c037b0a496c8fddf285fe21785d625a3e245fd38/lua/nio/control.lua#L247-L267) (i.e. `semaphore(1)`).

mehalter:
@pysan3 wow that discussion thread looks amazing, and I am super excited to see what you put together! I'll be honest and say I don't fully understand the saving/restoring procedure happening here, like why it is necessary or why there is a race condition. At least no more than the incredible comment description above by @georgeguimaraes (https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1354#issuecomment-1943755175) . They might be able to explain better/more in detail if you need that.

Where I am producing the issue with when using the `hjkl` based navigation code proposed in this discussion thread: https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/163

There is a minimal `repro.lua` file in a comment in the linked issue above (https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1310#issuecomment-1895833654). Don't worry about the comment saying that it doesn't reproduce the issue, it does. The weird part is because it's a race condition it is relatively non-deterministic on whether or not the behavior appears and when. But when havingating using only the `hjkl` navigation opening files and opening/closing directories specifically with the `h` and `l` keys will sometimes mess up where the cursor location in neo-tree gets placed. There is a video of this behavior linked in the original issue as well that might be clearer than my explanation 😅 

When replicating, I basically just frantically move about the file system using `hjkl` in neo-tree and see if it randomly jumps to a previously opened directory. Sorry if this isn't super great replication instructions 😞 Let me know if you have any questions that I can try and clear up!

Let me know if you want any help testing/working on your new refactor!

pysan3:
Oh I missed the video link. Your explanation is awesome.

So, in the video, I believe this tree is opened with `reveal=true`, so the expected behavior would be; 

- expand `event/` -> cursor stays on event
- go back to the left window -> neo-tree focus goes back to `layout.tsx`
- reenter neo-tree -> cursor on `layout.tsx`
- move down to `event/`
- close `event/` -> cursor stays at event

right?

I believe the original video is more of an issue with `:Neotree show` + follow file vs user manually expanding some other location.

And I think your issue is that if you expand two different directories simultaneously (or just move the cursor after commanding an expand but before it completes the rerender), the cursor goes back to the first directory, where you would want the cursor to be on the final location and don't mess with my cursor right?

@mehalter

mehalter:
So the video is not showing any issues with the open file following the cursor actually. It is showing that the user moved the cursor down in the neo-tree window, presses `l` to open the directory now under the cursor (with the custom binding), and upon opening the directory their cursor was ("restored") moved up to a previous location rather than staying in the newly open directory.

Here is the video I was referring to from the issue just in case there may have been another one somewhere that maybe wasn't correct (error happens between second 4 and 5): https://photos.onedrive.com/share/F80D48919F327BE9!32352?cid=F80D48919F327BE9&resId=F80D48919F327BE9!32352&authkey=!AI-hSiBNabNE5j4&ithint=video&e=IeXjS9


It appears this is happening because of some stale saved information which "restores" the cursor but because it is stale it restores it to a previous, incorrect location.

pysan3:
Sorry I updated my comment. I was too late ig?

> (cursor pos after follow file will be remembered, but when user comes into neo-tree window and move the cursor manually, that change in position is not captured (which should be force updated right before keybind is executed, but is not) and on the next rerender, cursor goes back to the _current file_) (but this explanation is old, it should be fixed by whatever PR that I don't remember).

pysan3:
What you are trying to fix is more of a race condition thing right?

> And I think your issue is that if you expand two different directories simultaneously (or just move the cursor after commanding an expand but before it completes the rerender), the cursor goes back to the first directory, where you would want the cursor to be on the final location and don't mess with my cursor right?

---

I think the video bug can be reproduced even when done slowly.

mehalter:
yeah, the race condition is not because 2 actions are being taken. It's just inherent in the code. It's not a race condition because the user is moving too quickly

mehalter:
the bug in the video is the one that this PR is trying to fix and doesn't. I reproduce it by navigating frantically just to increase the number of operations I am doing so I don't spend a long time trying to reproduce the bug 😅 I may have communicated this poorly. It isn't dependent on how fast the user is moving, just how much time they want to spend trying to move, open directories, close them, move, etc. But the speed of the user is not a part of the issue :D

mehalter:
The main thing is the behavior in the video doesn't happen 100% of the time. It happens non-deterministically. So there aren't a set of steps I can give you to reproduce the bug 100% of the time. I couldn't give you like a set of `hjkl` operations that if you did within a time frame would reproduce the bug. It's like, it could happen the first time you open and close a directory, it could happen the 100th time you open and close a directory

pysan3:
Huh, that's weird.

What is the command when opening neo-tree? what is reveal? show? focus?

mehalter:
ok maybe I should just clear this up because I think you think this bug is related to the cursor following the open file. But it is completely unrelated. One sec. I will do a recording

pysan3:
Thanks a lot. Sorry to bother your time...

XavierChanth:
@pysan3 my issue is slightly different than @mehalter but it seems to be related. I was able to reproduce consistently, but only for the first folder opened each time I toggle.

Updated my [message above](https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1354#issuecomment-1944414872) with the full config + the specific keybind / command I was using. Note that I put my config in a toggle block since it is quite long.

mehalter:
@pysan3  https://asciinema.org/a/TRr73ID4EOsEQ38IQJfquGV9v

This video shows that I opened neo-tree with `:Neotree`. That is the only neo-tree command issued. After that I was only navigating with `h`,`j`,`k`,`l` and between windows with `<C-w>h` and `<C-w>l`.

Also No worries at all! I am sorry it is hard to explain, especially because I'm also not 100% certain on my own understanding of the issue 😅 

It also might be important to note that this is the commit that introduces the race condition: https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1288. Before this, there is no issue at all.

pysan3:
Thank you guys that is awesome @XavierChanth @mehalter 

@XavierChanth 
However I can't see your cursor in asciinema and I can't quite follow what you are doing. Could you give me instructions (keybinds)?

@mehalter 
Ok I was able to reproduce, but only when `config.filesystem.follow_current_file.enabled = true` which is not in repro.lua in the old comment. Am I right?

mehalter:
@pysan3 I can reproduce this without the `follow_current_file` enabled: https://asciinema.org/a/wECW01iEk1zmnZHlNf7tJX1rM

<details>
<summary>repro.lua</summary>
```lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  -- add any other plugins here
}

local neotree_config = {
  dependencies = { "MunifTanjim/nui.nvim", "nvim-tree/nvim-web-devicons", "nvim-lua/plenary.nvim" },
  cmd = { "Neotree" },
  keys = {
  },
  opts = {
    log_level = "trace",
    log_to_file = "/tmp/neotree-repro.log",
    filesystem = {
      filtered_items = {
        visible = true,
        show_hidden_count = true,
        hide_dotfiles = false,
        hide_gitignored = true,
        hide_by_name = {
          -- '.git',
          -- '.DS_Store',
          -- 'thumbs.db',
        },
        never_show = {},
      },
      window = {
        mappings = {
          h = function(state)
            local node = state.tree:get_node()
            if (node.type == "directory" or node:has_children()) and node:is_expanded() then
              state.commands.toggle_node(state)
            else
              require("neo-tree.ui.renderer").focus_node(state, node:get_parent_id())
            end
          end,
          l = function(state)
            local node = state.tree:get_node()
            if node.type == "directory" or node:has_children() then
              if not node:is_expanded() then -- if unexpanded, expand
                state.commands.toggle_node(state)
              else -- if expanded and has children, seleect the next child
                require("neo-tree.ui.renderer").focus_node(state, node:get_child_ids()[1])
              end
            else -- if not a directory just open it
              state.commands.open(state)
            end
          end,
        },
      },
    },
    window = {
      position = "right",
    },
  },
}

  root = root .. "/plugins",
})

-- add anything else here
```

</details>

mehalter:
@pysan3 here is a replication using purely just the `repro.lua` given by csiekel in this comment: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1310#issuecomment-1895833654

https://asciinema.org/a/GHLL4T2FKwq4MJhGyYpmf72JK

So this doesn't even use changes in my implementation of the `l` command.

mehalter:
@georgeguimaraes while making these recordings I did find that I can replicate this issue 100% of the time actually!

Steps (with repro.lua from here: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1310#issuecomment-1895833654):

1. `nvim` - open Neovim
2. `:Neotree` - open neo-tree
3. Open a directory with `l`
4. `<C-w>h` - switch to the other window
5. `<C-w>l` - switch back into the neo-tree sidebar
6. Navigate to another closed directory and run `l`
7. See that the cursor jumps to the wrong location

https://asciinema.org/a/PGmF1DoW1X36FVbz1Ps0ATNzB

mehalter:
Also worth noting the above steps reproduce the bug even with this PR

pysan3:
@mehalter Oh you are right. I'm now on the same page.

And now I understand why you added `save(..., overwritable)` and that it doesn't solve the issue kekw.

OK, let me try to tackle the problem ;)

pysan3:
Hey @mehalter, could you test my fix? I think this should solve the issue.

<https://github.com/pysan3/neo-tree.nvim/tree/update-curpos-before-expand>

```lua
{
  "pysan3/neo-tree.nvim",
  branch = "update-curpos-before-expand",
  version = false,
```

pysan3:
I had to re-add `BufDelete` aucmd since user must force close neo-tree with `:q` or `:bd`.

Since the block was added to not overwrite saved position, we now need to be exactly sure that previous position must be restored and set to nil before when we really want to save (otherwise, there will be a position remaining when you really want to save the position).

Man it's pretty difficult to explain in words...
(that's why I'm on a rewrite to make the code a lot cleaner)

pysan3:
I'll wait for either @georgeguimaraes or @XavierChanth or any other person that previously had this issue, and if you give me a go, I'll make a PR.

You know that I didn't have this issue before so I can't test it with my workflow / config options lol.

XavierChanth:
> However I can't see your cursor in asciinema and I can't quite follow what you are doing. Could you give me instructions (keybinds)?

Sorry, I didn't realize that was the case. I'll test your branch right now, and if I still have this issue then I'll rerecord and explain.

XavierChanth:
> I'll wait for either @georgeguimaraes or @XavierChanth or any other person that previously had this issue, and if you give me a go, I'll make a PR.

Seems to be less reproducible, but still happening...

https://asciinema.org/a/idz4OwuEX6TteaZK7NkpfSxjZ

My cursor only seems to jump when I toggle open a folder. In both of my recordings, I'm using <leader>+e to toggle neotree opened / closed.

Im using j/k to navigate vertically in neotree, and I'm using <CR> to  toggle the directories opened/closed.

It doesn't seem to happen when I close directories, only open them.
In the recording, you see me often toggle directories closed, then toggle neotree open and closed. For me, the issue is only reproducible right after toggling neotree open, and it seems to only jump on the first directory toggled open after opening neotree (so I have to close directories and neotree to reset again).

pysan3:
Thanks mate!

Ok let me dig into the problem. It'll probably be in the weekends and I'll make a PR when I'm done.

cseickel:
Closing because I think this was fixed by #1355, let me know if that's wrong.

PR #1355, opened 2024-02-16
fix(renderer): clean out edge cases of cursor.position.restore

### LINKED ISSUE #1310: BUG: jump to selected item when opening folder

pysan3:
We need to be more careful when to save and restore the cursor position as we cannot overwrite the position anymore with https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1288.

closes #1310

Please read the LONG discussion from https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1354#issuecomment-1946307225 as it describes the issue and the reason thoroughly.

I will keep this as draft until I get some people to test this code as I might have missed some more edge cases.

Would you mind testing this code? @georgeguimaraes @mehalter @XavierChanth

```lua
{ 
  "pysan3/neo-tree.nvim",
  branch = "update-curpos-before-expand",
  version = false,
```

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -699,7 +699,7 @@ M.position = {
 M.redraw = function(state)
   if state.tree and M.tree_is_visible(state) then
     log.trace("Redrawing tree", state.name, state.id)
-    -- every now and then this will fail because the window was closed in 
+    -- every now and then this will fail because the window was closed in
     -- betweeen the start of an async refresh and the redraw call.
     -- This is not a problem, so we just ignore the error.
     local success = pcall(render_tree, state)
@@ -938,6 +938,9 @@ local get_buffer = function(bufname, state)
     vim.api.nvim_buf_set_option(bufnr, "filetype", "neo-tree")
     vim.api.nvim_buf_set_option(bufnr, "modifiable", false)
     vim.api.nvim_buf_set_option(bufnr, "undolevels", -1)
+    autocmd.buf.define(bufnr, "BufDelete", function()
+      M.position.save(state)
+    end)
   end
   return bufnr
 end
@@ -1036,6 +1039,10 @@ M.acquire_window = function(state)
     vim.api.nvim_buf_set_var(state.bufnr, "neo_tree_tabid", state.tabid)
     vim.api.nvim_buf_set_var(state.bufnr, "neo_tree_position", state.current_position)
     vim.api.nvim_buf_set_var(state.bufnr, "neo_tree_winid", state.winid)
+    -- Used to track the position of the cursor within the tree as it gains and loses focus
+    win:on({ "BufDelete" }, function()
+      M.position.save(state)
+    end)
   end
 
   if win ~= nil then
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -522,8 +522,10 @@ M.focus_node = function(state, id, do_not_focus_window, relative_movement, botto
     end
     local success, err = pcall(vim.api.nvim_win_set_cursor, state.winid, { linenr, col })
 
-    -- now ensure that the window is scrolled correctly
     if success then
+      -- forget about cursor position as it is overwritten
+      M.position.clear(state)
+      -- now ensure that the window is scrolled correctly
       local execute_win_command = function(cmd)
         if vim.api.nvim_get_current_win() == state.winid then
           vim.cmd(cmd)
@@ -667,6 +669,16 @@ M.position = {
     state.position.node_id = node_id
     state.position.is.restorable = true
   end,
+  clear = function (state)
+    log.debug("Forget about cursor position.")
+    -- Clear saved position, so that we can save another position later.
+    state.position.topline = nil
+    state.position.lnum = nil
+    -- After focusing a node, we clear it so that subsequent renderer.position.restore don't
+    -- focus on it anymore
+    state.position.node_id = nil
+    state.position.is.restorable = false
+  end,
   restore = function(state)
     if state.position.is.restorable then
       if state.position.topline and state.position.lnum then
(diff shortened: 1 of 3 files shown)


mehalter:
I did some testing and this does seem to fix the linked issue! ❤️

XavierChanth:
Working wonderfully @pysan3 ! Thanks so much

pysan3:
Thanks @mehalter @XavierChanth !

Ready to roll @cseickel

lua/neo-tree/ui/renderer.lua
```diff
@@ -522,8 +522,10 @@ M.focus_node = function(state, id, do_not_focus_window, relative_movement, botto
     end
     local success, err = pcall(vim.api.nvim_win_set_cursor, state.winid, { linenr, col })
 
-    -- now ensure that the window is scrolled correctly
     if success then
+      -- forget about cursor position as it is overwritten
+      M.position.clear(state)
```
cseickel, line 527: Couldn't we have just set 
```suggestion
      M.position.is.restorable = false
```
 at this location and accomplish the same thing without adding the clear method? Are we duplicating functionality by have both?

lua/neo-tree/ui/renderer.lua
```diff
@@ -938,7 +944,7 @@ local get_buffer = function(bufname, state)
     vim.api.nvim_buf_set_option(bufnr, "filetype", "neo-tree")
     vim.api.nvim_buf_set_option(bufnr, "modifiable", false)
     vim.api.nvim_buf_set_option(bufnr, "undolevels", -1)
-    autocmd.buf.define(bufnr, "WinLeave", function()
+    autocmd.buf.define(bufnr, "BufDelete", function()
```
cseickel, line 947: Why was this changed? Does it still work if the buffer is not being displayed in a window?

lua/neo-tree/ui/renderer.lua
pysan3: No, we need to make `topline` or `lnum` falsy to skip the first line in save function.
Due to how the code is structured, the oldest save is most respected now. As I've written in the comment, when cursor position is overwritten, we should forget about this so that a later call to save can actually save its value. Otherwise, if the position is still saved, then the next call will not overwrite since oldest is greatest.

Additionally, I thought these `state.position.*` variables as private properties, and didn't like the idea to modify them from outside the scope.
That's why I made a public facing method to change these internal properties. Just like getters and setters in other languages.

lua/neo-tree/ui/renderer.lua
pysan3: When the buffer is not displayed in a window, users cannot change the position of the cursor right?

And for places that changes the cursor programatically (like follow current file), those cursor movements are captured in the _correct_ place (i.e. in `render_tree` or `close` etc).

Fun fact, that WinLeave _is_ the direct cause of the bug @mehalter was talking about in the issue.

lua/neo-tree/ui/renderer.lua
cseickel: I personally prefer the clear method to the `is.restorable` property, but it seems like we should not have both. Can we get rid of the `is.restorable` or is it handling another situation?

lua/neo-tree/ui/renderer.lua
cseickel: Makes sense.

lua/neo-tree/ui/renderer.lua
pysan3: As the comment says.

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/db178f4a49c19f8e4ed5a01dafa9d79e76f0081e/lua/neo-tree/ui/renderer.lua#L657-L659

This variable is more to decide whether we should restore the cursor position in the next iteration.
What `.clear()` does is more of a `is.savable + is.setable` (`save()` AND `set()` will do the thing next time), which stores if the next call to save should overwrite the previous saved `topline / lnum`.

But I don't like the name _savable_, like looking at the variable name, I don't quite get what the variable means at glance, or I have to read thru all the discussions that's held now to get the context.
Whereas `clear()` (to me) sounds like it forgets about the current cursor position and that's what it's literally doing.

cseickel: I still don't see how setting `is.restorable = false` has a different outcome from using `clear()` in that situation and removing all references to `is.restorable`. If there is a subtle difference, then the system is too complicated.

pysan3: Do you mean to depreciate `is.restorable` and use clear on every place? Then that's no problem, clear is a superset of that.

cseickel: Yes, but I don't think there is no need to deprecate as this is all completely internal and shouldn't be used by end users. I searched and no external source I know of use it either.

pysan3: Ok, I'll delete all occurrences of is.restorable and replace it with clear. 

Plz wait for a couple of days due to real life issues.

pysan3: Done @cseickel ;)

cseickel (approved):

Issue #1371, opened 2024-03-02
FEATURE: Put cursor in the beginning of line in normal mode


luthfielaroeha:
### Did you check the docs?

- [X] I have read all the docs.

### Is your feature request related to a problem? Please describe.

Most of the time, when renaming a file I just want to rename the filename without changing the extension. Currently the cursor placed in the end of line.

### Describe the solution you'd like.

I think it would be better to put the cursor in the beginning of the line so we can just use `cw` or `ct.` to change the filename.

pysan3:
https://github.com/nvim-neo-tree/neo-tree.nvim/issues/602

luthfielaroeha:
Sorry probably I'm missing something.
I'm aware that the normal mode exists, but when I want to rename the file, the cursor started in the end of the line.
Is there any other config to make the cursor in the beginning of the line instead?

<img width="261" alt="image" src="https://github.com/nvim-neo-tree/neo-tree.nvim/assets/11254971/4ce2d962-4d43-4b5e-9f6b-1615e632f995">

pysan3:
Oh I'm sorry.

bwpge:
Just adding info for anyone looking to pick this up: you might have success getting the `nui` input/popup window ID ([`split.winid`](https://github.com/MunifTanjim/nui.nvim/tree/main/lua/nui/popup#popup)) and setting the cursor with [`vim.api.nvim_win_set_cursor()`](https://neovim.io/doc/user/api.html#nvim_win_set_cursor()).

pysan3:
`neo_tree_popup_input_ready` event has been added with https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1372.

Please read `:h neo-tree-events` for more information.

uthmanmoh:
Not sure if this is the right place to ask this, but does anyone know if there's a way to control what happens based on the different input popups?
I want to be able to customize what happens when the input box for the `move` operation opens, and a different action when the input box for `add` opens

uthmanmoh:
Also with this new suggested event:
```lua
{
  event = "neo_tree_popup_input_ready",
  ---@param input NuiInput
  handler = function(input)
    vim.cmd("stopinsert") -- Enter normal mode
  end,
}
```
[...]

PR #1372, opened 2024-03-02
feat(nui): allow callbacks in nui input with option

pysan3:
We already have an option to default to normal mode in nui inputs.
`config.enable_normal_mode_for_inputs`

However, it looks like users want more power of customization to the details. (https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1371)

So why not just let users do whatever they want with absolute power?

All `function/string/table` are considered true-thy so this will cover backwards compatibility.

### DIFF
--- a/lua/neo-tree/ui/inputs.lua
+++ b/lua/neo-tree/ui/inputs.lua
@@ -2,6 +2,7 @@ local vim = vim
 local Input = require("nui.input")
 local popups = require("neo-tree.ui.popups")
 local utils = require("neo-tree.utils")
+local events = require("neo-tree.events")
 
 local M = {}
 
@@ -14,16 +15,13 @@ M.show_input = function(input, callback)
   local config = require("neo-tree").config
   input:mount()
 
-  if config.enable_normal_mode_for_inputs and input.prompt_type ~= "confirm" then
+  if input.prompt_type ~= "confirm" then
     vim.schedule(function()
-      vim.cmd("stopinsert")
-      if type(config.enable_normal_mode_for_inputs) == "function" then
-        config.enable_normal_mode_for_inputs(input)
-      elseif type(config.enable_normal_mode_for_inputs) == "string" then
-        vim.cmd(config.enable_normal_mode_for_inputs)
-      elseif type(config.enable_normal_mode_for_inputs) == "table" then
-        vim.cmd(config.enable_normal_mode_for_inputs)
+      -- deprecate this option in next version
+      if config.enable_normal_mode_for_inputs then
+        vim.cmd("stopinsert")
       end
+      events.fire_event(events.NEO_TREE_POPUP_INPUT_READY)
     end)
   end
 
--- a/lua/neo-tree/ui/inputs.lua
+++ b/lua/neo-tree/ui/inputs.lua
@@ -17,6 +17,13 @@ M.show_input = function(input, callback)
   if config.enable_normal_mode_for_inputs and input.prompt_type ~= "confirm" then
     vim.schedule(function()
       vim.cmd("stopinsert")
+      if type(config.enable_normal_mode_for_inputs) == "function" then
+        config.enable_normal_mode_for_inputs(input)
+      elseif type(config.enable_normal_mode_for_inputs) == "string" then
+        vim.cmd(config.enable_normal_mode_for_inputs)
+      elseif type(config.enable_normal_mode_for_inputs) == "table" then
+        vim.cmd(config.enable_normal_mode_for_inputs)
+      end
     end)
   end
 
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -1431,6 +1431,21 @@ have.
 "neo_tree_popup_buffer_leave"~
 Fired after leaving a neo-tree popup buffer.
 
+"neo_tree_popup_input_ready"~
+Fired after NuiInput is ready. Use this event to default to normal mode etc.
+This is fired inside a vim.schedule.
+
+>lua
+    {
+      event = "neo_tree_popup_input_ready",
+      ---@param input NuiInput
+      handler = function(input)
+        -- enter input popup with normal mode by default.
+        vim.cmd("stopinsert")
+      end,
+    }
+<
+
 "neo_tree_window_before_open"~
 Fired before opening a new Neo-tree window. Called with the following arg:
                                                      *neo-tree-window-event-args*
--- a/lua/neo-tree/events/init.lua
+++ b/lua/neo-tree/events/init.lua
@@ -24,6 +24,7 @@ local M = {
   NEO_TREE_LSP_UPDATE = "neo_tree_lsp_update",
   NEO_TREE_POPUP_BUFFER_ENTER = "neo_tree_popup_buffer_enter",
   NEO_TREE_POPUP_BUFFER_LEAVE = "neo_tree_popup_buffer_leave",
+  NEO_TREE_POPUP_INPUT_READY = "neo_tree_popup_input_ready",
   NEO_TREE_WINDOW_AFTER_CLOSE = "neo_tree_window_after_close",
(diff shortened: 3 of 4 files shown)


pysan3:
Usecase:

Move cursor right before the extension and start insert mode.

```lua
config.enable_normal_mode_for_inputs = ":exe 'norm F.' | :startinsert"
```

cseickel:
> Usecase:
> 
> Move cursor right before the extension and start insert mode.
> 
> ```lua
> config.enable_normal_mode_for_inputs = ":exe 'norm F.' | :startinsert"
> ```

😱 In this example you're using an option called enable_**normal**_mode_for_inputs to start**insert** mode. 😱 

I think we already have a generic solution that would be perfect, which is the event system. In a quick attempt I could not get the existing `NEO_TREE_POPUP_BUFFER_ENTER` event to work for this situation, but I think either modifying the timing of `NEO_TREE_POPUP_BUFFER_ENTER` or adding a new event which fires at the appropriate time is the correct way to create a generic solution. Maybe `NEO_TREE_POPUP_INPUT_READY` would be a good event name.

pysan3:
> > Usecase:
> > Move cursor right before the extension and start insert mode.
> > ```lua
> > config.enable_normal_mode_for_inputs = ":exe 'norm F.' | :startinsert"
> > ```
> 
> 😱 In this example you're using an option called enable_**normal**_mode_for_inputs to start**insert** mode. 😱

Yeah :tada: That's what I mean by _absolute power_ lol. I plan to not document this anywhere so it's absolutely to the users own risk to do anything.
 
> I think we already have a generic solution that would be perfect, which is the event system. In a quick attempt I could not get the existing `NEO_TREE_POPUP_BUFFER_ENTER` event to work for this situation, but I think either modifying the timing of `NEO_TREE_POPUP_BUFFER_ENTER` or adding a new event which fires at the appropriate time is the correct way to create a generic solution. Maybe `NEO_TREE_POPUP_INPUT_READY` would be a good event name.

`NEO_TREE_POPUP_BUFFER_ENTER` This won't work as `stopinsert` is called way after buffer creation or even `input:mount`, and it's inside a `vim.schedule` which will delay it even more. The only way would be to `defer_fn` a couple of ms but that's really not good...

Creating a new event is also a good option, but the codebase is the same for `confirm` (used with eg `delete (y/n)` or `change_cwd`) and `input` (like `add`, `copy` etc).
I don't want to run the same event for those cases..

lua/neo-tree/ui/inputs.lua
```diff
@@ -17,6 +17,13 @@ M.show_input = function(input, callback)
   if config.enable_normal_mode_for_inputs and input.prompt_type ~= "confirm" then
```
cseickel, line 17: I think what we can do is to change this to:

```lua
if input.prompt_type ~= "confirm" then
  vim.schedule(function()
    if config.enable_normal_mode_for_inputs then
      vim.cmd("stopinsert")
    end
    events.fire_event(events.NEO_TREE_POPUP_INPUT_READY)
  end)
end
```

Then people can use the NEO_TREE_POPUP_INPUT_READY event for custom actions without altering any existing configuration or introducing a hack we don't even want to document.

pysan3: If we add this, what's the point of enable_normal_mode option now?

cseickel: Backwards compatibility. I would be in favor of deprecating it.

pysan3: Okay. I'll fix that tomorrow.

pysan3:
@cseickel Updated.

Regarding this change, I deprecated the `enable_normal_mode_for_inputs` option and added it to migrations.

As I wanted to give more information on how to migrate, I changed the migration module to be able to add descriptions in the buffer.

Besides, now bullet point for each change doesn't make much sense, so I changed it to use level 2 header `## Header 2` instead for each value.

image

cseickel:
It is definitely nice to be able to add a longer description to deprecations. 

One thing I was considering is that I think it is more standard for plugins to have a `start_in_insert_mode` configuration for these situations. Maybe we should do the same and change the config option to `start_inputs_in_insert_mode = false` instead? Then we can keep the event for more advanced configuration but also have the simple one which is pretty standard.

What do you think?

pysan3:
I don't quite like when behavior before event handler execution changes by a different option in the config.

If you have an option to alter the behavior other than `NEO_TREE_POPUP_INPUT_READY`, I'd need to mention that in the docs with if ... otherwise ...

When people asks how to do xxx in the popup, I need to instruct; if you have `config.start_inputs_in_insert_mode` to true, then set this handler, and if false, set this handler.

I'd prefer to have the handler start consistently, either always from insert mode or always from normal mode.

cseickel:
> When people asks how to do xxx in the popup, I need to instruct; if you have `config.start_inputs_in_insert_mode` to true, then set this handler, and if false, set this handler.

OK, fair enough. In that case, we stick with your idea.

cseickel (approved):
Sorry for the delay on this one. LGTM.

Sologala:
@pysan3 Hello,   I update to release `3.19`. The prompted `NeoTree migrations` reports some Error

```shell
E5108: Error executing lua ...m/lazy/neo-tree.nvim/lua/neo-tree/setup/deprecations.lua:18: Invalid option name: 'wrap'
stack traceback:
        [C]: in function 'nvim_buf_set_option'
        ...m/lazy/neo-tree.nvim/lua/neo-tree/setup/deprecations.lua:18: in function 'show_migrations'
        ...re/nvim/lazy/neo-tree.nvim/lua/neo-tree/command/init.lua:51: in function 'execute'
        ...re/nvim/lazy/neo-tree.nvim/lua/neo-tree/command/init.lua:169: in function '_command'
        [string ":lua"]:1: in main chunk

```
Do you know how to fix it or find more clues?

pysan3:
@Sologala 
I'm so sorry: https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1387

Please wait for the next release (which should happen right away).

You should see something like this so please follow the instructions.
image

cseickel:
The fix was just released as `3.20`.

Sologala:
@pysan3  @cseickel Thx for update, My neo-tree work delicately again!

rrrix:
@pysan3 I think you forgot to remove `enable_normal_mode_for_inputs` from `defaults.lua`. I'm still getting a deprecation/migration notification on every startup, despite no use of this key in any configuration file. 

The _only_ location I can find where `enable_normal_mode_for_inputs` is being set is in `defaults.lua`:

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/d8b4687c2457465199bb36be02316e0d3066605d/lua/neo-tree/defaults.lua#L22


I believe this is because the `removed(key, desc)` checks for `nil`, and `false` is not `nil`.

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/d8b4687c2457465199bb36be02316e0d3066605d/lua/neo-tree/setup/deprecations.lua#L61-L68

```lua
$ lua
Lua 5.4.6  Copyright (C) 1994-2023 Lua.org, PUC-Rio
> foo = false
> foo ~= nil
true
> foo = nil
> foo ~= nil
false
```

Commenting out `enable_normal_mode_for_inputs` fixes the erroneous migration notification.

pysan3:
My bad. Thank you @rrrix

davawen:
I may be missing a basic command, but can I make it such that enter the popup in insert mode, but pressing escape puts me in normal mode (instead of leaving the popup) ?

That's the behavior I'm getting from other popups.

pysan3:
@davawen 

Map `"i", "<Esc>"` to `vim.cmd.stopinsert` in the event. There's the exact example in `:h neo-tree-events` for you but you need to update to the latest main branch (not yet include in the latest release iirc).

Issue #1382, opened 2024-03-08
BUG: opening up a file via Neotree results in gopls not honoring the package environment (on Windows)


burgr033:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

v0.9.5

### Operating System / Version

Windows 11

### Describe the Bug

When I open up a go file through neotree on windows, gopls (the go language server) doesn't work properly because it thinks it's not inside the package environment.

This only happens on Windows. I tried reproducing it on Arch and Debian.

### Screenshots, Traceback

![ezgif-1-4543ac5d26](https://github.com/nvim-neo-tree/neo-tree.nvim/assets/17072527/7fd1375a-ab78-4fd0-ba7f-2ea59f1b8aaf)

### Steps to Reproduce

1. `go mod init test123` -- init a new go project inside the repro enviornment.
2. create a main.go file with 
```golang
package main
func main(){}
```
4. `nvim -u init.lua` -- open up neovim
5. `:Neotree` -- open up Neotree
6. select the file via Neotree and press return key

file should open now with a package warning through gopls along the lines of
`No packages found for open file C:\\_workspace\temp\nvim_test\main.go`

### Expected Behavior

File should be opened and the package environment should be recognized

(If you open the file without Neotree e.g. with netrw or via command line; everything is working as expected)

### Your Configuration

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
        vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
        vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
        "folke/tokyonight.nvim",
        "neovim/nvim-lspconfig",
        "williamboman/mason.nvim",
        "williamboman/mason-lspconfig.nvim",
        "neovim/nvim-lspconfig",
        -- add any other plugins here
}

local neotree_config = {
[...]

PR #1388, opened 2024-03-09
fix: continue improving Windows path escaping for commands

bwpge:
This PR addresses more edge cases with path escaping and vim commands, as noted in #1382.

After some testing, I found that only a small number of punctuation characters required the extra escape after  `vim.fn.fnameescape`. They are: `&`, `(`, `)`, `;`, `^`, and `` ` ``.

<details>
<summary>Details</summary>

Testing method used was running `:luafile %` with the following `test.lua` on a Windows machine:

```lua
-- copy/paste from neo-tree
local is_windows = vim.fn.has("win32") == 1 or vim.fn.has("win32unix") == 1

local escape_path_for_cmd = function(path)
  local escaped_path = vim.fn.fnameescape(path)
  if is_windows then
    for _, c in ipairs({ --[[ add characters as cases fail ]] }) do
      escaped_path = escaped_path:gsub("\\%" .. c, "\\%1")
    end
  end
  return escaped_path
end
-- end copy/paste

vim.fn.mkdir("tests", "p")

-- known allowed NTFS punctuation
local symbols = [[!#$%&'()+,-;=@[]^_`{}~ ]]

-- write file with a leading punctuation to a nested directory
for char in string.gmatch(symbols, ".") do
  -- double the punctuation to ensure it actually gets escaped correctly
  local f = string.format([[.\tests\%sfoo.txt]], char .. char)
  local p = escape_path_for_cmd(f)
  vim.cmd("silent w " .. p)
end

-- files outside of `tests` directory didn't get escaped correctly
for _, f in ipairs(vim.fn.glob("tests*.txt", false, true)) do
  vim.notify("Failed: " .. f, vim.log.levels.ERROR)
  os.remove(f)
end

-- these files were written to the correct path
for _, f in ipairs(vim.fn.glob("tests/*.txt", false, true)) do
  os.remove(f)
end
```

This testing logic could certainly be improved, I just quickly slapped this together.
</details>

The only reason I included the above is because it may be worth creating Windows path tests in the future and didn't want to lose this information.

I manually verified this fixes the issue in #1382, and tried to recreate some issues I've reported/worked on (such as #1352) and did not see any regressions. That being said, I would feel a bit better if someone could get a second pair of eyes on a Windows machine with this PR.

### DIFF
--- a/lua/neo-tree/utils/init.lua
+++ b/lua/neo-tree/utils/init.lua
@@ -1018,11 +1018,19 @@ end
 M.escape_path_for_cmd = function(path)
   local escaped_path = vim.fn.fnameescape(path)
   if M.is_windows then
-    -- on windows, any punctuation preceeded by a `\` needs to have a second `\`
-    -- added to preserve the path separator. this is a naive replacement and
+    -- on windows, some punctuation preceeded by a `\` needs to have a second
+    -- `\` added to preserve the path separator. this is a naive replacement and
     -- definitely not bullet proof. if we start finding issues with opening files
-    -- or changing directories, look here first.
-    escaped_path = escaped_path:gsub("\\%p", "\\%1")
+    -- or changing directories, look here first. #1382 was the first regression
+    -- from the implementation that used lua's %p to match punctuation, which
+    -- did not quite work. the following characters were tested on windows to
+    -- be known to require an extra escape character.
+    for _, c in ipairs({ "&", "(", ")", ";", "^", "`" }) do
+      -- lua doesn't seem to have a problem with an unnecessary `%` escape
+      -- (e.g., `%;`), so we can use it to ensure we match the punctuation
+      -- for the ones that do (e.g., `%(` and `%^`)
+      escaped_path = escaped_path:gsub("\\%" .. c, "\\%1")
+    end
   end
   return escaped_path
 end


lua/neo-tree/utils/init.lua
```diff
@@ -1018,11 +1018,19 @@ end
...
+    -- on windows, some punctuation preceeded by a `\` needs to have a second
+    -- `\` added to preserve the path separator. this is a naive replacement and
     -- definitely not bullet proof. if we start finding issues with opening files
-    -- or changing directories, look here first.
-    escaped_path = escaped_path:gsub("\\%p", "\\%1")
+    -- or changing directories, look here first. #1382 was the first regression
+    -- from the implementation that used lua's %p to match punctuation, which
+    -- did not quite work. the following characters were tested on windows to
+    -- be known to require an extra escape character.
+    for _, c in ipairs({ "&", "(", ")", ";", "^", "`" }) do
```
bwpge, line 1028: Alternatively, we could just use a table with pre-escaped patterns like:

```lua
for _, c in ipairs({ "&", "%(", "%)", ";", "%^", "`" }) do
```

And then do the substitution with:

```lua
escaped_path = escaped_path:gsub("\\" .. c, "\\%1")
```

I'm just not intimately familiar with lua patterns (e.g., [magic characters](https://www.lua.org/pil/20.2.html)) so I used a what I thought might be the "safest" approach. If someone puts a `p` (or any other character class like `a`, `d`, etc.) in this table we'll have problems though... :)

burgr033:
@bwpge asked me in #1382  to pull their fork and see if I find anything odd. I worked with their fork for half an hour and couldn't really find anything out of the ordinary.

I even checked with a few other characters.

```
_workspace
;workspace
(workspace
)workspace
&workspace
%workspace
`workspace
´workspace
^workspace
~workspace
```

Everything that I could check, works fine :)

lua/neo-tree/utils/init.lua
cseickel: I think it reads clearer the way you have it now.

cseickel (approved):
Thanks for working on this!

pysan3:
Thanks for working on this @bwpge !

Amazing work!

bwpge:
Happy to help! And just a reminder to @burgr033 to change your config back, thanks for the assist :)

burgr033:
thank you all <3

pysan3:
@bwpge 

> @bwpge Pinging you again :)
> 
> We needed `" "` (check for space) at the end of the list in #1388. Right?
>
> From https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1400#pullrequestreview-1939711304

Issue #833, opened 2023-03-27
bug: toggling a parent folder when using `group_empty_dirs` causes it to be wrongly coalesced


oddlama:
Hard to explain purely in written form, here is a step by step guide to reproduce this:

1. Start with some common nested folder structure while enabling `group_empty_dirs` in neotree. Open a folder in a folder.
   ![image](https://user-images.githubusercontent.com/31919558/228088887-ae126c57-de4f-4dd6-a67d-e411c9d1e7ad.png)
2. Press <kbd>Enter</kbd> on the topmost folder to close it temporarily and refresh <kbd>R</kbd>. All folders that were once opened will be re-opened (shouldn't happen) and all folders with a nesting >=2 will be coalesced wrongly, hiding all members of the folders in the middle
    ![image](https://user-images.githubusercontent.com/31919558/228089102-2504b717-cb85-453d-b8ac-b04399ca749a.png)
3. Close topmost folder, move cursor away and refresh again as a temporary fix

tincotema:
Hi, is there some update to the state of this issue? I encountered it too ...

PR #1393, opened 2024-03-12
fix(fs_scan): ensure parents of all expanded nodes are also scanned

pysan3:
On refresh a list of expanded nodes is gathered to decide what paths to scan. However, when an expanded node lives inside a closed parent node, parent node's number of children will be wrong.
This results in unexpected behavior with `group_with_dirs`, so this fix ensures that all parent nodes are scanned.

For more detailed explanation with screenshots please read: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/833.

Fixes: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/833


cseickel (approved):
Sorry I forgot about this one...

PR #1395, opened 2024-03-13
Delete `enable_normal_mode_for_inputs` from `defaults.lua`

rrrix:
Fixes the erroneous ``Some options have changed, please run `:Neotree migrations` to see the changes`` notification when neo-tree is opened. 

Follow up for bug introduced in PR #1372

See https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1372#issuecomment-1995921030 for discussion.


cseickel (approved):
Thanks!

PR #1397, opened 2024-03-15
fix(docs): use proper lua syntax & vim.keymap api

pheis:
Noticed mistake on docs when reading them today :)

Thanks for this awesome plugin and have a nice weekend for the maintainers!


cseickel (approved):
Thanks for the fix!

Issue #1396, opened 2024-03-15
Do not close popup with <Esc> in INSERT mode


cseickel:
Also with this new suggested event:
```lua
{
  event = "neo_tree_popup_input_ready",
  ---@param input NuiInput
  handler = function(input)
    vim.cmd("stopinsert") -- Enter normal mode
  end,
}
```
I am unable to enter normal mode after manually entering insert mode with `i`, since pressing esc even in insert mode just closes the input. Is there a way to make it so that I need to press `esc` from within normal mode only to exit the input box, as I believe it used to be before?

_Originally posted by @uthmanmoh in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1371#issuecomment-1998578046_

pysan3:
@uthmanmoh I've updated the help page and it should be more helpful for you this time.

And of course you can combine the two into one handler (tbh I recommend registering only one handler for this event).

```lua
    {
      event = "neo_tree_popup_input_ready",
      handler = function()
        -- enter input popup with normal mode by default.
        vim.cmd("stopinsert")
      end,
    },
    {
      event = "neo_tree_popup_input_ready",
      ---@param args { bufnr: integer, winid: integer }
      handler = function(args)
        -- map <esc> to enter normal mode (by default closes prompt)
        -- don't forget `opts.buffer` to specify the buffer of the popup.
        vim.keymap.set("i", "<esc>", vim.cmd.stopinsert, { noremap = true, buffer = args.bufnr }) 
      end,
    }
```

BTW if you have something like `nnoremap jk <esc>` globally, you don't need to remap `<esc>` for the popup but use `jk` to get out of insert mode OOTB.

PR #1398, opened 2024-03-15
fix(input): pass bufnr and winid to event instead

pysan3:
Closes: #1396

> [!WARNING]
> 
> Breaking change.
> I've deprecated `config.enable_normal_mode_for_inputs` entirely and it does absolutely nothing now.

I've also added examples to the help page.

```lua
    {
      event = "neo_tree_popup_input_ready",
      ---@param args { input: NuiInput }
      handler = function(args)
        -- enter input popup with normal mode by default.
        vim.cmd("stopinsert")
        -- map <esc> to enter normal mode (closes prompt by default)
        args.input:map("i", "<esc>", vim.cmd.stopinsert, { noremap = true })
        -- unmap <esc> in normal mode to do nothing (closes prompt by default) (`q` still works)
        args.input:map("n", "<esc>", function() end, { noremap = true })
      end,
    }
```

### DIFF
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -1440,11 +1440,13 @@ This is fired inside a vim.schedule.
       event = "neo_tree_popup_input_ready",
       ---@param args { input: NuiInput }
       handler = function(args)
-        -- enter input popup with normal mode by default.
+        -- Example Usages:
+
+        -- 1. enter input popup with normal mode by default.
         vim.cmd("stopinsert")
-        -- map <esc> to enter normal mode (closes prompt by default)
+        -- 2. map <esc> to enter normal mode (closes prompt by default)
         args.input:map("i", "<esc>", vim.cmd.stopinsert, { noremap = true })
-        -- unmap <esc> in normal mode to do nothing (closes prompt by default) (`q` still works)
+        -- 3. unmap <esc> in normal mode to do nothing (closes prompt by default) (`q` still works)
         args.input:map("n", "<esc>", function() end, { noremap = true })
       end,
     }
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -1438,16 +1438,18 @@ This is fired inside a vim.schedule.
 >lua
     {
       event = "neo_tree_popup_input_ready",
-      ---@param args { input: NuiInput }
-      handler = function(args)
-        -- Example Usages:
-
-        -- 1. enter input popup with normal mode by default.
+      handler = function()
+        -- enter input popup with normal mode by default.
         vim.cmd("stopinsert")
-        -- 2. map <esc> to enter normal mode (closes prompt by default)
-        args.input:map("i", "<esc>", vim.cmd.stopinsert, { noremap = true })
-        -- 3. unmap <esc> in normal mode to do nothing (closes prompt by default) (`q` still works)
-        args.input:map("n", "<esc>", function() end, { noremap = true })
+      end,
+    },
+    {
+      event = "neo_tree_popup_input_ready",
+      ---@param args { bufnr: integer, winid: integer }
+      handler = function(args)
+        -- map <esc> to enter normal mode (by default closes prompt)
+        -- don't forget `opts.buffer` to specify the buffer of the popup.
+        vim.keymap.set("i", "<esc>", vim.cmd.stopinsert, { noremap = true, buffer = args.bufnr })
       end,
     }
 <
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -1438,10 +1438,14 @@ This is fired inside a vim.schedule.
 >lua
     {
       event = "neo_tree_popup_input_ready",
-      ---@param input NuiInput
-      handler = function(input)
+      ---@param args { input: NuiInput }
+      handler = function(args)
         -- enter input popup with normal mode by default.
         vim.cmd("stopinsert")
+        -- map <esc> to enter normal mode (closes prompt by default)
+        args.input:map("i", "<esc>", vim.cmd.stopinsert, { noremap = true })
+        -- unmap <esc> in normal mode to do nothing (closes prompt by default) (`q` still works)
+        args.input:map("n", "<esc>", function() end, { noremap = true })
       end,
     }
 <
--- a/lua/neo-tree/setup/deprecations.lua
+++ b/lua/neo-tree/setup/deprecations.lua
@@ -118,8 +118,8 @@ See instructions in `:h neo-tree-events` for more details.
 event_handlers = {
   {
     event = "neo_tree_popup_input_ready",
-    ---@param input NuiInput
-    handler = function(input)
+    ---@param args { bufnr: integer, winid: integer }
+    handler = function(args)
       -- enter input popup with normal mode by default.
       vim.cmd("stopinsert")
     end,
--- a/lua/neo-tree/setup/deprecations.lua
+++ b/lua/neo-tree/setup/deprecations.lua
@@ -120,8 +120,8 @@ event_handlers = {
     event = "neo_tree_popup_input_ready",
     ---@param args { bufnr: integer, winid: integer }
     handler = function(args)
-      -- enter input popup with normal mode by default.
       vim.cmd("stopinsert")
+      vim.keymap.set("i", "<esc>", vim.cmd.stopinsert, { noremap = true, buffer = args.bufnr })
     end,
   }
 }
--- a/lua/neo-tree/ui/inputs.lua
+++ b/lua/neo-tree/ui/inputs.lua
@@ -39,7 +39,10 @@ M.show_input = function(input, callback)
 
   if input.prompt_type ~= "confirm" then
     vim.schedule(function()
-      events.fire_event(events.NEO_TREE_POPUP_INPUT_READY, { input = input })
+      events.fire_event(events.NEO_TREE_POPUP_INPUT_READY, {
+        bufnr = input.bufnr,
+        winid = input.winid,
+      })
     end)
   end
 end
--- a/lua/neo-tree/ui/inputs.lua
+++ b/lua/neo-tree/ui/inputs.lua
@@ -12,24 +12,11 @@ local should_use_popup_input = function()
 end
 
 M.show_input = function(input, callback)
-  local config = require("neo-tree").config
   input:mount()
-
-  if input.prompt_type ~= "confirm" then
-    vim.schedule(function()
-      -- deprecate this option in next version
-      if config.enable_normal_mode_for_inputs then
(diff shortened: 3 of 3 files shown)


uthmanmoh:
Why is esc being unmapped in normal mode? esc being used to exit in normal mode seems fine, just shouldn't be used to exit in insert mode. Just the first mapping may be good enough

pysan3:
No each line is an example of what you can do with the handler. You don't need to use all of them. Or even write your own map (eg `jk`) to escape and so on. I might have to make that more clear.

uthmanmoh:
Ahh that's perfect then, thanks for the clarification

cseickel:
> No each line is an example of what you can do with the handler. You don't need to use all of them.

I agree that its confusing as it's written. There should just be two completely separate examples each with it's own description. One for starting in normal modes and one for just allowing normal mode even though it still starts in insert mode.

pysan3:
No other event has even a single example nor a type definition and I kinda feel it's awkward to make such a long example for this single event. 

I might start to think to delete the whole example altogether to align with other events.

cseickel:
I have a few more thoughts for discussion:

Do we really want to pass the NuiInput just to enable mapping? I think it would be better for the user if we just used standard nvim buffer local mapping so that they can use familiar nvim syntax for changing those mappings.

BUT, if we do pass the NuiInout, we should include links to the Nui docs:
https://github.com/MunifTanjim/nui.nvim/tree/main/lua/nui/popup#popupmap
https://github.com/MunifTanjim/nui.nvim/tree/main/lua/nui/popup#popupunmap

And also the unnmap example should use the unmap method and not set the mapping to an empty fnction.

cseickel:
> No other event has even a single example nor a type definition and I kinda feel it's awkward to make such a long example for this single event.
> 
> I might start to think to delete the whole example altogether to align with other events.

No definitely not. This is pretty common functionality so it's worth having a longer doc. Also, it's not at all a good thing that the rest of the docs are sparse, it's a huge problem to be fixed. We should always endeavor for more docs.

pysan3:
TBH I'm not interested in this feature and never gonna use this so I don't care how much docs there will be. 

If you don't want to pass NuiInput, what do you think we should pass? Bufnr and winid?

cseickel:
> TBH I'm not interested in this feature and never gonna use this so I don't care how much docs there will be.
> 
> If you don't want to pass NuiInput, what do you think we should pass? Bufnr and winid?

Maybe, I don't think you really need to pass anything because the popup would be the active window/buffer when the event runs.

In the future we should be passing the type of window it is running in, such as `rename`, `move`, create`, etc.

pysan3:
> > If you don't want to pass NuiInput, what do you think we should pass? Bufnr and winid?
> 
> 
> 
> Maybe, but I don't think you really need to pass anything because the popup would be the active window/buffer when the event runs.
> 

This is not guaranteed as the event is called inside a vim.schedule. 

> 
> In the future we should be passing the type of window it is running in, such as `rename`, `move`, `create`, etc.

Are you sure? This will require a complete rewrite and does anyone have the motivation to do that?

pysan3:
> Do we really want to pass the NuiInput just to enable mapping? I think it would be better for the user if we just used standard nvim buffer local mapping so that they can use familiar nvim syntax for changing those mappings.
> 
> BUT, if we do pass the NuiInout, we should include links to the Nui docs: [MunifTanjim/nui.nvim@`main`/lua/nui/popup#popupmap](https://github.com/MunifTanjim/nui.nvim/tree/main/lua/nui/popup?rgh-link-date=2024-03-15T11%3A45%3A30Z#popupmap) [MunifTanjim/nui.nvim@`main`/lua/nui/popup#popupunmap](https://github.com/MunifTanjim/nui.nvim/tree/main/lua/nui/popup?rgh-link-date=2024-03-15T11%3A45%3A30Z#popupunmap)
> 
> And also the unnmap example should use the unmap method and not set the mapping to an empty fnction.

FYI `vim.keymap.set` and `input:map` share the exact same argument types and the only difference is that you don't need to pass `opts.buffer` (`{ buffer = <bufnr> }`) to the last argument.

cseickel:
>>Maybe, but I don't think you really need to pass anything because the popup would be the active window/buffer when the event runs.

>This is not guaranteed as the event is called inside a vim.schedule.

If that were true, then suggesting that they use the `stopinsert` command would not make sense because if we can't guarantee that the window has been focused yet.

>FYI vim.keymap.set and input:map share the exact same argument types and the only difference is that you don't need to pass opts.buffer ({ buffer = <bufnr> }) to the last argument.

Honestly, I still use vimscript for mappings and have never used the native lua API so I didn't realize this. Either way, for memory leak purposes I'd rather send two integers than a reference to a complex object with an unknown dependency graph.

> > In the future we should be passing the type of window it is running in, such as `rename`, `move`, `create`, etc.
> 
> Are you sure? This will require a complete rewrite and does anyone have the motivation to do that?

LOL, I think you have an itchy trigger finger on the whole "complete rewrite" thing. It's a simple thing to just pass one more argument when creating an input and then carry that through to the event. Like I said though, "in the future." I would be happy to do it when I get the time.

pysan3:
> > FYI vim.keymap.set and input:map share the exact same argument types and the only difference is that you don't need to pass opts.buffer ({ buffer =  }) to the last argument.
> 
> Honestly, I still use vimscript for mappings and have never used the native lua API so I didn't realize this. Either way, for memory leak purposes I'd rather send two integers than a reference to a complex object with an unknown dependency graph.

OK fix already pushed.

> > > In the future we should be passing the type of window it is running in, such as `rename`, `move`, `create`, etc.
> > 
> > 
> > Are you sure? This will require a complete rewrite and does anyone have the motivation to do that?
> 
> LOL, I think you have an itchy trigger finger on the whole "complete rewrite" thing. It's a simple thing to just pass one more argument when creating an input and then carry that through to the event. Like I said though, "in the future." I would be happy to do it when I get the time.

I hate when the fix is to add one more argument **that can be nil** to solve the problem. It creates more complexity and a lot of bugs related to wrong number of args / wrong type passed which is difficult to detect at first glance. You don't know how long it took me to debug this issue: https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1227.

cseickel:
> I hate when the fix is to add one more argument **that can be nil** to solve the problem. It creates more complexity and a lot of bugs related to wrong number of args / wrong type passed which is difficult to detect at first glance. You don't know how long it took me to debug this issue: #1227.

I would not make it `nil`able, it should be required when the argument is added.

pysan3:
I was talking about the other nil-able arguments that come before type. It will be something in the line of `message, default_value, nil, nil, nil, "move"`

How many nils should I put in between? I dunno. Is there a test that validates before a PR? No. What if I typoed with `"nove"`? No penalty. 

Also to make it work with nio and to align with other nvim apis, callback should always be the last argument. 
This makes it impossible to add random arguments afterwards anyways since the number of arguments matters (which is a good this imo). 
I think we should use `opts` table to pass unnecessary (optional) arguments and put only non-nil-able arguments + boolean|nil flags as real arguments to a function. 

Anyways I'm not interested in changing ui/inputs (at least for now) and I think this is outside of the scope of this PR. 
I've already pushed the changes. Is there something else I need to change?

cseickel:
> Anyways I'm not interested in changing ui/inputs (at least for now) and I think this is outside of the scope of this PR.
> I've already pushed the changes. Is there something else I need to change?

Oh absolutely. Remember that I had said "in the future".

As far as args, the new argument would be at the beginning. I never put non-nil args after nilable args.

lua/neo-tree/setup/deprecations.lua
```diff
@@ -118,8 +118,8 @@ See instructions in `:h neo-tree-events` for more details.
 event_handlers = {
   {
     event = "neo_tree_popup_input_ready",
-    ---@param input NuiInput
-    handler = function(input)
+    ---@param args { bufnr: integer, winid: integer }
+    handler = function(args)
```
cseickel, line 122: The example here should perfectly replicate the old behavior, which mans it needs an `<esc>` mapping, right?

doc/neo-tree.txt
```diff
@@ -1438,11 +1438,19 @@ This is fired inside a vim.schedule.
 >lua
     {
       event = "neo_tree_popup_input_ready",
-      ---@param input NuiInput
-      handler = function(input)
+      handler = function()
         -- enter input popup with normal mode by default.
         vim.cmd("stopinsert")
```
cseickel, line 1443: This example should probably contain an override for the `<esc>` mapping too, shouldn't it? I don't think that anyone that uses normal mode for inputs would want `<esc>` to close the input from insert mode.

lua/neo-tree/setup/deprecations.lua
pysan3: Done.

doc/neo-tree.txt
pysan3: That's in the second event handler. I thought you said we should split the examples apart.

pysan3: https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1398#issuecomment-1999472063

cseickel: Yes, theree should be separate examples, but all of them requires changing the default `<esc>` mapping in a different way. Here are some examples:

1. Start in normal mode, change escape so that it only acts as the standard "back to normal mode" action
2. Start in insert mode, but allow escape to exit insert mode, and the next escape from normal mode to close the popup.
3. Start in insert mode, but allow escape to exit insert mode, use q to close the popup

It goes on and on, which is why we are going with the one standard "plus make your own." The issue is that in every scenario where you start in normal mode, you would want to change the default `<esc>` mapping because in some way so that it is possible to leave insert mode and go back to normal mode.

pysan3: If you have something like `nnoremap jk <Esc>`, you don't need `<Esc>` remapped to go to normal mode. In fact this is what I do; `jk` to go to normal mode, `<Esc>` to quit from within insert mode and ngl it's pretty convenient.

> The issue is that in every scenario where you start in normal mode, you would want to change the default <esc> mapping in some way so that it is possible to leave insert mode and go back to normal mode.

So I don't agree this is for every case.

cseickel: OK, fair enough. I guess that goes back to the point of why we are doing this in the first place, just too many variations.

cseickel (approved):

doc/neo-tree.txt
pysan3: > just too many variations.

99% of the issues we have comes down to this for real lol.

cseickel: Very true.

PR #1400, opened 2024-03-15
Update fs_actions.lua fixes #1399

### LINKED ISSUE #1399: Huge error in Windows Deletion command, deletes wrong dir.
if you're on Windows like this guy you know that when you make a copy of a dir, it will just append " - Copy" on the end, and that's a fine enough solution to that, but the problem is that passing a path to this line: 

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/120a83eb7c3c240d5298122111e9f65f0ee5bc8f/lua/neo-tree/sources/filesystem/lib/fs_actions.lua#L499

will delete the ORIGINAL DIR!

so trying to delete a dir  called "testDir - copy" will delete "testDir". 

cmd.exe treats spaces significantly. 

anyway, it seems fine when i changed the code to this: 

```lua        
local result = vim.fn.system({ "cmd.exe", "/c", "rmdir", "/s", "/q", '"', path, '"' })
```

WillEhrendreich:


lua/neo-tree/sources/filesystem/lib/fs_actions.lua
```diff
@@ -496,7 +496,7 @@ M.delete_node = function(path, callback, noconfirm)
       -- first try using native system commands, which are recursive
       local success = false
       if utils.is_windows then
-        local result = vim.fn.system({ "cmd.exe", "/c", "rmdir", "/s", "/q", path })
+        local result = vim.fn.system({ "cmd.exe", "/c", "rmdir", "/s", "/q", '"', path, '"' })
```
cseickel, line 499: Sorry about the bug and thanks for fixing, but I think there may be a better way to go about quoting. Isn't there a vim api to shellescape paths? My concern is that there may still be other file path issues we are not handling here and we should take this opportunity to address them all. 

@pysan3 what do you think?

pysan3: Dang it. I explicitly said we need to add check for `<Space>` on top of the punctuations but then completely forgot about that when reviewing the PR: https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1388

pysan3: @bwpge Pinging you again :)

We needed `" "` (check for space) at the end of the list in https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1388. Right?

pysan3: It's that `cmd.exe /C` treats all following arguments as single string and runs that command so it forgets that `path` is one argument.

bwpge: I'll definitely take a look when I get some time tonight.

To address @cseickel's point, the user can change their shell/shell escaping options (example [`:h shell-powershell`](https://neovim.io/doc/user/options.html#shell-powershell)), so if we're deliberately using `cmd` for Windows (which is a good approach for compatibility like if a user can't use powershell on corporate machines) we'll probably have to escape that ourselves.

And of course `cmd` plays by its own made up rules when it comes to escapes and quoting...

pysan3: Would it be less pain to not run cmd.exe when path contains space or puncts and fallback to recursively `vim.loop.fs_rmdir`?

pysan3: Using `vim.fn.shellescape(path)` might be better than `'"' .. path .. '"'` to say the least.

pysan3: After some amount of testing, I think `shellescape` does the job. I could not find any edge cases.

@WillEhrendreich Could you change your fix to use `vim.fn.shellescape(path)` instead? It would be nice if you could test your issue with this change.

@bwpge May I get a confirmation from you as well?

pysan3: LGTM @cseickel

cseickel (approved):
Thanks everyone for helping out with this.
@WillEhrendreich @pysan3 @bwpge

lua/neo-tree/sources/filesystem/lib/fs_actions.lua
bwpge: @pysan3 just a note, `vim.fn.shellescape` relies on the user's `shellslash` option (only applies to Windows). While I highly doubt in general people will mess with this setting, and it is off by default, it can create the following (difficult to troubleshoot) issue with a path like `foo/bar's` (note the use of single vs double quotes):

- With `shellslash` false: `"foo/bar's"` -> valid
- With `shellslash` true: `'foo/bar'\''s'` -> invalid

Obviously is a bit contrived, and is definitely better than what we had before, but is something to be aware of.

PR #1406, opened 2024-03-20
fix(renderer): check if current buffer is loaded

pysan3:
Somehow this commit has caused a bug with `:Neotree current toggle`. 

1. If you have more than one split, 
2. and you use `:Neotree current toggle`
3. then open a file
4. then `:Neotree current toggle`
5. you'll get an error (see below)
6. then your current window will be closed

Error from step 5:
```
Error executing vim.schedule lua callback: ...is/.local/share/nvim/lazy/nui.nvim/lua/nui/tree/init.lua:188: missing bufnr                                                     
stack traceback:                                                                                                                                                              
        [C]: in function 'error'                                                                                                                                              
        ...is/.local/share/nvim/lazy/nui.nvim/lua/nui/tree/init.lua:188: in function 'init'                                                                                   
        .../.local/share/nvim/lazy/nui.nvim/lua/nui/object/init.lua:132: in function 'NuiTree'                                                                                
        ...e/chris/repos/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:778: in function 'create_tree'                                                                            
        ...e/chris/repos/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:1190: in function 'draw'                                                                                  
        ...e/chris/repos/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:1326: in function 'show_nodes'                                                                            
        ...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:96: in function 'render_context'                                                                          
        ...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:165: in function <...ree.nvim/lua/neo-tree/sources/filesystem/lib/fs_scan.lua:164> 
```

@pysan3 I'm not sure I am going to be able to debug so I was thinking about reverting it for now. Do you want to look into it before I do?

_Originally posted by @cseickel in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1390#issuecomment-2008589757_

PR #1408, opened 2024-03-20
fix: winfixbuf/`current toggle` bug from #1390

cseickel:
This fixes the bug from #1390, but I have not tested it against `winfixbuf` yet. When you get a chance, can you review and test it? Thanks!

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -164,12 +164,15 @@ M.close = function(state, focus_prior_window)
     state.winid = nil
   end
   local bufnr = utils.get_value(state, "bufnr", 0, true)
-  state.bufnr = nil
-  vim.schedule(function()
-    if bufnr > 0 and vim.api.nvim_buf_is_valid(bufnr) then
-      vim.api.nvim_buf_delete(bufnr, { force = true })
+  if bufnr > 0 and vim.api.nvim_buf_is_valid(bufnr) then
+    state.bufnr = nil
+    local success, err = pcall(vim.api.nvim_buf_delete, bufnr, { force = true })
+    if not success and err:match("E523") then
+      vim.schedule_wrap(function()
+        vim.api.nvim_buf_delete(bufnr, { force = true })
+      end)()
     end
-  end)
+  end
   return window_existed
 end
 


pysan3:
And as long as we have this bug, `winfixbuf` won't work as well. @cseickel

pysan3:
BTW you've mentioned the wrong name lol. I have a _3_ at the end ;)

`@pysan3`

cseickel:
OK, we are going to have to explain why exactly this difference causes each bug. I'll try out your other fix later.

> BTW you've mentioned the wrong name lol. I have a _3_ at the end ;)

Oops 😊

pysan3:
Prior to winfixbuf, when neo-tree is floating, it was guaranteed that there is at least one ground (non-floating) window where we can send the `utils.open_file` or _stealing buffer_ (a buffer that was detected by the steal prevention mechanism).

However, when all ground windows are winfixbuf, we now need to make a new split before placing that buffer, just like the mechanism used for `left/right` neo-trees (when neo-tree is the only opened window).

This is why `float + vsplit` is now an official issue we need to take care of instead of ignoring it and saying _oh you stupid user_.

So, why does this happen in the first place and why does bufdelete wrapped inside a `vim.schedule` solve it?
Forget about winfixbuf.

### Section 1

`:e LICENSE`
`:Neotree float`

- **winid: 1000, GROUND, bufname: LICENSE**
- **winid: 1001, FLOAT , bufname: neo-tree filesystem [1]**

### Section 2

`:vs README.md`


- winid: 1000, GROUND, bufname: LICENSE
- winid: 1001, FLOAT , bufname: neo-tree filesystem [1]
- **winid: 1002, GROUND, bufname: README.md** (on the right of window 1000)

### Section 3

Since `:vs README.md` is a `BufEnter` event, neo-tree's steal prevention mechanism kicks in.
Keep in mind that the current window is 1002 so `utils.is_floating() == false` regardless of where we came from.
When you look inside that function, `prior_type == "neo-tree"` is true (around L217 in `setup/init.lua`) so `vim.cmd("b#")` is executed and the alt buffer ("#") is neo-tree's float so that's now in focus. Remember that a `utils.open_file` to open `README.md` is `vim.schedule`-ed right after that **\*1**.

- winid: 1000, GROUND, bufname: LICENSE
- winid: 1001, FLOAT , bufname: neo-tree filesystem [1]
- winid: 1002, GROUND, bufname: **neo-tree filesystem [1]**

### Section 4

At the same time, the operation is also a `WinEnter` event so `setup/init > M.win_enter_event` (around L264) is also executed. Inside that function, `manager.close_all("float")` is executed. This is what makes floating neo-tree automatically disappear when lost focus.
This will eventually end up at `ui/renderer > M.close`.
You can follow the conditions and eventually `pcall(vim.api.nvim_win_close, state.winid, ...` is called (L162) which closes the float window, good.

- winid: 1000, GROUND, bufname: LICENSE
- ~~winid: 1001, FLOAT~~ **REMOVED**
- winid: 1002, GROUND, bufname: neo-tree filesystem [1]

After that, `pcall(vim.api.nvim_buf_delete, bufnr, ...` (`bufnr == state.bufnr`) is called. 
Since vim will close the window together when the current buffer closes, this operation closes window 1002 as you can see from the above list.

- winid: 1000, GROUND, bufname: LICENSE
- ~~winid: 1001, FLOAT~~ **REMOVED**
- ~~winid: 1002, GROUND, bufname: neo-tree filesystem [1]~~ REMOVED

### Section 5

Do you remember **\*1**: the `utils.open_file` scheduled in _section 3_? That is executed now since all actions from the previous loop is finished.
Basically `utils.open_file` will open the specified buffer with `:b` and as you can see, the current focused buffer is window 1000, as this is the only window left.

- winid: 1000, bufname: **README.md**, GROUND

This is why there's only a single window left even when we did `:vs`.

---

## Why `vim.schedule(bufdel)` fixes that?

Let's rewind back to _section 4_.

- winid: 1000, GROUND, bufname: LICENSE
- ~~winid: 1001, FLOAT~~ **REMOVED**
- winid: 1002, GROUND, bufname: neo-tree filesystem [1]

We don't `vim.api.nvim_buf_delete` here so the situation remains like this, but we `schedule` to `bufdel(neo-tree filesystem [1])`.

_section 5_ starts before the scheduled bufdel so window 1002 is still alive and is the current window. So after section 5, we will now have ...

- winid: 1000, GROUND, bufname: LICENSE
- ~~winid: 1001, FLOAT~~
- winid: 1002, GROUND, bufname: **README.md**

And then bufdel happens but since `neo-tree filesystem [1]` buffer is not associated to any window, it is killed on its own without falling together with a window, so we have the split remaining.

## Just don't `vim.cmd("b#")` in the event if prior is float?

No if we don't do that, floating neo-tree becomes _steal-able_. Meaning that `:e README.md` on a floating window will open README.md inside that float, stealing neo-tree's window.

### Better if conditions?

That's exactly what I'm doing with the rewrite: https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/1340!!

pysan3:
@cseickel

cseickel:
Thanks for the detailed run down @pysan3. As far as I can see, the code in this PR still handles the float + winfixbuf window situation perfectly and creates a split when needed. You may not have noticed that it still has the scheduled buffer delete, it's just that decision on whether or not that buffer should be deleted happens inline instead of being scheduled.

I'm going to merge this because it seems to work as is.

pysan3:
No. It doesn't. You definitely missed something.

cseickel:
> No. It doesn't. You definitely missed something.

Huh, can you give me a numbered list of instructions how to recreate the bug?

pysan3:
@cseickel https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1390#issuecomment-1991553599

I didn't say anything about winfixbuf. It is a bug even from before the PR.

cseickel:
> @cseickel [#1390 (comment)](https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1390#issuecomment-1991553599)
> 
> I didn't say anything about winfixbuf. It is a bug even from before the PR.

I'm lost.

pysan3:
@cseickel Please do this and show me a screen record for both your fix and my PR. Sorry to bother your time.

- `:e <file-A>`
- `:Neotree float`
  - Don't close this
- `:vsplit <file-B>`

_!Don't close this_ was just a comment, meaning to literally press `:` and start writing _vsplit_ while floating neo-tree is in focus.

https://github.com/nvim-neo-tree/neo-tree.nvim/assets/41065736/b574ae94-7de4-4b5f-92c0-2201d38392fe

Lazy won't close automatically so there's a difference, but you can see that the new buffer should be created as a new split in the background.

cseickel:
Yeah, I have the same results when doing that particular test. That is such an oddball thing to do though that it's not worth causing bugs in the normal expected workflow in order to handle this really wacky edge case.

pysan3:
It is a real bug!!

Please read my comment https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1408#issuecomment-2009773131.
I explain why we now need to deal with this bug now seriously due to winfixbuf in the first few paragraphs.

cseickel:
Honestly, when I read that the first time I was assuming you meant to issue an open split command with a Neotree mapping, **because why would anyone actually issue `:vsplit <filename>` while the Neotree float is open?**. 

Logic would dictate that if you opened a file explorer, you would open the file from that file explorer. It's hard to imagine why you would open a floating file picker and then manually issue a command to open a file while that floating window is still open.

pysan3:
> However, when all ground windows are winfixbuf, we now need to make a new split before placing that buffer, just like the mechanism used for `left/right` neo-trees (when neo-tree is the only opened window).
> 
> This is why `float + vsplit` is now an official issue we need to take care of instead of ignoring it and saying _oh you stupid user_.

As I said here, when there's no ground window available because all are winfixbuf, we need to make a new split before opening the buffer.

This is exactly the same as calling `:vs README.md` while the floating window is open.

So if `float + :vs` doesn't work, winfixbuf will also not work.
It is easier and more obvious to run `:vs` than preparing a setup with winfixbuf.

Does this make sense?

cseickel:
> > However, when all ground windows are winfixbuf, we now need to make a new split before placing that buffer, just like the mechanism used for `left/right` neo-trees (when neo-tree is the only opened window).
> > This is why `float + vsplit` is now an official issue we need to take care of instead of ignoring it and saying _oh you stupid user_.
> 
> As I said here, when there's no ground window available because all are winfixbuf, we need to make a new split before opening the buffer to make a place where the buffer can be opened.
> 
> This is exactly the same as calling `:vs README.md` while the floating window is open.

Nope, I don't agree with that at all.

> 
> So if `float + :vs` doesn't work, winfixbuf support is broken as well. It is easier and more obvious to run `:vs` than preparing a setup with winfixbuf so that's what I used for testing.

Nope.

> 
> Does this make sense?
> 
> * `:e release.sh`
> * `:set winfixbuf`
> * `:Neotree float`
> * focus README.md
> * `<CR>` (open the file)
> 
> Now, because the ground window (where `release.sh` lives) is a winfixbuf, README.md cannot be opened in that window. Since that's the only available option, we need to make a blank split next to `release.sh` instead to open README.md.
> 

Those instructions do make sense, and that actually works fine in `main` right now, even though issuing a `:vsplit` command does not make a split when neotree float is open.

regardless, I think #1406 is a good improvement that handles everything so I guess the discussion is resolved.

pysan3:
> > > However, when all ground windows are winfixbuf, we now need to make a new split before placing that buffer, just like the mechanism used for `left/right` neo-trees (when neo-tree is the only opened window).
> > > This is why `float + vsplit` is now an official issue we need to take care of instead of ignoring it and saying _oh you stupid user_.
> > 
> > 
> > As I said here, when there's no ground window available because all are winfixbuf, we need to make a new split before opening the buffer to make a place where the buffer can be opened.
> > This is exactly the same as calling `:vs README.md` while the floating window is open.
> 
> Nope, I don't agree with that at all.

Why? You do deal with this problem with `left/right` windows.

- `:e release.sh`
- `:Neotree left`
- `<C-w>l` (go back to release.sh)
- `<C-w>q` (delete that window)
- now neo-tree covers the whole screen
- focus README.md
- `<CR>`


_when there's no ~~ground~~ window available ~~because all are winfixbuf~~ **except for neo-tree**, we need to make a new split before opening the buffer to make a place where the buffer can be opened.
This is exactly the same as calling `:vs README.md` while the ~~floating~~ **side** window is open._

I think I'm saying the same thing? And why do you deal with this case but not for the floating window? I don't agree with this inconsistency.

pysan3:
@cseickel

cseickel:
I'm sorry I should have explained better. What I meant was that I disagree with the assertion that manually issuing a `:vsplit` command is somehow equivalent to using the neotree float window to open a file. The process that it goes through to end up creating a new split when it is needed is completely different and the manual split command was not a valid test case.

pysan3:
> The process that it goes through to end up creating a new split when it is needed is completely different and the manual split command was not a valid test case.

Could you explain the difference?

Manual split works -> but neo-tree doesn't may happen but I don't think for the other way around.
If manual split doesn't work, neo-tree's creating new split code will also always not work as well.

So manual splitting should be a valid test case.

cseickel:
I know empirically that it is not the same because during testing of the  various states of changes, using Neotree to open a file either normally or in vsplit would work just fine while your `:vsplit <filename>` did not work as expected.

Without that evidence though, I know that it opening a file from neotree does not issue a vsplit while the floating window is in focus. It switches to the last used window first then sees if it is usable. If not, it keeps switching windows until it finds one it likes and then issues the command. There are more nuances if the only open window is a neo-tree window that is supposed to be a sidebar.

Issue #1415, opened 2024-03-25
BUG: Buffer with this name already exists after manually closing neo-tree


idanarye:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

0.9.5

### Operating System / Version

Arch Linux 6.7.0-arch3-1

### Describe the Bug

This looks similar to #1365, but since I don't use sessions (the discussion there is mostly about how fix it by fixing session management with a new `)auto_restore_session_experimental = true` setting) and since I've only started experiencing this a couple of days ago (that ticket is almost a month old), I suspect it may be a different bug.

When closing the Neo-tree panel using Vim commands (like `:quit` or like running `:only` from another window), another attempt to open the Neo-tree panel will result in a `Buffer with this name already exists` error.

Note that when I close the panel using Neo-tree (e.g. - by hitting `q` inside it) this problem does not occur.

### Screenshots, Traceback

```
[Neo-tree ERROR] debounce  filesystem_navigate  error:  ...e-bug/plugins/neo-tree.nvim/lua/neo-tree/ui/renderer.lua:1036: Vim:E95: Buffer with this name already exists
```

### Steps to Reproduce

1. Run `:Neotree` to open the Neo-tree panel.
2. Run `:close` inside the Neo-tree panel to close it (don't use any of Neo-tree's shortcuts/commands for closing the panel - use Vim's window management commands)
3. Run `:Neotree` again.

### Expected Behavior

I'd expect the Neo-tree panel to just open again.

### Your Configuration

```Lua
local lazypath = "/tmp/reproduce-bug/lazy/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
    vim.fn.system({
        "git",
        "clone",
        "--filter=blob:none",
        "https://github.com/folke/lazy.nvim.git",
        "--branch=stable", -- latest stable release
        lazypath,
    })
end
vim.opt.rtp:prepend(lazypath)

require"lazy".setup({
    {
        "nvim-neo-tree/neo-tree.nvim",
        branch = "v3.x",
        dependencies = {
            "nvim-lua/plenary.nvim",
            "nvim-tree/nvim-web-devicons", -- not strictly required, but recommended
            "MunifTanjim/nui.nvim",
            -- "3rd/image.nvim", -- Optional image support in preview window: See `# Preview Mode` for more information
        }
    }
[...]

PR #1418, opened 2024-03-27
fix(renderer): check if current buffer is loaded

pysan3:
closes: #1415

I failed to detect one edge case in
- https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1406

Old code was guaranteed that `state.bufnr` was deleted inside `close_old_window` function but now only when `window_existed`.
Since you changed the code so that buffer might survive even when window is closed a couple months ago, we needed to delete the old remaining buffer regardless of `window_existed` for this specific if-cases.

### DIFF
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -778,6 +778,7 @@ create_tree = function(state)
   state.tree = NuiTree({
     ns_id = highlights.ns_id,
     winid = state.winid,
+    bufnr = state.bufnr,
     get_node_id = function(node)
       return node.id
     end,
--- a/lua/neo-tree/ui/renderer.lua
+++ b/lua/neo-tree/ui/renderer.lua
@@ -163,13 +163,15 @@ M.close = function(state, focus_prior_window)
     end
     state.winid = nil
   end
-  local bufnr = utils.get_value(state, "bufnr", 0, true)
-  state.bufnr = nil
-  vim.schedule(function()
-    if bufnr > 0 and vim.api.nvim_buf_is_valid(bufnr) then
-      vim.api.nvim_buf_delete(bufnr, { force = true })
-    end
-  end)
+  if window_existed then
+    local bufnr = utils.get_value(state, "bufnr", 0, true)
+    state.bufnr = nil
+    vim.schedule(function()
+      if bufnr > 0 and vim.api.nvim_buf_is_valid(bufnr) then
+        vim.api.nvim_buf_delete(bufnr, { force = true })
+      end
+    end)
+  end
   return window_existed
 end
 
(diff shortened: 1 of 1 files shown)


pysan3:
Umm I might've done something wrong with git history.

Only the last commit is what is related to this PR.

lua/neo-tree/ui/renderer.lua
```diff
@@ -163,13 +163,15 @@ M.close = function(state, focus_prior_window)
...
     state.winid = nil
   end
-  local bufnr = utils.get_value(state, "bufnr", 0, true)
-  state.bufnr = nil
-  vim.schedule(function()
-    if bufnr > 0 and vim.api.nvim_buf_is_valid(bufnr) then
-      vim.api.nvim_buf_delete(bufnr, { force = true })
-    end
-  end)
+  if window_existed then
```
cseickel, line 166: Why does it matter if the window existed? Why should a buffer be allowed to remain if the window was closed by some other means, but not if it was closed by this function?

Also, why is it **always** scheduled even though it usually should be deleted synchronously?

I think the code from 3.22 was already correct and would fix this issue and we should just go back to that:

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/8afbb06081ce1e4beb5b18945d14a608b10babeb/lua/neo-tree/ui/renderer.lua#L167-L175

Whatever the other edge case was which the changes were trying to fix should be handled another way.

lua/neo-tree/ui/renderer.lua
```diff
@@ -1010,6 +1013,9 @@ M.acquire_window = function(state)
       vim.api.nvim_win_set_buf(state.winid, state.bufnr)
     else
       close_old_window()
+      if state.bufnr and vim.api.nvim_buf_is_valid(state.bufnr) then
```
cseickel, line 1016: `M.close` is already deleting the buffer. Why are we double deleting? Is it because of your other change which will only sometimes delete the old buffer? It seems like these two changes are in conflict with another.

Also, if i remember correctly, the original code from a few months ago already did what we are slowly getting back to, which is to immediately attempt to delete the buffer and then schedule a delete for later only if we get a specific error.

I think the right place to make this fix is in the `M.close()` code, we shouldn't be deleting the buffer multiple times and from different locations.

lua/neo-tree/ui/renderer.lua
cseickel: I have confirmed that the old code works correctly and the bug in #1415 was the direct result of changing that code. I'm going to close this because reverting that code is the real fix.

cseickel:
Closing in favor of #1433

chaozwn:
> Closing in favor of #1433

hi, after test. this issue not fixed. you can test when neo-tree is open, then `<C-w>o`, reopen neo-tree.
image

chaozwn:
> Closing in favor of #1433

but this issue can fixed it.

cseickel:
> > Closing in favor of #1433
> 
> but this pr can fix it.

That PR was merged to `main` and it works for me. Can you confirm that it is fixed for you using the `main` branch?

Issue #1419, opened 2024-03-30
FEATURE: Enhancement for the `set_root` command


xvzc:
### Did you check the docs?

- [X] I have read all the docs.

### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like.

Hi, I've been using my custom command called "sync_root" which looks like below.
```lua
  commands = {
    sync_root = function(state)
      if state.search_pattern then
        fs.reset_search(state, false)
      end

      local node = state.tree:get_node()
      local id = node.type == "directory" and node.id or node._parent_id
      fs._navigate_internal(state, id, nil, nil, false)
    end,
  },
```
While the built in `set_root` command allows us to run it only when the neovim cursor is pointing to directories, the `sync_root` function above allows us to set the current directory as root regardless of the location of cursor.

So my question is, would you be interested if i make a PR that enhances the `set_root` commands for `filesystem` and `buffer`?

cseickel:
Sure, I don't see how it can hurt.

PR #1420, opened 2024-03-30
feat(set_root): allow filesystem and buffers to run set_root when the cursor is on both files and directories

xvzc:
This Pull Request adds features that has been mentioned in #1419, and enhances the functionality of `set_root` for `Neotree buffers` and `Neotree filesystem`.

### DIFF
--- a/lua/neo-tree/sources/filesystem/commands.lua
+++ b/lua/neo-tree/sources/filesystem/commands.lua
@@ -222,14 +222,13 @@ M.rename = function(state)
 end
 
 M.set_root = function(state)
-  local tree = state.tree
-  local node = tree:get_node()
-  if node.type == "directory" then
-    if state.search_pattern then
-      fs.reset_search(state, false)
-    end
-    fs._navigate_internal(state, node.id, nil, nil, false)
+  if state.search_pattern then
+    fs.reset_search(state, false)
   end
+
+  local node = state.tree:get_node()
+  local id = node.type == "directory" and node.id or node._parent_id
+  fs._navigate_internal(state, id, nil, nil, false)
 end
 
 ---Toggles whether hidden files are shown or not.
--- a/lua/neo-tree/sources/filesystem/commands.lua
+++ b/lua/neo-tree/sources/filesystem/commands.lua
@@ -227,16 +227,16 @@ M.set_root = function(state)
   end
 
   local node = state.tree:get_node()
+  while node and node.type ~= "directory" do
+    local parent_id = node:get_parent_id()
+    node = parent_id and state.tree:get_node(parent_id) or nil
+  end
+
   if not node then
     return
   end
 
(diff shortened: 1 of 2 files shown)


lua/neo-tree/sources/filesystem/commands.lua
```diff
@@ -222,14 +222,13 @@ M.rename = function(state)
...
-    if state.search_pattern then
-      fs.reset_search(state, false)
-    end
-    fs._navigate_internal(state, node.id, nil, nil, false)
+  if state.search_pattern then
+    fs.reset_search(state, false)
   end
+
+  local node = state.tree:get_node()
+  local id = node.type == "directory" and node.id or node._parent_id
```
pysan3, line 230: No, access to `node.id` and `node._parent_id` is not allowed.

Use the get methods listed in the doc.
https://github.com/MunifTanjim/nui.nvim/tree/main/lua/nui/tree#nuitreenode

pysan3:
@xvzc Sorry for many requests. 

`tree:get_node()` as well as `get_parent_id()` may return `nil` so could you check and bail out when that's the case?

(Cursor is on a line with no nodes, cursor is on the root node)

A very simple `if not xxx then return end` is enough. 
Thank you ;)

xvzc:
No problem. Thank you for your instruction :)

pysan3:
My bad, you'll need to `if not node return` before local id because `<nil>.type` will error out. 

6 lines = 3 for not node and 3 for not id.

pysan3:
Thanks! LGTM.

lua/neo-tree/sources/filesystem/commands.lua
```diff
@@ -222,14 +222,21 @@ M.rename = function(state)
...
+  if state.search_pattern then
+    fs.reset_search(state, false)
   end
+
+  local node = state.tree:get_node()
+  if not node then
+    return
+  end
+
+  local id = node.type == "directory" and node:get_id() or node:get_parent_id()
```
cseickel, line 234: Sorry to be late to the party, but this is not quite right because it assumes there are only files and folders, but there may also be nested files. To get a parent folder via navigating up nodes, it needs to be a while loop which tests for `node.type == "directory"` in each loop and has some safety bail out to prevent an infinite loop.

The alternative is to get the parent by splitting the filename/path with `utils.split_path`.

xvzc: Hi, Thank you for your comment, but i actually have no idea what nested files are like. So could you please provide some specific example of `nested files`?

cseickel: Yes, that's perfect.

BTW - this is the nested file feature: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/118

cseickel (approved):

PR #1435, opened 2024-04-15
fix: sources.manager.dispose* functions not working

AThePeanut4:
This is a simple change to fix the `sources.manager.dispose*` functions, which aren't currently working properly.

The bug is a classic remove while looping issue, solved by iterating in reverse. I've just copied the change from the `dispose_invalid_tabs` function.


cseickel (approved):
Thanks for the fix!

PR #1443, opened 2024-04-23
doc: minor updates to help doc

dmspelma:
I went through the neotree help doc to set up and customize my neotree.

I want to leave the Documentation in a better place after reading.
These are just minor discrepancies I noticed.


pysan3 (approved):
My bad. This slipped off my radar. Thanks for your contribution!

PR #1488, opened 2024-05-30
fix(popups): dont add spacing when the title is empty

sand4rt:
Screenshot 2024-05-31 at 00 30 11

vs

Screenshot 2024-05-31 at 00 30 43


pysan3 (approved):
Amazing. Thanks!

PR #1492, opened 2024-06-04
Pre-defined File Nesting Rules in Neotree

saifulapm:
Not sure this will be merge or not. If anyone looking for it, they can easily find it if it's on readme. This is ported from vscode plugin. 

Here is the link [neotree-file-nesting-config](https://github.com/saifulapm/neotree-file-nesting-config)

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -573,7 +573,7 @@ See `:h neo-tree-commands` for details and a full listing of available arguments
 
 ### File Nesting
 
-See `:h neo-tree-file-nesting` for more details about file nesting.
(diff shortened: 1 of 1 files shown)


miversen33 (commented):
Thank you for this! IMO it makes more sense to put something like nesting examples in the wiki as opposed to a link to a repo in the readme.

Checkout the [recipes section of the wiki ](https://github.com/nvim-neo-tree/neo-tree.nvim/wiki/Recipes), that seems like a great spot for the examples in the linked repo :)

PR #1510, opened 2024-07-05
feat(components): add support for `mini.icons`

mehalter:
This adds support for `mini.icons` for providing icons for not just files but also directories

2024-07-05_10:09:40_screenshot

### DIFF
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -293,22 +293,31 @@ end
 M.icon = function(config, node, state)
   local icon = config.default or " "
   local highlight = config.highlight or highlights.FILE_ICON
+  local _, mini_icons = pcall(require, "mini.icons")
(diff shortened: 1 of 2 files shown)


miversen33:
I maybe making this a bigger thing than it should be, but this PR makes me wonder if we should simply allow for the user to supply an "icon" function that we can call to get the icon for a node.

I can easily see a future where more icon packs come out and we end up having to support `xyz` icon packs and its done via nasty "if/elseif" logic. 

I don't imagine it would be super hard to do this either as we _already_ have some logic in place for getting the icon for a node in the components (like you are doing now with this).

That is just my 2cents though, I could also see the argument that it's not important, or [a user can already do that if they want](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/main/doc/neo-tree.txt#L1019).

mehalter:
I actually really like this idea. Let me revise this PR and make a recommendation for a general system that can work here!

PR #1558, opened 2024-09-13
Check for nil state.tree

nickvonkaenel:
I've been getting a crash with certain directories because it's possible for the state table to not contain a tree temporarily. This fixes the issue.


miversen33 (approved):
LGTM

Issue #1459, opened 2024-05-05
BUG: Filter doesn't work


asynchroza:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.9.5

### Operating System / Version

MacOS Sonoma 14.4.1

### Describe the Bug

Filter doesn't seem to be doing anything. No exception messages are shown, but no results are returned.

### Screenshots, Traceback

![image](https://github.com/nvim-neo-tree/neo-tree.nvim/assets/104720011/9036c9ca-adea-4241-ad9c-8cfae706fc1d)

### Steps to Reproduce

1. Pull my config -> https://github.com/asynchroza/nvim-config
2. Open neo-tree
3. Try to filter by `.lua` extension.

### Expected Behavior

It should display all files including `.lua`

### Your Configuration

```Lua
-- references:
-- https://github.com/nvim-neo-tree/neo-tree.nvim
-- https://github.com/nvim-neo-tree/neo-tree.nvim/wiki/Recipes
return {
	"nvim-neo-tree/neo-tree.nvim",
	dependencies = {
		"nvim-lua/plenary.nvim",
		"nvim-tree/nvim-web-devicons",
		"MunifTanjim/nui.nvim",
	},
	event = "VeryLazy",
	keys = {
		{ "<leader>e", ":Neotree position=float reveal_force_cwd<CR>", silent = true, desc = "Float File Explorer" },
		{ "<leader><tab>", ":Neotree toggle left<CR>", silent = true, desc = "Left File Explorer" },
	},
	config = function()
		require("neo-tree").setup({
			close_if_last_window = true,
			popup_border_style = "single",
			enable_git_status = true,
			enable_modified_markers = true,
			enable_diagnostics = true,
			sort_case_insensitive = true,
			default_component_configs = {
				indent = {
					with_markers = true,
					with_expanders = true,
				},
				modified = {
					symbol = " ",
					highlight = "NeoTreeModified",
				},
				icon = {
					folder_closed = "",
					folder_open = "",
					folder_empty = "",
					folder_empty_open = "",
				},
				git_status = {
					symbols = {
						-- Change type
						added = "",
						deleted = "",
						modified = "",
						renamed = "",
						-- Status type
						untracked = "",
						ignored = "",
						unstaged = "",
						staged = "",
						conflict = "",
					},
				},
			},
[...]

PR #1564, opened 2024-09-18
fix(filesystem): ignore dotfiles properly for find filter

obukhovaa:
Not sure what was the purpose of that exclusion, but it was causing issues when `find` command is used within `filter` logic. 
```lua
if not ignore.dotfiles then
  append("-not", "-path", "*/.*")
end
```
You would expect that when `ignore-dotfiles` is `false` you don't want to exclude dot files and directories. In fact it was not just excluding dot files, but all the other valid options if current `path` contains any `.` directories. For example for me the resulting command was
```bash
find  "/Users/nouwa/.config/nvim" -not -path '*/.*' -iname "*lua*"
```
and because my path has `.config` -- all files were excluded, producing empty search result.

Once this section removed, eveyrhing works as expected with both `ignore.dotfiles = true` and `false`.

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/filter_external.lua
+++ b/lua/neo-tree/sources/filesystem/lib/filter_external.lua
@@ -193,9 +193,6 @@ M.filter_files_external = function(
     if types.executable then
       append("-executable")
     end
-    if not ignore.dotfiles then
-      append("-not", "-path", "*/.*")
-    end
     if glob ~= nil and not full_path then
       append("-iname", glob)
     elseif glob ~= nil and full_path then
--- a/lua/neo-tree/sources/filesystem/lib/filter_external.lua
+++ b/lua/neo-tree/sources/filesystem/lib/filter_external.lua
@@ -184,6 +184,9 @@ M.filter_files_external = function(
         file_types[#file_types + 1] = file_type_map[k]
       end
     end
+    if ignore.dotfiles then
+      append("-name", ".*", "-prune", "-o")
+    end
     if #file_types > 0 then
       append("-type", table.concat(file_types, ","))
     end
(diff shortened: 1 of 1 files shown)


asynchroza:
Good stuff, @obukhovaa! I just pulled it locally and it works.

obukhovaa:
Still an issue https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1459, would be nice to move it forward

cseickel:
Sorry, I can't properly evaluate this. See https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1606

lua/neo-tree/sources/filesystem/lib/filter_external.lua
```diff
@@ -193,9 +193,6 @@ M.filter_files_external = function(
     if types.executable then
       append("-executable")
     end
-    if not ignore.dotfiles then
-      append("-not", "-path", "*/.*")
-    end
```
pynappo, line 198: I agree that it's obviously breaking but I don't know if it's necessarily redundant. Based on the code for `fd`, the intention seems to be to optionally exclude hidden folders and hopefully reduce the amount of lines from the external cmd that neo-tree has to process?

To get the same functionality as the 'fd' block, could you try this instead? seems to work on my end:
```lua
    -- right after `for k, v in pairs(types) do ... end`
    if ignore.dotfiles then
      append("-name", ".*", "-prune", "-o")
    end
```

```lua
    -- right before the call to append_find_args:
    -- explicitly print out entries since -o may cause it to not work as intended
    append("-print")
```

obukhovaa: Indeed, I've tested with both enabled/disabled setting, works well so far

obukhovaa:
I've pushed the suggested improvement on top of the original fix, tested a few direct cases, didn't find any issues so far, so I think we good.

pynappo (approved):

PR #1568, opened 2024-09-23
feat: add close to fuzzy_finder_mappings

agoodshort:
Useful for users who wants to map something else than `<esc>` (e.g. `<C-e>` or `<C-c>)` to close the fuzzy finder input window

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/filter.lua
+++ b/lua/neo-tree/sources/filesystem/lib/filter.lua
@@ -12,17 +12,6 @@ local manager = require("neo-tree.sources.manager")
 
 local M = {}
 
-local cmds = {
-  move_cursor_down = function(state, scroll_padding)
-    renderer.focus_node(state, nil, true, 1, scroll_padding)
-  end,
-
-  move_cursor_up = function(state, scroll_padding)
-    renderer.focus_node(state, nil, true, -1, scroll_padding)
-    vim.cmd("redraw!")
-  end,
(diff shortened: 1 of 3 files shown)


cseickel (commented):
This is probably good, but I don't have the time to properly review. See: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1606

lua/neo-tree/sources/filesystem/lib/filter.lua
```diff
@@ -215,22 +206,41 @@ M.show_filter = function(state, search_as_you_type, fuzzy_finder_mode, use_fzy)
...
 
   if fuzzy_finder_mode then
     local config = require("neo-tree").config
     for lhs, cmd_name in pairs(config.filesystem.window.fuzzy_finder_mappings) do
       local t = type(cmd_name)
       if t == "string" then
-        local cmd = cmds[cmd_name]
-        if cmd then
-          input:map("i", lhs, create_input_mapping_handle(cmd, state, scroll_padding), { noremap = true })
+        if cmd_name == "close" then
```
pynappo, line 220: I think it might make more sense to rename the close_input function and put it in the cmds local along with `move_cursor_up/down` like:

```lua
local cmds = {
  -- ...
  close = function()
    -- ...
  end
}
```

Then we can remove the extra nesting of if-else.

pynappo (approved):

Issue #1512, opened 2024-07-06
FEATURE: indent marker level


sand4rt:
### Did you check the docs?

- [X] I have read all the docs.

### Is your feature request related to a problem? Please describe.

An option to adjust the indentation marker line level, beginning from the first column:
![20240707_00h26m34s_grim](https://github.com/nvim-neo-tree/neo-tree.nvim/assets/17591696/078d26fa-330d-4919-8f22-fa10d51ee062)

### Describe the solution you'd like.

```lua
{
    default_component_configs = {
        indent = {
            level = 0,
        }
    }
}
```

PR #1570, opened 2024-09-28
feat: configurable indent marker level

sand4rt:
closes https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1512

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ use {
           },
           indent = {
             indent_size = 2,
+            level = 2,
             padding = 1, -- extra padding on left hand side
             -- indent guides
             with_markers = true,
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -1035,6 +1035,7 @@ for each source, or just do it once in the default_component_configs section:
           indent_marker = "│",
           last_indent_marker = "└",
           indent_size = 2,
+          level = 2
         },
       },
     })
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -389,6 +389,7 @@ M.indent = function(config, node, state)
   local skip_marker = state.skip_marker_at_level
   local indent_size = config.indent_size or 2
   local padding = config.padding or 0
+  local start_level = config.level or 2
   local level = node.level
   local with_markers = config.with_markers
   local with_expanders = config.with_expanders == nil and file_nesting.is_enabled()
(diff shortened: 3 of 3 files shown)


cseickel (commented):
This one is questionable. I don't know for sure, but other code may assume that level 0 means the root node and this could be more than just a visual tweak.

There is also a padding option on the container which may be more appropriate. Either way, see: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1606

sand4rt:
@pynappo could this be merged? i use this for a while now without issues

pynappo:
The option should be renamed to something more descriptive like `guide_start_level`. also, add it to defaults.lua

also imo the cases of level 0/1 should be better handled to have no padding on the left, similar to level 2 and up

0 should have no padding on the left here:

image

1 should have no padding on the left here:
image

2 looks fine:

image

and-rs:
I would like this merged too. Any way in which I could contribute?

and-rs (approved):
I'd say this is good

sand4rt:
> I would like this merged too. Any way in which I could contribute?

Gladly! Feel free to create a PR against this branch or rebase it and open a new PR, then ill close this one.

> I'd say this is good

There are still some issues in there. This is described in: https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1570#issuecomment-2676081748

and-rs:
alrightt of course, there were some issue with the implementation and I also have some questions about how we should handle the end result.

1. Do we really want the indentation marker to start from the node 0? Or is it fine at the node 1? Because I honestly think that node 0 is just not useful at all. I wanted to implement this because if I hid the root node with `hide_root_node = true`, the first nodes did not have a marker. Take a look at this:

combined

2. if there are hidden files in some directory after a node with markers, the marker doesn't get rendered, and I am just not sure how to deal with that. 

- This is another example on how it looks with `guide_level_start = 0` and with `hide_root_node = false`.

Screenshot 2025-02-26 at 8 27 30 PM

sand4rt:
> Do we really want the indentation marker to start from the node 0? / This is another example on how it looks with `guide_level_start = 0`

Yes, this is what the issue is about. I like how it looks. Perhaps this is fixable with an `if opts.hide_root_node then ..`

> there are hidden files in some directory after a node with markers

Not sure either, could send what you already have? Maybe someone could help out with it?

and-rs:
Alright if you like the node 0 implementation, I'd say we leave it like that, I don't even think a conditional would be necessary, and it would allow for more customization, anyone could decide if they want the indent marker from the node 0 with or without the root node hidden. I also recommend calling the option, `marker_start_level` just like the other options related to the indent marker. 

I was wondering if I could commit directly to this PR branch, or if another PR is needed, I appreciate the guidance.

and-rs:
alright, I just made a new PR #1705 with the changes we discussed here. feel free to test them

sand4rt:
closing in favour of: https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1705

PR #1581, opened 2024-10-12
feat(common): add `rename_basename` command

dunix241:
Implemented:

- command `rename_basename` that behaves analogously to `rename` while allowing to rename the node without the extension.

Motivation:

I don't think I need to change a file's extension too often, so, this would shorten the time it takes to rename the file when not touching its extension.


cseickel:
https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1606

lua/neo-tree/defaults.lua
```diff
@@ -406,6 +406,7 @@ local config = {
       ["A"] = "add_directory", -- also accepts the config.show_path and config.insert_as options.
       ["d"] = "delete",
       ["r"] = "rename",
+      ["b"] = "rename_basename",
```
pynappo, line 409: Can you also add it here in the README as well: https://github.com/nvim-neo-tree/neo-tree.nvim/blob/395ce4675a54828f9fd730d78f7910c114ab0a49/README.md?plain=1#L286

lua/neo-tree/sources/filesystem/lib/fs_actions.lua
```diff
@@ -617,4 +611,34 @@ M.rename_node = function(path, callback)
...
+  local parent_path, name = utils.split_path(path)
+  local base_name = vim.fn.fnamemodify(path, ":t:r")
+  local extension = vim.fn.fnamemodify(path, ":e")
+
+  local msg = string.format('Enter new name for "%s":', name)
+
+  local get_destination = function(new_name)
+    return parent_path
+      .. utils.path_separator
+      .. new_name
```
pynappo, line 637:
```suggestion
  local get_destination = function(new_base_name)
    return parent_path
      .. utils.path_separator
      .. new_base_name
```

pynappo (changes requested):
Overall looks good to me. if you don't mind waiting a bit I'll test out the PR after my final exams end on Tuesday, just to make sure this works as intended.

dunix241:
> Overall looks good to me. if you don't mind waiting a bit I'll test out the PR after my final exams end on Tuesday, just to make sure this works as intended.

Good luck with your exams! 😅
Hope you'll do well I don't mind waiting

pynappo (approved):

Issue #1301, opened 2024-01-11
FEATURE: Copy directories/files over different instances of neo-tree


jugarpeupv:
### Did you check the docs?

- [X] I have read all the docs.

### Is your feature request related to a problem? Please describe.

The problem is better described in nvim-tree feature request https://github.com/nvim-tree/nvim-tree.lua/issues/1549

In this thread there is a solution that is working in nvim-tree, i wonder if there is plan on neo-tree to support this feature.

### Describe the solution you'd like.

The solution could be the same as the one described in nvim-tree feature request, i guess

cseickel:
It would be cool, and not that hard to do. I would personally prefer the temp file approach rather than using the system clipboard.

I will accept a PR for this but I probably won't have time to do it myself.

Issue #1235, opened 2023-11-17
FEATURE: Shared neo-tree window across tabs


sedmicha:
### Did you check the docs?

- [X] I have read all the docs.

### Is your feature request related to a problem? Please describe.

The README states this as one of the features:
- Neo-tree windows in different tabs are completely separate

While this might be the optimal behavior for many people, it doesn't really work for me, and I would like a configuration option to do the opposite - have one Neotree instance shared across all my tabs. 

### Describe the solution you'd like.

The easiest solution that comes to mind would be a true/false flag in the config options, e.g. something like
```lua
separate_windows_across_tabs = false
```

### Describe alternatives you've considered.

Another option which I think would be cool and more flexible would be the ability to provide an something like an "instance ID" to the Neotree command as an optional argument. So for example doing
```
:Neotree float id=1
```
would always bring up the same Neotree instance as long as the same id is provided. This would default to the tabid if not provided, as is now.

### Additional Context

I hope I'm not missing something obvious, but I spent a lot of time searching the internet, reading the docs, and even looking through the source code to find if there's a way to do this already, with no success.

pysan3:
I also open one neovim instance per directory, so even in different tabs, the tree will be identical and I actually wanted this kind of feature as well.

I think the cleanest solution would be to hijack `vim.api.nvim_get_current_tabpage()` and replace all occurrence with a function like below.

```lua
local function get_current_tabpage(tabid)
  if require("neo-tree").config.shared_tree_across_tabs then
    return 0 -- always return a fake constant
  else
    return tabid or vim.api.nvim_get_current_tabpage()
  end
end
```

I need to take care of few places where [`to_tabnr`](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/230ff118613fa07138ba579b89d13ec2201530b9/lua/neo-tree/sources/manager.lua#L295) is called to not break it but this should not be a big deal.

I honestly think this is somewhat in the region of _clever solution_ rather than a nasty hack, but @cseickel would you agree on this approach? Or even this idea in the first place?

If it is ok, I'll definitely test it thoroughly and will make a more detailed PR, but I'd love to hear your opinion before start working on it.

cseickel:
@pysan3 First off, I think it's likely that this is something where there are going to be a lot of unexpected bugs that pop up that you won't find until it gets released. I say that because one tree per tab was such a core assumption in the original design.

[...]

Issue #168, opened 2022-03-03
Cut/Copy/paste between windows


cseickel:
Right now the cut/copy clipboard only works within a window. It might be nice to have a clipboard that is shared within a tab so that you can create two neo-tree windows in splits and be able to paste those items into a second tree.

miversen33:
Yes. In fact, this could be a shared clipboard with neo-tree and any providers within it (netman would love this <3 )

PR #1594, opened 2024-10-28
feat(filesystem): Implement shared clipboard

miroshQa:
**This pull request aims to solve https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1301 issue by adding the  shared_clipboard functionality**

And it also probably solves these two:
https://github.com/nvim-neo-tree/neo-tree.nvim/issues/168
https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1235

You can watch how it works here:

https://github.com/user-attachments/assets/a651d061-f447-442a-b684-9cbc08ff569c

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/shared_clipboard.lua
+++ b/lua/neo-tree/sources/filesystem/lib/shared_clipboard.lua
@@ -1,0 +1,92 @@
+local fs_watch = require("neo-tree.sources.filesystem.lib.fs_watch")
+local manager = require("neo-tree.sources.manager")
+local events = require("neo-tree.events")
+local renderer = require("neo-tree.ui.renderer")
+
+local M = {}
+
+local clipboard_state_dir_path = vim.fn.stdpath("state") .. "/neo-tree/"
+local clipboard_file_path = clipboard_state_dir_path .. "filesystem-clipboard.json"
+local clipboard_file_last_mtime = nil
+
+
+M.save_clipboard = function(clipboard)
+  local file = io.open(clipboard_file_path, "w+")
+  -- We want to erase data in the file if clipboard is nil instead writing null
+  if not clipboard or not file then
+    return
+  end
+
+  file:write(vim.json.encode(clipboard))
+  file:flush()
+  clipboard_file_last_mtime = vim.uv.fs_stat(clipboard_file_path).mtime
+end
+
+M._load_clipboard = function()
+  local file = io.open(clipboard_file_path, "r")
+  if not file then
+    return nil
+  end
+  local content = file:read("*a")
+  local is_success, clipboard = pcall(vim.json.decode, content)
+  if not is_success then
+    return nil
+  end
+  return clipboard
+end
+
+M._update_cilpboard = function()
+  if not vim.fn.filereadable(clipboard_file_path) then
+    return
+  end
+  local cur_mtime = vim.uv.fs_stat(clipboard_file_path).mtime
+  if
+    not clipboard_file_last_mtime
+    -- We need exactly >= because it allows us to synchronize the clipboard
+    -- even with trees openned in another window (in the current Neovim instance)
+    or M._is_left_mtime_greater_or_equal(cur_mtime, clipboard_file_last_mtime)
+  then
+    local clipboard = M._load_clipboard()
+    manager._for_each_state("filesystem", function(state)
+      state.clipboard = clipboard
+      vim.schedule(function()
+        renderer.redraw(state)
+      end)
+    end)
+    clipboard_file_last_mtime = cur_mtime
+  end
+end
+
+M._is_left_mtime_greater_or_equal = function(a, b)
+  if a.sec > b.sec then
+    return true
+  elseif a.sec == b.sec and a.nsec >= b.nsec then
+    return true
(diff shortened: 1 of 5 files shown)


lua/neo-tree/sources/filesystem/lib/shared_clipboard.lua
```diff
@@ -0,0 +1,92 @@
...
+M._update_cilpboard = function()
+  if not vim.fn.filereadable(clipboard_file_path) then
+    return
+  end
+  local cur_mtime = vim.uv.fs_stat(clipboard_file_path).mtime
+  if
+    not clipboard_file_last_mtime
+    -- We need exactly >= because it allows us to synchronize the clipboard
+    -- even with trees openned in another window (in the current Neovim instance)
+    or M._is_left_mtime_greater_or_equal(cur_mtime, clipboard_file_last_mtime)
```
cseickel, line 47: It seems like a bad idea to read from the file we just wrote as a means of syncing within an instance. 

I would just extract out the update logic so it can also be called from `save_clipboard` to handle the internal syncing, then make this a greater than (but not equal to) check.

OR: Another way to handle this which may be cleaner is to have each instance write it's own file, maybe use the process ID or something to identify it. Then you can ignore your own file from the event and just handle any other events that are thrown regardless of the modified time.

cseickel (commented):
Thanks for the PR,I can't wait to use it! There are a few things I think we can improve.

miroshQa:
Fixed it!

lua/neo-tree/sources/filesystem/lib/shared_clipboard.lua
```diff
@@ -0,0 +1,79 @@
...
+  if not clipboard or not file then
+    return
+  end
+
+  local is_success, data = pcall(vim.json.encode, clipboard)
+  if not is_success then
+    log.error("Failed to save clipboard. JSON serialization error")
+    return
+  end
+  file:write(data)
```
cseickel, line 25: I actually meant to put the error handling n the `file:write` / `file:flush` code. That's where you are going to get frequent errors due to permissions, file locked, disk full, etc.

lua/neo-tree/sources/filesystem/lib/shared_clipboard.lua
```diff
@@ -0,0 +1,79 @@
...
+      end
+      vim.schedule(function()
+        M._update_all_cilpboards(M._load_clipboard())
+      end)
+    end,
+  })
+
+  -- Using watch_folder because we haven't "watch_file" :)
+  fs_watch.watch_folder(clipboard_state_dir_path, function()
+    if not clipboard_file_change_triggered_by_cur_neovim_instance then
```
cseickel, line 72: This is a bad idea. You can't say for sure that a file event came from this instance just because you wrote a file recently. This is fertile ground for hard to recreate bugs.

Between this and the modified time based check I would rather use the first solution where you just process every write, even your own.

I think I have a better idea:

1. When you convert the clipboard to json, save the json string as `last_clipboard_saved`
2. When you read in a new changed file event, compare the contents of the changed file to the `last_clipboard_saved`
3. If they are equal, nil out the `last_clipboard_saved` and ignore the event.

lua/neo-tree/sources/filesystem/lib/shared_clipboard.lua
```diff
@@ -0,0 +1,79 @@
...
+  end
+  local content = file:read("*a")
+  local is_success, clipboard = pcall(vim.json.decode, content)
+  if not is_success then
+    return nil
+  end
+  return clipboard
+end
+
+M._update_all_cilpboards = function(clipboard)
```
cseickel, line 44: Typo:
```suggestion
M._update_all_clipboards = function(clipboard)
```

lua/neo-tree/sources/filesystem/lib/shared_clipboard.lua
miroshQa: Actually I can't understand why we can't be sure about this. Can you give me an example, please?

1. The variable `clipboard_file_change_triggered_by_cur_neovim_instance` is set in the `save_clipboard` function after writing the clipboard to the file
2. After that an event is immediately triggered in response to the file change.

Why can't we be sure that the write occurred in the current instance? We are not performing any other actions between these two operations, are we?

cseickel: We do not have exclusive access to this folder. Another program could write a file in this folder for some reason. 

For that matter, neo-tree itself might write a file here. It doesn't do that now, but someone else may do so in the future and they likely won't know that some other section of the code is naively reading every file written without checking the contents or even the file name.

Finally, I don't think you should be so confident about the order of events when you are dealing with asynchronous actions.

cseickel:
https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1606

Redoxahmii:
any updates on this ?

loeffel-io:
Please! 🙏

k0va1:
i'd really appreciate to seeing this merged 💪

pynappo:
Closing in favor of #1800

PR #1602, opened 2024-11-16
fix(file-nesting): ensure deterministic rule application

### LINKED ISSUE #1494: BUG: File Nesting sometimes not working.

### Neovim Version (nvim -v)

v0.10.0

macOS Sonoma 14.5 arm64

### Describe the Bug

Hi, When I added large nesting rules tables, then sometimes nesting rules works and sometimes not. I understand that it's happening because of large table. Not sure it's possible to fix...

You can check video: In LazyVim folder, I have rules for readmes. sometime readmes nested sometime not.

https://github.com/nvim-neo-tree/neo-tree.nvim/assets/3833316/abc31823-a006-4d40-94e0-6c809d18efcf

Here is those rules:
```lua
  ['README.*'] = {
    files = { 'AUTHORS', 'BACKERS*', 'CHANGELOG*', 'CITATION*', 'CODE_OF_CONDUCT*', 'CODEOWNERS', 'CONTRIBUTING*', 'CONTRIBUTORS', 'COPYING*', 'CREDITS', 'GOVERNANCE%.MD', 'HISTORY%.MD', 'LICENSE*', 'MAINTAINERS', 'RELEASE_NOTES*', 'SECURITY%.MD', 'SPONSORS*', 'README-*' },
    ignore_case = true,
    pattern = 'README%.(.*)$'
  },
```
If I just added this rule no any other rule, then it works every time. So this issue happening because large table.

Thanks

### Screenshots, Traceback

_No response_

### Steps to Reproduce

1. Install nesting rules 
```lua
-- init.lua
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
if not (vim.uv or vim.loop).fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    "--branch=stable", -- latest stable release
    lazypath,
  })
end
vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
{
    branch = "v3.x",
    dependencies = {
      "nvim-lua/plenary.nvim",
      "nvim-tree/nvim-web-devicons", -- not strictly required, but recommended
      "MunifTanjim/nui.nvim",
      "saifulapm/neotree-file-nesting-config",
    },
    keys = {
      {
        "<C-n>",
        function()
          require("neo-tree.command").execute({ toggle = true })
        end,
        desc = "Explorer NeoTree",
      },
    },
    config = function(_, opts)
      opts.nesting_rules = require('neotree-file-nesting-config').nesting_rules
      require('neo-tree').setup(opts)
    end,
}
})

```
2. `git clone https://github.com/lazyVim/LazyVim/` and open in neovim. Sometime readmes nested sometimes not,

Thanks

### Expected Behavior

It should be nested every time. In the video, you can see what's happening. 

### Your Configuration

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  -- add any other plugins here
}

local neotree_config = {
  dependencies = { "MunifTanjim/nui.nvim", "nvim-tree/nvim-web-devicons", "nvim-lua/plenary.nvim", "saifulapm/neotree-file-nesting-config" },
  cmd = { "Neotree" },
  keys = {
  },
  opts = {
    -- Your config here
    -- ...
  },
  config = function(_, opts)
    opts.nesting_rules = require('neotree-file-nesting-config').nesting_rules
    require('neo-tree').setup(opts)
  end,
}

  root = root .. "/plugins",
})

-- add anything else here
```

saifulapm:
When multiple file nesting rules were configured, rules were being applied inconsistently. For example, sometimes when both `package.json` and `vite.config.*` rules existed, only one set of rules would be applied while the other would be ignored intermittently.

Example configuration that was working inconsistently:
```lua
{
  nesting_rules = require('neotree-file-nesting-config').nesting_rules,
}
```

The fix modifies the nesting logic to:
- Collect all matching rules before applying any nesting
- Use file IDs to prevent duplicate nesting
- Track nested status properly
- Apply rules in a deterministic order

This ensures that all configured nesting rules are applied consistently regardless of their order or complexity.

Fixes #1494


cseickel:
It sounds like a good solution, but I'm afraid I don't have the time to properly evaluate it. See https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1606.

lua/neo-tree/sources/common/file-nesting.lua
```diff
@@ -179,24 +199,39 @@ function M.nest_items(context)
...
-            table.remove(folder.children, index)
+    if files and #files > 0 then
+      table.insert(all_nesting_relationships, {
+        parent = config,
+        children = files,
+      })
+    end
+  end
+
+  -- Then apply them in order
```
pynappo, line 215: if i'm not mistaken, the iteration order of `pairs` is non-deterministic in Lua/LuaJIT, so  there isn't really an order here. As an example, using your nesting rules plugin (nice btw) with something like this directory structure:

image

the file that holds babelrc isn't the same across nvim launches.

It's not too important, this PR is still an improvement over the previous behavior so I'm down to merge this as-is. As a temporary fix, we could iterate over context.nesting using `vim.spairs` (which will iterate in order of the keys, i.e. the absolute filepaths), but I think the easiest route would be a `priority` field for the nesting rules? what would you prefer?
```suggestion
  -- Then apply them (TODO: Determine how to order the rules)
```

pynappo:
Made the rule application actually deterministic. Rules now have an optional `priority` field that works similarly to lazy.nvim's and defaults to 100, where higher priority is run first. if priorities are equal, rules are applied in order of the keys (that is, alphabetically ascending order).

feel free to try out before i merge.

Issue #1067, opened 2023-07-20
BUG: neotree directory change error after auto-closing a tab


Emm54321:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

v0.10.0-dev-679+gb60a2ab4c

### Operating System / Version

debian sid

### Describe the Bug

I get the following error after deleting a buffer in a tab from a TermClose autocmd when autochdir is set:
[Neo-tree ERROR] Error in event handler for event vim_dir_changed[buffers.vim_dir_changed]: ...nvim/lazy/neo-tree.nvim/lua/neo-tree/sources/manager.lua:296: Invalid tabpage id: 2

### Steps to Reproduce

1. mkdir dir && touch file1 && touch dir/file2
1. nvim -u repro.lua
2. :Neotree
3. :edit file1
4. :edit dir/file2
5. :tabnew term://bash
6. enter 'exit' in the shell, which closes its tab due to the autocmd in repro.lua
8. use :bn once or twice to switch between file buffers

### Expected Behavior

No error, directory changed normally.

### Your Configuration

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  -- add any other plugins here
}

local neotree_config = {
  "nvim-neo-tree/neo-tree.nvim",
  dependencies = { "MunifTanjim/nui.nvim", "nvim-tree/nvim-web-devicons", "nvim-lua/plenary.nvim" },
  cmd = { "Neotree" },
  keys = {
    { "<Leader>e", "<Cmd>Neotree<CR>" }, -- change or remove this line if relevant.
  },
  opts = {
    -- Your config here
    -- ...
  },
}

[...]

PR #1604, opened 2024-11-22
fix: remove invalid_tabs before _for_each_state

hrsh7th:
fix https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1067

# problem

Probably, neovim does not fire the `TabClosed` autocmd in some conditions.
So invalid tabs might be remained to the manager states.

# solution

check & dispose invalid states before enumerate it.


cseickel (approved):
I actually came here OK to put an "unmaintained notice" on all PRs, this one is too obvious not to merge.

If you have any interest in maintaining another plugin, neo-tree is up for grabs.

PR #1609, opened 2024-11-26
feat(symlink_target): show relative path if in cwd

elohmeier:
Before:
CleanShot 2024-11-26 at 04 19 33

After:
CleanShot 2024-11-26 at 04 18 39


elohmeier:
Added handling of relative parent dirs:
CleanShot 2024-11-26 at 05 02 38

pynappo:
I think this behavior is probably better than the neo-tree default, but symlinks can be absolute or relative:

image

so I think it'd be better if this was an option. something like:

```lua
symlink_target = {
  enabled = true,
  display_targets = "force_relative_within_cwd" -- this pr (or maybe just "force_relative")
  -- "force_absolute" -- current neo-tree behavior
  -- "auto" -- uses vim.loop.fs_readlink(node.path) to display relative/absolute ones respectively
}
```

Technically I think node.link_to should be derived from fs_readlink instead of fs_realpath as it is right now, but that might be a breaking change.

lua/neo-tree/sources/common/components.lua
```diff
@@ -15,9 +15,54 @@ local utils = require("neo-tree.utils")
...
+          return ""
+        end
+
+        result = result .. table.concat(path_elements, "/", #path_elements - j + 2)
+        return result
+      end
+    end
+
+    result = "../" .. result
+  end
```
pynappo, line 50: This might be a simpler (and technically faster) algorithm, although I haven't tested it much:
```suggestion
    if o_path.path.root ~= ref_path.path.root then
      return o_path:absolute()
    end
    local i, j = #parents, #ref_parents
    -- find the deepest common directory
    while parents[i] == ref_parents[j] and math.min(j, i) > 0 do
      i = i - 1
      j = j - 1
    end
    local common_dir = parents[i + 1]
    local steps_out = string.rep(".." .. sep, j)
    return steps_out .. original_path:sub(#common_dir + 2)
  end
```

lua/neo-tree/sources/common/components.lua
```diff
@@ -15,9 +15,54 @@ local utils = require("neo-tree.utils")
...
+local relative_path = function(original_path, reference_path)
+  -- Use plenary's make_relative to clean paths
+  original_path = path:new(original_path):make_relative(".")
+  reference_path = path:new(reference_path):make_relative(".")
+  local o_path = path:new(original_path)
+  local ref_path = path:new(reference_path)
+  local parents = o_path:parents()
+  local ref_parents = ref_path:parents()
+
+  local path_elements = vim.split(o_path.filename, "/")
```
pynappo, line 32: Windows uses backslashes, so this might not always work. to take that into account:

```suggestion
  local sep = o_path.sep
  local path_elements = vim.split(o_path.filename, sep)
```

lua/neo-tree/sources/common/components.lua
```diff
@@ -15,9 +15,54 @@ local utils = require("neo-tree.utils")
...
 
 local M = {}
 
+-- Workaround for https://github.com/nvim-lua/plenary.nvim/issues/411
+local relative_path = function(original_path, reference_path)
+  -- Use plenary's make_relative to clean paths
+  original_path = path:new(original_path):make_relative(".")
+  reference_path = path:new(reference_path):make_relative(".")
+  local o_path = path:new(original_path)
+  local ref_path = path:new(reference_path)
```
pynappo, line 28: `path` from plenary is capitalized in other modules:
```suggestion
local Path = require("plenary.path")

local M = {}

-- Workaround for https://github.com/nvim-lua/plenary.nvim/issues/411
local relative_path = function(original_path, reference_path)
  -- Use plenary's make_relative to clean paths
  original_path = Path:new(original_path):make_relative(".")
  reference_path = Path:new(reference_path):make_relative(".")
  local o_path = Path:new(original_path)
  local ref_path = Path:new(reference_path)
```

pynappo (changes requested):
Minor style/correctness concerns:

PR #1506, opened 2024-06-26
feat: Add date format string option for `last_modified` and `created` components

switch-blade-stuff:
Implementation of feature request #1505

Issue #1505, opened 2024-06-26
FEATURE: Custom date format string in `last_modified` and `created` components


switch-blade-stuff:
### Did you check the docs?

- [X] I have read all the docs.

### Is your feature request related to a problem? Please describe.

Personally, I prefer using a 24-hour time format, which is currently not possible with neotree.

### Describe the solution you'd like.

A `date_format` string as part of the component config which would be used for `os.date` instead of the hard-coded format string.

pynappo:
Closed by the merged PR linked above

PR #1614, opened 2024-12-07
feat(filesystem): customizable timestamp formats

MastarCheeze:
I've added a `format` field in the configs for the components `last_modified` and `created` so that users can change the format string used to display timestamps in the file explorer. The field also accepts a function that returns a string so that features like relative timestamps can be implemented. The change is also applied to the file details popup window.

Here's an example setup:

```lua
require("neo-tree").setup({
  default_component_configs = {
    last_modified = {
      enabled = true,
      format = function(seconds)
        local now = os.time()
        local diff = now - seconds

        local function pluralise(value, unit)
          return value .. " " .. unit .. (value == 1 and "" or "s") .. " ago"
        end

        if diff < 60 then
          return "Just now"
        elseif diff < 3600 then
          local minutes = math.floor(diff / 60)
          return pluralise(minutes, "minute")
        elseif diff < 86400 then
          local hours = math.floor(diff / 3600)
          return pluralise(hours, "hour")
        elseif diff < 86400 * 30 then
          local days = math.floor(diff / 86400)
          return pluralise(days, "day")
        elseif diff < 86400 * 365 then
          local months = math.floor(diff / (86400 * 30))
          return pluralise(months, "month")
        else
          local years = math.floor(diff / (86400 * 365))
          return pluralise(years, "year")
        end
      end,
    },
    created = {
      enabled = true,
      format = "%e %b %y at %I:%M %p",
    },
  },
})
```

File explorer
File details popup window

The columns are offset in the image because the the timestamp is longer than the column heading. I'll be opening a separate pull request that deals with that.

### DIFF
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -440,7 +440,7 @@ local config = {
         ["."] = "set_root",
         ["[g"] = "prev_git_modified",
         ["]g"] = "next_git_modified",
-        ["i"] = { "show_file_details", config = { created_format = "%Y-%m-%d %I:%M %p", modified_format = "%Y-%m-%d %I:%M %p" } }, -- (see `:h neo-tree-mappings`)
+        ["i"] = "show_file_details", -- see `:h neo-tree-file-actions` for options to customize the window.
         ["o"] = { "show_help", nowait=false, config = { title = "Order by", prefix_key = "o" }},
         ["oc"] = { "order_by_created", nowait = false },
         ["od"] = { "order_by_diagnostics", nowait = false },
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -257,10 +257,14 @@ local config = {
     last_modified = {
       enabled = true,
       required_width = 88, -- min width of window required to show this column
+      format = "%Y-%m-%d %I:%M %p", -- format string for timestamp (see ISO C function `strftime`, `format` parameter)
+                                    -- or use a function that takes in the date in seconds and returns a string
     },
     created = {
       enabled = false,
       required_width = 120, -- min width of window required to show this column
+      format = "%Y-%m-%d %I:%M %p", -- format string for timestamp (see ISO C function `strftime`, `format` parameter)
+                                    -- or use a function that takes in the date in seconds and returns a string
     },
     symlink_target = {
       enabled = false,
--- a/lua/neo-tree/defaults.lua
+++ b/lua/neo-tree/defaults.lua
@@ -258,13 +258,13 @@ local config = {
       enabled = true,
       required_width = 88, -- min width of window required to show this column
       format = "%Y-%m-%d %I:%M %p", -- format string for timestamp (see `:h os.date()`)
-                                    -- or use a function that takes in the date in seconds and returns a string
+                                    -- or use a function that takes in the date in seconds and returns a string to display
     },
     created = {
       enabled = false,
       required_width = 120, -- min width of window required to show this column
       format = "%Y-%m-%d %I:%M %p", -- format string for timestamp (see `:h os.date()`)
-                                    -- or use a function that takes in the date in seconds and returns a string
+                                    -- or use a function that takes in the date in seconds and returns a string to display
     },
     symlink_target = {
       enabled = false,
@@ -440,7 +440,7 @@ local config = {
         ["."] = "set_root",
         ["[g"] = "prev_git_modified",
         ["]g"] = "next_git_modified",
-        ["i"] = "show_file_details",
+        ["i"] = { "show_file_details", created_format = "%Y-%m-%d %I:%M %p", modified_format = "%Y-%m-%d %I:%M %p" }, -- (see `:h neo-tree-mappings`)
         ["o"] = { "show_help", nowait=false, config = { title = "Order by", prefix_key = "o" }},
         ["oc"] = { "order_by_created", nowait = false },
         ["od"] = { "order_by_diagnostics", nowait = false },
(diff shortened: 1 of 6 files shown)


MastarCheeze:
I just found out that there is already an existing issue (#1505) and a PR (#1506) for this exact topic.

lua/neo-tree/defaults.lua
```diff
@@ -257,10 +257,14 @@ local config = {
...
       enabled = true,
       required_width = 88, -- min width of window required to show this column
+      format = "%Y-%m-%d %I:%M %p", -- format string for timestamp (see ISO C function `strftime`, `format` parameter)
+                                    -- or use a function that takes in the date in seconds and returns a string
     },
     created = {
       enabled = false,
       required_width = 120, -- min width of window required to show this column
+      format = "%Y-%m-%d %I:%M %p", -- format string for timestamp (see ISO C function `strftime`, `format` parameter)
+                                    -- or use a function that takes in the date in seconds and returns a string
```
pynappo, line 267: There's builtin help for lua's `os.date()` (which also leads to vimscript's `strftime()`) so that might be a bit more convenient to refer to:
```suggestion
      format = "%Y-%m-%d %I:%M %p", -- format string for timestamp (see `:h os.date()`)
                                    -- or use a function that takes in the date in seconds and returns a string
    },
    created = {
      enabled = false,
      required_width = 120, -- min width of window required to show this column
      format = "%Y-%m-%d %I:%M %p", -- format string for timestamp (see `:h os.date()`)
                                    -- or use a function that takes in the date in seconds and returns a string
```

MastarCheeze: Alright, I'll update the documentation.

lua/neo-tree/sources/common/commands.lua
```diff
@@ -530,9 +530,19 @@ M.show_file_details = function(state)
     table.insert(left, "Size")
     table.insert(right, utils.human_size(stat.size))
     table.insert(left, "Created")
-    table.insert(right, os.date("%Y-%m-%d %I:%M %p", stat.birthtime.sec))
+    local created_format = require("neo-tree").config.default_component_configs.created.format
```
pynappo, line 533: hm, maybe it would be useful if this (and `modified_format`) were to be configurable in the mapping level as well? could be something like:

```suggestion
    local created_format = state.config.created_format or require("neo-tree").config.default_component_configs.created.format
```

Then the user could optionally specify:

```lua
mappings = {
  ['I'] = {
    'show_file_details',
    config = {
      created_format = '...',
      modified_format = '...'
    }
  }
}
```

so the details popup can time differently than the file tree.

It would require more documentation in `defaults.lua`, the vim help doc, and maybe the README though. LMK your thoughts on it.

MastarCheeze: Sounds good!

I think the mappings config should be separate from the default component config to avoid confusion. Changes to the mappings config should only affect the file details popup, and changes to the component config should only affect the info column in the file explorer.

I'll work on an implementation and add in the necessary documentation soon!

MastarCheeze:
I've added the changes you suggested @pynappo.

On second thought, I think its better to use `default_component_configs.created.format` and `default_component_configs.modified.format` as fallback values for the mapping level configs, as you suggested, to reduce the amount of configuration the user has to do. I've implemented this in my commit, along with documentation.

Thanks for the suggestions!

MastarCheeze:
I'm thinking of adding relative timestamps as a builtin feature that can be enabled by specifying:

```lua
default_component_configs = {
  last_modified = {
    relative = true
  }
}
```

Or perhaps like this:

```lua
default_component_configs = {
  last_modified = {
    format = "relative"
  }
}
```

Or even just include a commented-out implementation of a relative timestamp function in `README.md`.

Any opinions about this? Should this be in its own PR?

lua/neo-tree/sources/common/commands.lua
```diff
@@ -530,9 +530,21 @@ M.show_file_details = function(state)
...
     table.insert(right, utils.human_size(stat.size))
     table.insert(left, "Created")
-    table.insert(right, os.date("%Y-%m-%d %I:%M %p", stat.birthtime.sec))
+    local created_format = state.config.created_format
+      or require("neo-tree").config.default_component_configs.created.format
+    if type(created_format) == "function" then
+      table.insert(right, created_format(stat.birthtime.sec))
+    else
+      table.insert(right, os.date(created_format, stat.birthtime.sec))
+    end
```
pynappo, line 539: I think you could pull this logic out into a new file like `sources/common/lib/format.lua` just for code reuse purposes:

```lua
-- sources/common/lib/format.lua
local M = {}
M.date = function(format, seconds)
  -- add logic for special case for format = "relative"
  return type(format) == "function" and format(seconds) or os.date(format, seconds)
end
return M 
```

```suggestion
    local date = require('neotree.sources.common.lib.format').date
    local created_format = state.config.created_format
      or require("neo-tree").config.default_component_configs.created.format
    table.insert(right, date(created_format, stat.birthtime.sec))
    local modified_format = state.config.modified_format
      or require("neo-tree").config.default_component_configs.modified.format
    table.insert(right, date(modified_format, stat.mtime.sec))
```

pynappo (changes requested):

lua/neo-tree/sources/common/commands.lua
MastarCheeze: It would be cleaner to implement relative timestamps in the utils module as you suggested, then add a commented-out setting in `defaults.lua`.

```lua
-- lua/neo-tree/utils/init.lua
M.relative_date = function(seconds)
  -- Implementation
end
```

```lua
-- lua/neo-tree/defaults.lua
...
last_modified = {
  ...
  --format = require("neo-tree.utils").relative_date
},
...
```

What do you think? I'll implement this along with this PR.

pynappo: sounds good to me!

MastarCheeze:
All done! I've added the suggested changes.

MastarCheeze:
Any updates on this @pynappo?

pynappo:
Apologies, had to think about the best way to reduce the amount of configuration needed. I've decided to remove the fallback of file_details's timestamps on the component configs because I think it should always default to giving a detailed absolute time. I also added `format = "relative"` as shorthand.

pynappo (approved):

PR #1615, opened 2024-12-07
feat(filesystem): Customisable column width

MastarCheeze:
I've added a `width` field to the configs so that the width of the columns displayed in the file explorer can be customised. Text that are too long are also truncated with `…` to prevent the columns from being offset.

Edit: `vim.fn.printf` is used instead of `string.format` to prevent an error when the width is more than 99.

Demo:
image

```lua
require("neo-tree").setup({
  default_component_configs = {
    last_modified = {
      width = 10,
    },
    created = {
      enabled = true,
      width = 30,
    },
  },
})
```

This is the fix mentioned in #1614.

### DIFF
--- a/lua/neo-tree/sources/common/components.lua
+++ b/lua/neo-tree/sources/common/components.lua
@@ -457,20 +457,27 @@ M.indent = function(config, node, state)
   return indent
 end
 
+local truncate_string = function(str, max_length)
+  if #str <= max_length then
+    return str
+  end
+  return str:sub(1, max_length - 1) .. "…"
+end
+
 local get_header = function (state, label, size)
   if state.sort and state.sort.label == label then
     local icon = state.sort.direction == 1 and "▲" or "▼"
     size = size - 2
-    return string.format("%" .. size .. "s %s  ", label, icon)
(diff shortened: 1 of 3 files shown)


lua/neo-tree/sources/common/components.lua
```diff
@@ -488,7 +495,7 @@ M.file_size = function (config, node, state)
   end
 
   return {
-    text = string.format("%12s  ", text),
+    text = vim.fn.printf("%" .. config.width .. "s  ", truncate_string(text, config.width)),
```
pynappo, line 498: If the string is truncated to the correct length anyways, maybe simplify this to:

```suggestion
    text = truncate_string(text, config.width) .. "  ",
```

pynappo:
otherwise LGTM, could you add it to the README as well?

MastarCheeze:
I've added the options into the readme, thanks for the suggestion!

pynappo:
ty! pinging @cseickel.

pynappo (approved):

cseickel (approved):

Issue #1617, opened 2024-12-12
BUG: Buffer path with differently-cased drive letter to root directory does not show up in buffers source on Windows


dylangiles:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.10.1 Build type: Release LuaJIT 2.1.1713484068 Run "nvim -V1 -v" for more info

### Operating System / Version

Windows 11 Enterprise 23H2 22631.4460

### Describe the Bug

On Windows, file paths are case-insensitive. When using an LSP integration with Neovim, and using the integration to automatically navigate to the definition of a symbol, Neovim will open a new buffer for the file containing the definition, but sometimes it may open the file using a path that has a lowercase version of the drive letter on which the file resides. Neovim is fine with this due to case-insensitivity of file paths on the Windows platform, but the buffers source in Neo-tree does not display this new buffer, because the root folder and the buffer's path differ in casing of the drive letter. Setting the `bind_to_cwd` option to `false` for the buffers source has no effect.

Linux is not affected - Linux paths are case sensitive, so trying to open a file using the wrong casing won't open the file in the first place. I have not tested macOS.

### Steps to Reproduce

1. Be on Windows (I know - I don't like it either.)
2. Using the init script below, open Neovim in a directory containing more than one file
3. Open one file via Neotree or built-in Neovim functionality, doesn't matter.
4. Open a second file using the `:edit` command, and type the full absolute path to the second file, but change the drive letter to the opposite casing of what you opened Neovim with.
    - Example: If you opened Neovim into `C:\dev\myproject` and you are opening `C:\dev\myproject\mysecondfile.txt`, issue the command `:e c:\dev\myproject\mysecondfile.txt`. 
6. Open the Neotree buffers source (`:Neotree buffers`). **The second file is not visible in the buffers list**.
 

### Expected Behavior

Due to Windows' case-insensitivity rules, I expect the root path and the newly opened buffer's path to be compared in a case-insensitive manner to display correctly in the buffers source.

### Your Configuration

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
[...]

PR #1618, opened 2024-12-12
fix(buffers): Normalize paths for buffers source

dylangiles:
Firstly, for my first contribution to neo-tree, I would like to thank the neo-tree maintainers for producing an such an awesome piece of software. Neo-tree is an integral part of my experience with Neovim as my daily driver, and an amazing example of the wider Neovim ecosystem. I also thank the maintainers for the opportunity to contribute, and welcome any feedback on my contributions.

----------

Paths should be normalized when: 
- Checking their membership in the root directory, and 
- Acquiring their parent paths 

before adding them to the buffers source.

This PR fixes issue #1617.

### DIFF
--- a/lua/neo-tree/sources/buffers/lib/items.lua
+++ b/lua/neo-tree/sources/buffers/lib/items.lua
@@ -70,9 +70,8 @@ M.get_opened_buffers = function(state)
       add_buffer(b, path)
     else
       if #state.path > 1 then
-        local rootsub = path:sub(1, #state.path)
         -- make sure this is within the root path
-        if utils.normalize_path(rootsub) == utils.normalize_path(state.path) then
+        if utils.is_subpath(state.path, path) then
           add_buffer(b, path)
         end
       else
--- a/lua/neo-tree/sources/buffers/lib/items.lua
+++ b/lua/neo-tree/sources/buffers/lib/items.lua
@@ -72,7 +72,7 @@ M.get_opened_buffers = function(state)
       if #state.path > 1 then
         local rootsub = path:sub(1, #state.path)
         -- make sure this is within the root path
-        if rootsub == state.path then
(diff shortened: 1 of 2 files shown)


lua/neo-tree/sources/buffers/lib/items.lua
```diff
@@ -72,7 +72,7 @@ M.get_opened_buffers = function(state)
       if #state.path > 1 then
         local rootsub = path:sub(1, #state.path)
         -- make sure this is within the root path
-        if rootsub == state.path then
+        if utils.normalize_path(rootsub) == utils.normalize_path(state.path) then
```
pynappo, line 75: `utils` has a method for this already which also does the normalization automatically:

```suggestion
        if utils.is_subpath(state.path, path) then
```

(also we can discard the rootsub local with this change)

lua/neo-tree/sources/buffers/lib/items.lua
```diff
@@ -72,7 +72,7 @@ M.get_opened_buffers = function(state)
       if #state.path > 1 then
         local rootsub = path:sub(1, #state.path)
```
pynappo, line 73:
```suggestion
```

See below comment

pynappo (changes requested):

lua/neo-tree/sources/buffers/lib/items.lua
dylangiles: Thank you @pynappo, I've made another commit with these changes

lua/neo-tree/sources/buffers/lib/items.lua
dylangiles: In the same additional commit as above

pynappo (approved):
checked it out on Windows, LGTM @cseickel

cseickel (approved):

PR #1501, opened 2024-06-18
fix(command): allow reveal_file to be used with dir

alanoliveira:
fix https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1500

PR #1620, opened 2024-12-23
Revert "fix(command): allow reveal_file to be used with dir"

pynappo:
Reverts nvim-neo-tree/neo-tree.nvim#1501 due to unintended errors


pynappo:
Actually I can fix this asap

PR #1621, opened 2024-12-23
Revert "fix(command): allow reveal_file to be used with dir (#1501)"

alanoliveira:
This reverts commit 385270ea21a53e06d7fff4730996bd0bb220255d.

https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1501

### DIFF
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -114,7 +114,7 @@ reveal~
 This is a boolean flag. Adding this will make Neotree automatically find and 
 focus the current file when it opens.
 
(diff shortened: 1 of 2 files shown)


pynappo (approved):

pynappo (approved):

pynappo:
I'm thinking the way to implement the behavior you want will with less disruption will actually just be this:
```lua
  local cwd = args.dir or state.path
```

I'll see about it tmrw. for now, thanks for the quick response and help!

Issue #1561, opened 2024-09-16
QUESTION: open relative path


LeonB:
When using the 'open' command the absolute path of the selected file will be opened. Is it possible to configure/customize this behaviour so the relative path to the cwd is used?

vsajko:
`file_opened` event is a one way to solve problem

```
      event_handlers = {
        {
          event = "file_opened",
          handler = function(file_path)
            vim.cmd("cd .")
          end
        },
      }

```

PR #1634, opened 2024-12-30
feat: add open_files_using_relative_paths option

epheien:
fix https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1561

Both NvimTree and NERDTree open files with relative paths by default.


pynappo (approved):
ty!

PR #1635, opened 2024-12-30
feat(fs_actions): file completion for vim.ui.input

parisni:
currently default fs-action (=popup) does not support completion. 
It's possible to fallback on using the neovim native input by setting `use_popups_for_input=false`
However so far file completion is not activated for relevant action such as move, or create.

This PR enable by default the file-completion on native input actions.

Action includes:
- move
- create
- rename
- copy

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/fs_actions.lua
+++ b/lua/neo-tree/sources/filesystem/lib/fs_actions.lua
@@ -557,7 +557,7 @@ M.create_node = function(in_directory, callback, input_root)
       end
       complete()
     end
-  end, {}, setup_file_completion(input_root or base))
+  end, {}, setup_file_completion(input_root))
 end
 
 ---Recursively delete a directory and its children.


lua/neo-tree/sources/filesystem/lib/fs_actions.lua
```diff
@@ -163,7 +163,7 @@ local function get_unused_name(
         local new_path = parent_path and parent_path .. utils.path_separator .. new_name or new_name
         get_unused_name(new_path, using_root_directory, name_chosen_callback)
       end
-    end)
+    end, {}, "file")
```
pynappo, line 166: `file` completion completes relative to the cwd of neo-tree, not relative to the file being renamed. For configs that use `show_path="absolute"` it's fine but the default is to prefill just the filename/dirname. I think you might be able to modify the buffer-local `'path'` option and use `file_in_path` completion to get better completion?

pynappo (changes requested):

lua/neo-tree/sources/filesystem/lib/fs_actions.lua
parisni: Thanks for the suggestion. First of all, I realize file completion for `rename` has no value. So only remains `copy`, `add`, `add_directory` and `move` to be relevant with file completion.

> For configs that use show_path="absolute" it's fine 

It's also perfect for `show_path="relative"`

> but the default is to prefill just the filename/dirname

Yes currently the default does not work. 

> modify the buffer-local 'path' option

This sounds kind of hacky WDYT?
BTW the completion needs `use_popups_for_input=false`, so I wonder if it's worth adding the code you suggest or just activating (and documenting) file completion in case both are meet:
```
use_popups_for_input=false
show_path=relative|absolute
```

An other reason not to support `show_path=none` is most of the action here are relevant on the `cwd`. For eg, `move` has no value if not using absolute or relative path, you ight wan't to move the file at least in the whole project, not only in the current node

Issue #1500, opened 2024-06-18
BUG: reveal ignores dir when called together with reveal_file


alanoliveira:
### Did you check docs and existing issues?

- [X] I have read all the docs.
- [X] I have searched the existing issues.
- [X] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.10.0

### Operating System / Version

Arch Linux 6.9.5-arch1-1

### Describe the Bug

when reveal is called with `dir` and `reveal_file`. The `dir` is ignored unless `reveal_file` is subpath of current neo-tree cwd

### Steps to Reproduce

Using neo-tree repository file-tree as example
1. in the terminal `cd tests`
2. open nvim
3. run `Neotree reveal reveal_file=$PWD/../lua/neo-tree/command/init.lua dir=$PWD/../lua/neo-tree/`

### Expected Behavior

The expected is neo-tree shows this tree
```
neo-tree
├── command
│   ├── completion.lua
│   ├── init.lua
│   └── parser.lua
├── events
├── git
├── setup
├── sources
├── ui
├── utils
├── defaults.lua
├── collections.lua
└── log.lua
```

But it actually shows
```
command
├── completion.lua
├── init.lua
└── parser.lua
```

### Your Configuration

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  -- add any other plugins here
}

local neotree_config = {
  "nvim-neo-tree/neo-tree.nvim",
  dependencies = { "MunifTanjim/nui.nvim", "nvim-tree/nvim-web-devicons", "nvim-lua/plenary.nvim" },
  cmd = { "Neotree" },
  keys = {
    { "<Leader>e", "<Cmd>Neotree<CR>" }, -- change or remove this line if relevant.
  },
  opts = {
[...]

PR #1649, opened 2025-01-03
fix(command): allow reveal_file to be used with dir

alanoliveira:
This PR changes the behaviour of the command when used with `dir` option.
With this change the root/cwd will be set to `dir` if the revealing file is a children of the `dir`, otherwise the root/cwd will be set to the directory of the revealing file.

related PR/issues:
https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1500
https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1501

### DIFF
--- a/lua/neo-tree/command/init.lua
+++ b/lua/neo-tree/command/init.lua
@@ -210,19 +210,26 @@ handle_reveal = function(args, state)
   -- Deal with cwd if we need to
   local cwd = state.path
   local path = args.reveal_file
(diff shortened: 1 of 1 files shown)


pynappo (approved):
updated the PR because yesterday i found an old github issue where cseickel clarified that if dir+reveal are specified together, the behavior should be to always change directory to `dir`. thank you for the help!

PR #1651, opened 2025-01-08
fix(preview): make scroll_preview fallback when inactive

emxxjnm:
scroll_preview fallback default when the preview isn't active. that's don't always hijack the keybind


pynappo (approved):
sorry for the delay!

PR #1653, opened 2025-01-10
fix(preview): stop preview from moving floating tree

### LINKED ISSUE #514: Preview in float changes position of neo-tree
Love this feature. Just wanted to raise a bug where the calling window (e.g. side = right) doesn't get restored on revert_preview (become side = left).

_Originally posted by @axieax in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/502#issuecomment-1244785594_

LionC:
Fixes half of #514 - at least the "moving up" part.

It seems that the added `-1` here was exactly what moved the floating window by 1 each time a preview window was created. Seems to work correctly without it in float and non-float mode.

To check, set neo tree to float and toggle preview on a file several times with and without this fix.


pynappo (approved):
LGTM, thanks! will look at restoring the window position as well.

PR #1655, opened 2025-01-13
fix(utils): handle edge case for `is_subpath`

saytaM12:
I noticed that when I renamed a file in neo-tree some of my opened buffers got changed weirdly. I tracked this to the is_subpath function.
The problem is with the following example
```
/home/user/project/file
/home/user/project/file_for_example
```
These paths would be matched as a correct subpath, even though they aren't.

I added a simple check I believe fixes this issue

### DIFF
--- a/lua/neo-tree/utils/init.lua
+++ b/lua/neo-tree/utils/init.lua
@@ -883,11 +883,12 @@ M.is_subpath = function(base, path)
   base = M.normalize_path(base)
   path = M.normalize_path(path)
   if path:sub(1, #base) == base then
-    local base_parent, base_tail = M.split_path(base)
-    local path_parent, path_tail = M.split_path(path)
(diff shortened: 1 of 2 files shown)


lua/neo-tree/utils/init.lua
```diff
@@ -881,6 +881,9 @@ M.is_subpath = function(base, path)
   end
   base = M.normalize_path(base)
   path = M.normalize_path(path)
+  if base:match("^.*()/") and path:match("^.*()/") then
+    return base == path
+  end
```
pynappo, line 886: This doesn't work, it just turns is_subpath into `base == path`, which is not what the method is supposed to do.

pynappo (changes requested):

pynappo (approved):

PR #1657, opened 2025-01-14
feat(filesystem): add `BEFORE_FILE_{ADD/DELETE}` events

### LINKED ISSUE #1276: Add `BEFORE_FILE_ADD` and `BEFORE_FILE_DELETE` events
### Did you check the docs?

### Is your feature request related to a problem? Please describe.

The events `BEFORE_FILE_MOVE` and `BEFORE_FILE_RENAME` allow us to add support for LSP file operation notifications and queries automatically when performing file moving and renaming. There are actually similar file operations in the LSP spec for file creation and deletion which require events to be triggered before the files are added or deleted. This is currently not possible since neo-tree doesn't emit these events.

### Describe the solution you'd like.

Add `BEFORE_FILE_ADD` and `BEFORE_FILE_DELETE` events similar to the currently existing `BEFORE_FILE_RENAME` and `BEFORE_FILE_MOVE` events to allow the user to add support for the rest of the LSP file operation specification.

### Describe alternatives you've considered.

_No response_

### Additional Context

_No response_

pynappo:
Closes #1276 

Pinging @igorlfs and @mehalter - does anything needs a change? the LSP says only the URIs need to be passed in so I'm just passing the filenames over.


mehalter:
This looks great! One thing I noticed is you are returning immediately after if the event returns that the operation was handled. This seems like a good idea since it gives the user control over the add/delete, but should we also trigger the after events as well just in case the user has added custom events for that?

igorlfs:
Hey there! I think you meant to ping @antosha417, the original author of `nvim-lsp-file-operations`? I only have a fork with some (very) minor changes

pynappo:
> This looks great! One thing I noticed is you are returning immediately after if the event returns that the operation was handled. This seems like a good idea since it gives the user control over the add/delete, but should we also trigger the after events as well just in case the user has added custom events for that?

I think it makes sense, although the other BEFORE events will need to be updated as well.

> Hey there! I think you meant to ping @antosha417, the original author of `nvim-lsp-file-operations`? I only have a fork with some (very) minor changes

Yeah i meant to ping them instead, apologies.

mehalter (approved):
Very excited for this!

PR #1658, opened 2025-01-16
fix(setup): defer merging config until needed

pynappo:
neo-tree has a ton of configuration, so sometimes merging the config table takes a significant portion of startup time. This leads to people lazy-loading neo-tree which sometimes causes issues to be filed, mostly with netrw hijacking. This PR defers the merging until:

- When netrw may need to be hijacked
- `:Neotree` command parsing/execution

On my machine, this reduces startuptime to ~2-3ms when it was previously ~30ms.

Some precautions are taken to prevent breaking people's setups who already lazy-load neo-tree.

Also turned the plugin file to lua because it's easier for me to work with.

Edit: fixed a bug where `source=` cmdline completion wouldn't work.

### DIFF
--- a/lua/neo-tree.lua
+++ b/lua/neo-tree.lua
@@ -1,6 +1,4 @@
 local vim = vim
-local utils = require("neo-tree.utils")
-local log = require("neo-tree.log")
 local M = {}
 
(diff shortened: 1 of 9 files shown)


cseickel (approved):
Nice work

Issue #1678, opened 2025-02-03
BUG: Document symbols - Error when opening  preview


nnajm:
### Did you check docs and existing issues?

- [x] I have read all the docs.
- [x] I have searched the existing issues.
- [x] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.10.4

### Operating System / Version

Linux 6.12.10-arch1-1

### Describe the Bug

When previewing a symbol from the document symbols source, the following error is displayed:

(branch v1.3)

```preview.lua:350: Invalid 'col_end': Expected Lua number```

### Screenshots, Traceback

![Image](https://github.com/user-attachments/assets/cc98fc28-001e-4d52-9817-10846dd85b3c)

### Steps to Reproduce

1. open document symbols view
2. press 'P'

### Expected Behavior

Preview rendered without issues.

### Your Configuration

```Lua
-- template from https://lazy.folke.io/developers#reprolua, feel free to replace if you have your own minimal init.lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    {
      "nvim-neo-tree/neo-tree.nvim",
      branch = "v3.x", -- or "main"
      dependencies = {
        "nvim-lua/plenary.nvim",
        "nvim-tree/nvim-web-devicons", -- not strictly required, but recommended
        "MunifTanjim/nui.nvim",
        -- { "3rd/image.nvim", opts = {} }, -- Optional image support
      },
      opts = {
        -- fill any relevant options here
        sources = {
          'filesystem',
          'document_symbols',
        },
      },
    }
  },
})
vim.g.mapleader = " "
vim.keymap.set("n", "<leader>e", "<Cmd>Neotree<CR>")
vim.keymap.set("n", "<leader>s", "<Cmd>Neotree document_symbols<CR>")
-- do anything else you need to do to reproduce the issue
```

PR #1679, opened 2025-02-03
fix(document_symbols): parse preview range

nnajm:
A proposed fix for issue #1678


pynappo (approved):
Thanks, missed that one. lgtm.

PR #1333, opened 2024-02-02
refactor(type): add type annotation for defaults.lua

pysan3:
Hi ;)

I think I can start working on the big rewrite.

I am going to bring a huge diff to the codebase, so I'd like to ask your opinion first.

# Add Type Annotations

- `.github/workflows/{.luarc.json,lua_ls-typecheck.yml}`
- `lua/neo-tree/defaults.lua`: https://github.com/pysan3/neo-tree.nvim/blob/add-type-annotations/lua/neo-tree/defaults.lua
  - diff: https://github.com/pysan3/neo-tree.nvim/commit/014f98a1a7741f83d6c327c8ebfcc85ed1fe8dab

I've refactored `defaults.lua` to have type annotations for every field in the table.
With a recent change to GitHub's desktop UI, you should be able to jump to references to see the description.

`lua_ls-typecheck.yml` is configured to only check types on `defaults.lua` for now, and I will gradually make the list longer as I add the correct annotations.

# Apply `stylua`

- `.github/workflows/stylua.yml`

This is a github action to format the code automatically.

Changes other than `defaults.lua` are all made by this action (`' -> "` etc), and I'd like to add this action as well.

# Roadmap

@cseickel I've written my roadmap about what I'm planning to do (no due date tho). May I hear your opinion about the list?

- https://github.com/pysan3/neo-tree.nvim/issues/1

This PR is not meant to be merged, but just a POC.
I'd like to fill in `NeotreeTypes.node` and `NeotreeTypes.state` before this gets merged.

Thanks in advance.

PR #1689, opened 2025-02-15
refactor(types): add config types

pynappo:
Basically the same as #1333, but I would rather not modify pysan's branch and they've been pretty inactive so I've redone it by myself.


pynappo:
temporarily closing as i fix the ci

PR #1694, opened 2025-02-19
feat(commands): Add descending order_by_name support

zackwithak13:
Currently, `order_by_name` does not respect sort direction. 
This addition would add support to toggle between ascending & descending name sort.

Potential issues - this does not add a sort order indicator.


pynappo:
> Potential issues - this does not add a sort order indicator.

Well I don't think any of them do, so that's okay.

I think along with this there should probably be an order_by_default command which just sets the sort_field_provider to nil? I might just merge this as-is though, i'll think about it in the morning.

PR #1697, opened 2025-02-20
feat(commands): Add descending order_by_name support

zackwithak13:
Expands #1694 

> Currently, `order_by_name` does not respect sort direction.
This addition would add support to toggle between ascending & descending name sort.

* Updates the state creation with a default sort of `{ label = "Name", direction = 1 }`
  * This prevents needing to send the `order_by_name` command twice to reverse sort order
* Adds a sort icon to the "Name" column when in "window" view
  * Should not apply to the "tab" view


pynappo (approved):
lgtm!

PR #1700, opened 2025-02-23
docs: update show logs comment in defaults.lua

616b2f:
found this small thing when I was trying to debug my custom source.


pynappo:
I was honestly considering reintroducing the command, but having just the lua command should be okay for now and the comment should have been updated regardless. ty!

pynappo (approved):

PR #1705, opened 2025-02-28
feat: configurable indent marker level

and-rs:
The current marker starts after the 2nd node, and this PR modifies the indent function to make the indent marker start from node 0 or 1, also making it configurable. Just like this:

combined

I tried covering some edge cases due to the different combinations from:
```lua
  local indent_size = config.indent_size
  local padding = config.padding
  local marker_start_level = config.marker_start_level
  local level = node.level
  local with_markers = config.with_markers
  local with_expanders = config.with_expanders == nil and file_nesting.is_enabled()
    or config.with_expanders
```

Most of the logic was changed, of course I am open to improving the quality of the code and implement perhaps more edge cases or anything else. Thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ use {
           },
           indent = {
             indent_size = 2,
+            level = 2,
             padding = 1, -- extra padding on left hand side
             -- indent guides
             with_markers = true,
--- a/doc/neo-tree.txt
+++ b/doc/neo-tree.txt
@@ -1035,6 +1035,7 @@ for each source, or just do it once in the default_component_configs section:
           indent_marker = "│",
           last_indent_marker = "└",
           indent_size = 2,
+          level = 2
         },
       },
     })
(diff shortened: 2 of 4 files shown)


and-rs:
Alright, I refactored the code again, and this should handle most cases for indentation options. Perhaps other people trying it would be great.

sand4rt:
It works great on my machine. Got the same behaviour as the screenshots in the PR description. 

Not a showstopper IMO, but it would be nice if `marker_start_level = 0` + `hide_root_node = false`:

image

would behave same as with `marker_start_level = 0` + `hide_root_node = true`:

image

sand4rt:
found something:

image


the indenting / marker looks off when using: `["<"] = "prev_source"` `[">"] = "next_source"`, this is probably the git view?

and-rs:
Mmm yea I just saw that too, I guess I never use other sources, that's a really weird extra padding. Will check that out tomorrow or Tuesday. Thank you for testing it out

sand4rt:
Did you make any more progress? Maybe we should merge it as is?

PR #1735, opened 2025-03-18
fix(filesystem): don't try to close nil dir

AThePeanut4:
In `sync_scan`, if `fs_opendir` errors then it will still try to run `fs_closedir(dir)`, causing an error since `dir` is `nil`. The easiest way to reproduce the issue is to create a folder and remove all its permissions, then try to navigate to it in neo-tree.

I've also removed the `pcall`'s around the uv functions, which are unnecessary since errors are returned rather than being "thrown".


pynappo (approved):
ty!

PR #1765, opened 2025-04-15
fix(filesystem): handle case-insensitive renames on all OSs

iusx:
## error

**helloworld.c** and **Helloworld.c** are recognized as the same file

I found a small issue with neo-tree. For example, if your file is named:

`helloworld.c`

and you want to rename it to:

`Helloworld.c`

you will get the ` warning: already exists, canceling`.

---

Neovim version information:

```
NVIM v0.10.4
Build type: Release
LuaJIT 2.1.1741730670
Run ":verbose version" for more info
```

---

System info:
```
Darwin IUSX 24.1.0 Darwin Kernel Version 24.1.0: Thu Oct 10 21:03:11 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T6020 arm64
```

I know this issue doesn’t exist on Linux, but it does exist on my macOS system. This PR might have some issues, but it works fine on my system and fixes the problem. However, the code might not be very elegant.

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/fs_actions.lua
+++ b/lua/neo-tree/sources/filesystem/lib/fs_actions.lua
@@ -26,7 +26,7 @@ local function rename_is_safe(original_path, destination)
     return true
   end
 
-  if utils.is_windows then
+  if utils.is_windows or utils.is_macos then
     -- check to see if we're just renaming the original to a different case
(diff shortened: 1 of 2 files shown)


iusx (commented):
OMG, it can do that too? That’s amazing

PR #1774, opened 2025-04-24
perf: refresh less

pynappo:
should help neo-tree from getting insanely slow in some cases


zoumi:
Sometimes I have already closed the neo-tree window, but it keeps refreshing in the background.

PR #1779, opened 2025-05-02
fix(filesystem): prevent infinite loop when scanning a directory containing entries more than ENTRIES_BATCH_SIZE

Honey-Teio:
`sync_scan` function causes infinite loop when scanning a directory containing entries more than `ENTRIES_BATCH_SIZE` in `shallow` scan mode.
This PR fixes that issue.


pynappo:
oh i forgot to check previous PRs before fixing this on my own (#1784), so sorry. thank you for the effort though!

PR #1788, opened 2025-05-10
feat(fuzzy_finder): normal mode navigation, normalize mappings

pynappo:


fivemservereyestore:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

pynappo:
gonna split this up actually

Issue #1682, opened 2025-02-04
BUG: Deleted whole workspace while attempting to delete selection


SichangHe:
### Did you check docs and existing issues?

- [x] I have read all the docs.
- [x] I have searched the existing issues.
- [x] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.10.2

### Operating System / Version

#140~20.04.1-Ubuntu

### Describe the Bug

I selected two files in NeoTree using `v` `k`, pressed `d` and `y`, then it lagged a bit and I lost the whole workspace.

A little prior, I pressed `i` to check the sizes of those two files. A while prior, I cut-pasted some files to another directory.

Different from #1174 and #1472, I was not deleting the file I was working on; I did not even open the files I selected for deletion (if I recall correctly).

Commit for NeoTree used: 521bb4253b1adc2a1ff6c584bc839eba9bc99c0b

I will never use NeoTree to modify any files again! All my data for the project are gone.

### Steps to Reproduce

1. Use Neovim, NeoTree for a while (this I do not how to reproduce).
2. Select two files in NeoTree using visual mode and press `d` and `y` to delete them.

### Expected Behavior

Only the two selected files are deleted.

### Your Configuration

```Lua
-- My configuration is at <https://github.com/SichangHe/.config/tree/2d879868c1fb408e4f9870021747d2400f83670d>.
-- I have not tried to reproduce losing a month of work.
```

immdevrov:
Same stuff happend to me

pynappo:
@immdevrov was this also with visual deletion?

i've tried to repro this a few times in the past but no luck. i'll add some extra prompts/checks.

in the next breaking ver delete should probably be trash by default instead

immdevrov:
@pynappo 
Yes, something along quick up-down, visual selecting and deleting. I think I typed it out of vim-habit. Got very frustrated, deleted the plugin and installed oil.nvim :D
I guess, me trying vim commands inside a file tree already means that I need to use oil, not a file tree

PR #1793, opened 2025-05-24
fix(filesystem): more paranoid visual deletion

pynappo:
re: #1682

maybe getting selected nodes could be messing up

todo: more checks


pynappo:
> It would be good indicator also to have the number of files/dirs that are going to be deleted 🤔

there's already a confirmation of how many nodes you select for visual selection but i could see about doing a full traversal of the entire node tree ig

PR #1832, opened 2025-08-13
docs(README): fix mini.deps instructions

gps08:
Fix checkout property in mini.deps install snippet


pynappo:
thanks for catching that haha

PR #1956, opened 2025-12-26
fix(file-items): ensure parent directories have children table

junqizhang:
## Description

Fixes a crash in the git_status source when processing files in directories with spaces in their paths.

## Problem

When opening git_status view with files like:
```
/repo/learning matierial classes/class6_heap_BFS/file.py
```

Neo-tree crashes with:
```
Error creating item for <path>: bad argument #1 to 'insert' (table expected, got nil)
```

## Root Cause

The issue is a **caching problem** in `create_item()`:

1. When a parent directory is first created elsewhere without specifying the type, `create_item()` attempts type detection via `uv.fs_lstat()`
2. If detection fails (e.g., path doesn't exist yet during git status parsing), the type becomes "unknown" instead of "directory"
3. Without type="directory", `item.children` is never initialized
4. The item gets cached in `context.folders[path]`
5. Later, when `set_parents()` calls `create_item(context, parent_path, "directory")`, it returns the cached item (which lacks `children`)
6. Attempting `table.insert(parent.children, item)` crashes because `parent.children` is nil

## Solution

1. **Patch cached items** (line 175-179 in file-items.lua):
   - When returning a cached item requested as "directory", ensure it has a `children` table
   - Initialize `children = {}` if missing

2. **Pass explicit type in set_parents()** (line 304):
   - Call `create_item(context, item.parent_path, "directory")` instead of omitting the type
   - Ensures newly created parents are properly initialized

3. **Add safety guards**:
   - Early return when parent creation fails (line 307)
   - Safety check before insertion (line 313-316)

## Testing

Tested with repository containing directories with spaces:
- ✅ Git status source displays files correctly
- ✅ No crashes when opening git_status view  
- ✅ Files can be staged/unstaged normally
- ✅ Parent directories have proper `children` tables

## Notes

- The crash occurs specifically when directories with spaces are involved because git status parsing creates parent directory items on-the-fly
- The fix is general and improves robustness for all parent directory creation scenarios
- Bug introduced after git refactor in #1892 (Dec 4, 2025)

### DIFF
--- a/lua/neo-tree/sources/common/file-items.lua
+++ b/lua/neo-tree/sources/common/file-items.lua
@@ -172,11 +172,6 @@ local function create_item(context, path, _type, bufnr)
     -- avoid creating duplicate items
     local cached_item = context.folders[id] or context.nesting[id] or context.item_exists[id]
     if cached_item then
-      -- If requesting as directory but cached item lacks children, initialize it
-      if _type == "directory" and not cached_item.children then
-        cached_item.children = {}
-        cached_item.type = "directory"
-      end
       return cached_item
     end
   end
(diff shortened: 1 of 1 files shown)


pynappo:
Looked a bit further and wasn't a caching issue so i changed the implementation, but ty!

Issue #724, opened 2023-02-02
Watching for git changes is broken


cseickel:
That does not seem to work for me. I've put the log level to trace and I can see the git watcher has started, but after that, nothing happens when I commit my files from a terminal.

_Originally posted by @primalmotion in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/337#issuecomment-1411257704_

cseickel:
I think this problem only applies if your root is higher up the tree than the repo of the file you are working with. If the root of neo-tree is the same as the repo, this works as expected. 

Can you confirm that this is the case for you as well?

It should still work regardless of what you rroot is, I just want to confirm we are talking about the same problem.

primalmotion:
Hmm no, I open nvim /a/b/c and the git root is at /a/b/c/.git, and that doesn't work

primalmotion:
Actually that's a problem too..

- With git_status_async=true: it never works
- With git_status_async=false: it works when neotree root == git root, but, as you mentioned not when neotree root is higher up in the hierarchy thab git root

austinbutler:
Doesn't seem to work for me even with async disabled.

```lua
	require("neo-tree").setup({
		git_status_async = false,
		filesystem = {
			use_libuv_file_watcher = true,
		},
	})
```

cseickel:
@Hubro the `git_status` source of neo-tree is already subscribing to that Fugitive event and refreshing on its own. If you are using Fugitive to commit, it should work without you having to do anything.

This is really a separate issue and you should open a new issue with a minimal init script and instructions on how to recreate your issue.

This issue is specifically for the `filesystem` source and NOT for when using Fugitive, which AFAIK works perfectly for the `git_status` source. The only situation where the Fugitive event is not handled for the `filesystem` source is if you have an older style `before_render` event handler in your `filesystem` config.

Hubro:
@cseickel You're right, the git_status source actually updates fine when I'm running Fugitive commands in the same tab as neo-tree, so I don't actually need the `FugitiveChanged` autocommand.

The problem seems to be that I always run Fugitive in its own tab (`:tab G`), so any `FugitiveChanged` events that happens are ignored by neo-tree, since the refresh function only applies to neo-tree instances in the current tab. Refreshing on `TabEnter` fixes that.

cseickel:
> the refresh function only applies to neo-tree instances in the current tab

[...]

PR #1965, opened 2026-01-11
fix(git): upkeep watcher refs, enable watch for sync git_status

xieyonn:
close(https://github.com/nvim-neo-tree/neo-tree.nvim/issues/724)

1. After each render, there will do `watcher.references - 1`. When `watcher.references == 0`, the watcher will stop. Therefore, every time a git status refresh is triggered, watch_folder need to be triggered to do `watcher.references + 1`.

2. If set `filesystem.use_libuv_file_watcher = true`, it means that when the git state of the current repository is modified outside of neovim—for example, by using the git command line, any change to the `.git` directory will trigger `Neotree` to refresh git-status. This is unrelated to whether `git_status_async` is enabled.


pynappo:
merging this for now b/c it should improve watching reliability, although i think the core issue is that watching system is generally being too eager to decrement references

PR #1966, opened 2026-01-11
fix(git): process GIT_EVENT for git_status_async again

xieyonn:
Found the boundary condition in the previous commit.A corner case for issue: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/724

Handle event `GIT_EVENT` in `enable_git_status = true` both when use git_status_sync/git_status_async. I’m not sure why this check was added, but it works after removing it.

### DIFF
--- a/lua/neo-tree/sources/filesystem/init.lua
+++ b/lua/neo-tree/sources/filesystem/init.lua
@@ -402,12 +402,10 @@ M.setup = function(config, global_config)
 
   -- Respond to git events from git_status source or Fugitive
   if global_config.enable_git_status then
-    if not global_config.git_status_async then
-      manager.subscribe(M.name, {
-        event = events.GIT_EVENT,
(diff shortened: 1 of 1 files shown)


xieyonn:
Just setting `filesystem.use_libuv_file_watcher = true`, then, do git commands (add/commit ...) outside neovim, will refresh git status successfully.

xieyonn:
This check was added in: https://github.com/nvim-neo-tree/neo-tree.nvim/pull/1958

pynappo:
i think i added the check out of caution to prevent infinite calls to status but on second thought an infinite loop isn't possible in this case

an infinite loop was possible when git.status() would fire git_status_changed which would cause the tree to re-render and run git.status() again but that's been fixed w/ the output caching. git_event does not suffer from the same root issue

Issue #1984, opened 2026-02-17
action="show" steals focus when creating neo-tree window


jensenojs:
Happy Chinese New Year!

---

## Problem

When using `action="show"`, neo-tree's window creation temporarily moves focus to the neo-tree window via `nvim_set_current_win()`, even though the user expects focus to remain in the editor.

## Videos

Clearly, having a video comparison is better

https://github.com/user-attachments/assets/605dd23c-1d82-4be2-8fc5-28766a405cd4

## Impact

This causes two types of issues:

### 1. WinLeave Side Effects (All Sources)

When neo-tree creates its window with `action="show"`, the momentary focus switch triggers `WinLeave` on the current window. This can unintentionally close third-party floating windows that listen to `WinLeave` for auto-dismiss behavior (e.g., yazi.nvim, which closes its file picker on `WinLeave`).

**Reproduction:**

```bash
cd /path/to/neo-tree.nvim
nvim --clean -u repro_winleave.lua
```

On `main` branch: The test float window disappears when neo-tree opens.

**Real-world example:** yazi.nvim registers a `WinLeave` autocmd that calls `self:close()` immediately ([source](https://github.com/mikavilpas/yazi.nvim/blob/main/lua/yazi/window.lua#L139-L144)). Any `nvim_set_current_win()` that moves focus away from yazi's float will trigger this autocmd and close the picker, even if the focus is restored milliseconds later.

### 2. Focus Loss with document_symbols (Async LSP)

When using `action="show" source=document_symbols"`, focus is stolen and **never restored**.

**Root cause:** document_symbols' `navigate()` depends on an async LSP request. Upstream's focus-restore callback is scheduled via `vim.schedule()` and executes **before** the LSP response arrives. When the LSP response eventually triggers window creation (via `NuiSplit:mount()`), the callback has already been consumed and cannot restore focus.

**Timeline:**
```
T1: action="show" saves current_win, calls navigate()
T2: navigate() fires LSP request (async), schedules callback, returns immediately
T3: Next event loop tick: callback executes nvim_set_current_win(current_win)
    → Focus is already in current_win, this is a no-op
    → Callback lifecycle ends
T4: LSP response arrives (50-500ms later): NuiSplit:mount() steals focus
    → No code exists to restore focus
```

**Reproduction:**
```bash
cd /path/to/neo-tree.nvim
nvim --clean -u repro_focus.lua
```

[...]

PR #1985, opened 2026-02-17
fix(command): prevent focus stealing in action="show"

jensenojs:
When action="show", use nvim_open_win(enter=false) instead of NuiSplit to avoid moving focus. This fixes two issues:
- WinLeave side effects (closes third-party floats like yazi.nvim)
- document_symbols async LSP leaving focus in symbol panel

Requires nvim 0.10+. Older versions fall back to NuiSplit as previous

for

- https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1984

### DIFF
--- a/lua/neo-tree/command/init.lua
+++ b/lua/neo-tree/command/init.lua
@@ -36,6 +36,7 @@ local function do_show_or_focus(args, state, force_navigate)
       return
     end
     -- close_other_sources()
+    state._no_focus = true
     local current_win = vim.api.nvim_get_current_win()
     manager.navigate(state, args.dir, args.reveal_file, function()
       -- navigate changes the window to neo-tree, so just quickly hop back to the original window
--- a/lua/neo-tree/command/init.lua
+++ b/lua/neo-tree/command/init.lua
@@ -39,10 +39,14 @@ local function do_show_or_focus(args, state, force_navigate)
     -- Prevent focus stealing: use nvim_open_win(enter=false) in acquire_window.
     -- This avoids WinEnter/WinLeave side effects (e.g. closing yazi.nvim floats).
     state._no_focus = true
-    local current_win = vim.api.nvim_get_current_win()
+    local previous_win = vim.api.nvim_get_current_win()
     manager.navigate(state, args.dir, args.reveal_file, function()
       -- navigate changes the window to neo-tree, so just quickly hop back to the original window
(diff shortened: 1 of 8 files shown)


jensenojs:
This PR prevents neo-tree from briefly stealing focus when opened with action="show". The key problem: even a split-second focus change fires WinLeave on the original window, which can destroy third-party floating windows (e.g. yazi.nvim, telescope, noice popups).

Upstream recently added a callback that jumps back to the original window after opening — but by then WinEnter/WinLeave have already fired, so the damage is done. This PR stops focus from moving in the first place: state._no_focus tells acquire_window to use nvim_open_win(enter=false), creating the split without ever touching the original window's focus.
The upstream callback is kept as a harmless fallback for Neovim < 0.10 (where nvim_open_win lacks the split parameter).

pynappo:
The PR implementation does not use `nvim_open_win` correctly, the split is created relative to the current window instead of the editor. for example, vsplit a buffer and call Neotree show when focusing the right-side window and neo-tree appears in the wrong spot.

also currently neo-tree relies on buffer enter to set various options correctly for the window/buffer (see buffer_enter_event in setup). currently options are not set correctly in the neotree window until it is focused.

this probably requires a bigger refactor of how neo-tree's buffer/window options are set. i'll keep this open for now but will probably make my own pr making all the required changes

jensenojs:
Thanks for jumping in and fixing the split position! Really appreciate all the time and effort you put into maintaining this project.

pynappo (approved):

Frederick888:
After this PR I have an issue where neo-tree side window becomes wider after vertically resizing Neovim.

Steps to reproduce:

1. `nvim some_file`
2. `:Neotree reveal show`
3. Resize the Neovim vertically

```sh
$ nvim --version
NVIM v0.12.2
Build type: Release
LuaJIT 2.1.1780076327
Run "nvim -V1 -v" for more info
```

My neo-tree config is at https://git.tsundere.moe/Frederick888/frederick-settings/-/blob/master/.config/nvim/lua/plugins/tree.lua. I can find some time to do a proper report with a minimal init.lua if others can't reproduce this issue 🙏

pynappo:
hm weird thing is that I can reproduce that in lazyvim with that spec but not if it's just the given config... will take a closer look and fix asap

pynappo:
forgot to implement the window options nuisplit implicitly adds by default, should be fixed

PR #2004, opened 2026-03-30
feat(commands): quick jump

stupidbullet:
Command `quick_jump` can dynamically compute, assign and render hotkeys for jumping to a specific node, while also monitoring key inputs to handle jump/toggle/cancel actions.

A default leader key for this function `;` has been added in `defaults.lua` now, but no description has been written in doc.

When neo-tree is focused, pressing `;` re-renders all icons and displays maps that can be used to jump to/toggle nodes.

Besides, it is possible to bind this action to a global hotkey, for example:
```lua
    local manager = require("neo-tree.sources.manager")
    local commands = require("neo-tree.sources.common.commands")

    for _, win in ipairs(vim.api.nvim_tabpage_list_wins(0)) do
      local state = manager.get_state_for_window(win)
      if state and state.tree then
        vim.api.nvim_set_current_win(win)
        commands.quick_jump(state)
        return
      end
    end
```
This searches for neo-tree window, focuses it, then triggers the function.
It may be helpful to embed it as actions (like focus, show).


lua/neo-tree/sources/common/commands.lua
```diff
@@ -992,6 +992,125 @@ M.vsplit_with_window_picker = function(state, toggle_directory)
...
+        elseif target_node.type == "terminal" then
+          utils.open_file(state, target_node:get_id(), "e", target_node.extra and target_node.extra.bufnr)
+        end
+      elseif state.name == "git_status" then
+        if target_node.type == "file" then
+          utils.open_file(state, target_node.path, "e", target_node.extra and target_node.extra.bufnr)
+        elseif target_node.type == "directory" then
+          toggle_directory(target_node)
+        end
+      end
```
pynappo: I think the default expectation for a jump command is that it only does the jumping, not also open/toggle the destination node by default.

One possible remedy is to just replace the whole section with a customizable callback so then you and others can do whatever they want post-jump:

```suggestion
	local callback = state.config and state.config.callback
	if type(state.config.callback) == "function" then
	  callback(state, target_node)
	end
```

(or instead of 'callback', something like 'on_jump' or 'on_destination' is probably more descriptive)

lua/neo-tree/utils/init.lua
```diff
@@ -1729,6 +1729,184 @@ M.truncate_by_cell = function(str, col_limit, align)
...
 end
 
+-- A helper generates double-character hotkeys table.
+-- Hotkeys are classified by the first character, and the inner priority uses qwerty distance.
+-- That is, the closer two characters are, the higher their priority.
+-- If two candidates have the same distance, we break ties in the order:
+-- east -> north -> south -> west -> ne -> se -> nw -> sw.
+-- Example: hh -> hj -> hy -> hn -> hg -> hu -> hb -> ht -> hm -> hk -> ... -> hz.
+---@return table<string, string[]> tbl
+local generate_hotkeys = function()
```
pynappo: I appreciate the effort, but i think having a constant string of jump labels, like most jumping plugins, is probably a simpler approach and allows for easy customization (esp for non-qwerty users). like:

```lua
local a2z = "abcdefghijklmnopqrstuvwxyz"
state.config.jump_labels = a2z
```

(the default labels could also be ordered to favor qwerty home row, like "jfkdlsa..." by default, like what https://github.com/rlane/pounce.nvim has, up to you)

The iteration of labels could be ({letter}{letter}, {letter}{jump_labels[1]}, {letter}{jump_labels[2]}, and so on)

lua/neo-tree/utils/init.lua
```diff
@@ -1729,6 +1729,184 @@ M.truncate_by_cell = function(str, col_limit, align)
...
+      end
+    end
+  end
+  return candidate
+end
+
+-- Generate bimap node -> hotkey & hotkey -> node.
+---@param name_nodes table<string, { b: boolean, node: neotree.FileNode }>
+---@return table<neotree.FileNode, string> node2key
+M.assign_hotkeys = function(name_nodes)
```
pynappo: This should fallback to other unused labels if all the labels for a given starting character are exhausted. Say you have a directory with a bunch of files whose first starting character is identical (say, a directory of daily logs, generated by this llm-generated script)

```bash
start_date="2025-12-01"
end_date="2026-03-01"

# Create a directory for the logs and move into it
mkdir -p daily_logs
cd daily_logs || exit

current_date="$start_date"

echo "Generating logs from $start_date to $end_date..."

# Loop until current_date exceeds end_date
while [[ "$current_date" < "$end_date" ]] || [[ "$current_date" == "$end_date" ]]; do
    # Create the empty log file
    touch "${current_date}.log"
    
    # Increment the date by 1 day using the 'date' command
    # This works on most modern Linux (GNU) systems
    current_date=$(date -I -d "$current_date + 1 day")
done
```

Currently trying to quick jump within this directory will lead to a lot of entries not receiving jump labels and thus not being jumpable. The mechanism for fixing this is up to you. The first option that comes to mind is to keep a running list of entries starting with first characters that exhaust their entries, then assign them unused labels generated from the aforementioned jump_labels option (like jump_labels[1]}{jump_labels[2]}, {jump_labels[1]}{jump_labels[2]}, {jump_labels[1]}{jump_labels[3]}, etc)

pynappo (changes requested):
Cool idea. not totally sure if it has to be in neo-tree but will suggest improvements nonetheless

stupidbullet:
Thanks a lot for your helpful advice, especially for someone new to plugin development :-)

I had barely considered customizability, but this part has been added into the code in the newest commit.

Now it is possible to write different post-jump callback for each sources, as well as jump labels.
However, I decided to keep the builtin open-or-toggle node function, it was my original intention for writing this feature.
Besides, perhaps it is the key to distinguishing the built-in jump from other jump plugins. Of course, we can also use a slightly more complex function to implement the same callback with other plugins (as long as they support it), but having a slightly more advanced built-in advanced function might be a reason people choose to use it.

Rendering is also a point. Most jump plugins attends to overlap its hotkeys among the original word, and it is sometimes inconvenient that I cannot read the filename. Replacing the icon was easily implemented internally, and it perfectly solved this problem.

stupidbullet:
Also, hotkeys will no longer be exhausted now.

lua/neo-tree/sources/common/commands.lua
```diff
@@ -992,6 +992,95 @@ M.vsplit_with_window_picker = function(state, toggle_directory)
...
+      node2key = candidate
+      depth = depth + 1
+      redraw_jump()
+    else
+      local target_node = next(candidate)
+      if target_node == nil then
+        break
+      end
+
+      local on_jump = state.on_jump or state.config.on_jump
```
pynappo: not much need for on_jump config to be linked to the state itself
```suggestion
      local on_jump = state.config.on_jump
```

lua/neo-tree/defaults.lua
```diff
@@ -411,6 +411,7 @@ local config = {
       nowait = true,
     },
     mappings = {
+      [";"] = { "quick_jump" },
```
pynappo: config should be here instead of in the source configs

Also on second thought, i know [some people rebind ; to :](https://github.com/search?q=%22%27%3B%27%2C+%27%3A%27%22+language%3ALua+path%3Anvim&type=code) so to be slightly less disruptive to those users it might just be better to bind this to like `<C-s>` or `gs` by default? (just looked at other mappings used by leap.nvim/flash.nvim)

```suggestion
      ["<C-s>"] = {
        "quick_jump"
        config = {
          -- This will automaticly open / toggle the target node after jumping.
          -- You can set it to `nil` to perform only the jump action,
          -- or write your own callback function.
          on_jump = "open_or_toggle",
          jump_labels = "jfkdlsahgnuvrbytmiceoxwpqz",
        }
      },
```

lua/neo-tree/utils/init.lua
```diff
@@ -1729,6 +1729,170 @@ M.truncate_by_cell = function(str, col_limit, align)
...
+
+  return "j"
+end
+
+-- Return key - node pairs whose key starts by ch.
+---@param node2key table<neotree.FileNode, string>
+---@param ch string
+---@param depth integer
+---@return table<neotree.FileNode, string> candidate
+M.get_candidate = function(node2key, ch, depth)
```
pynappo: These seem too highly specific to belong to utils. I would keep all of these functions in sources/common/commands.lua or somewhere like sources/common/lib/jump.lua if you want a separate file

pynappo (changes requested):

lua/neo-tree/defaults.lua
stupidbullet: It seems fine. I prefer <C-s> since it simply takes less keystrokes.

lua/neo-tree/utils/init.lua
stupidbullet: Perhaps a separate file is better, as commands is long enough now and these helpers also are not used anywhere else.

pynappo (approved):

PR #2017, opened 2026-04-19
fix(command): Focus neo-tree window when toggling to a different source

### LINKED ISSUE #2016: BUG: `:Neotree toggle/reveal` don't focus neo-tree window when switching source

- [x] I have read all the docs.
- [x] I have searched the existing issues.
- [x] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.12.1

Windows 11 25H2 / Ubuntu 24.04.4 LTS

### Describe the Bug

When Neotree is open in a window with source A (e.g. `buffers`) and the cursor is currently in another window, a `:Neotree toggle filesystem` does switch the `source to `filesystem`, but does not bring that window in focus (i.e. move the cursor there).

The same is true for `:Neotree reveal`.

---

The following change fixes it for me (with all tests still passing and based on my testing other behaviour staying as expected) and I *can* create a PR with that change, but I am not entirely sure whether `manager.navigate` is actually expected to always navigate to `state.winid` (a call [a few lines above](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/4d0828d95adaf4250c1373d2cd9318d3509712b7/lua/neo-tree/command/init.lua#L41-L42) somewhat indicates that).

```diff
diff --git a/lua/neo-tree/command/init.lua b/lua/neo-tree/command/init.lua
index 3c3145a..0a7a688 100644
--- a/lua/neo-tree/command/init.lua
+++ b/lua/neo-tree/command/init.lua
@@ -48,7 +48,11 @@ local function do_show_or_focus(args, state, force_navigate)
     end
     if force_navigate or not window_exists then
       -- close_other_sources()
-      manager.navigate(state, args.dir, args.reveal_file, nil, false)
+      manager.navigate(state, args.dir, args.reveal_file, function()
+        if state.winid and vim.api.nvim_win_is_valid(state.winid) then
+          vim.api.nvim_set_current_win(state.winid)
+        end
+      end, false)
     end
   end
 end
```

### Screenshots, Traceback

_No response_

### Steps to Reproduce

1. Open Neovim (config see below)
2. `:Neotree reveal filesystem`
3. `<C-w>l` (move to another window)
4. `:Neotree toggle buffers` or `:Neotree reveal buffers` (the buffer displays the buffers, but it the cursor is not in the neotree window)

### Expected Behavior

The Neotree buffer gets focus

### Your Configuration

```Lua
-- template from https://lazy.folke.io/developers#reprolua, feel free to replace if you have your own minimal init.lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    {
      branch = "main"
      dependencies = {
        "nvim-lua/plenary.nvim",
        "nvim-tree/nvim-web-devicons", -- not strictly required, but recommended
        "MunifTanjim/nui.nvim",
      },
      lazy = false,
      ---@module "neo-tree"
      ---@type neotree.Config?
      opts = {},
    }
  },
})
vim.g.mapleader = " "
vim.keymap.set("n", "<leader>e", "<Cmd>Neotree<CR>")
-- do anything else you need to do to reproduce the issue
```

Wikiwix:
When neo-tree is already open (e.g. with the filesystem source) and the cursor is in another window, running `:Neotree toggle <other_source>` (e.g. `:Neotree toggle buffers`) would switch the panel's content but leave the cursor in the previous window.

Fix #2016

### DIFF
--- a/lua/neo-tree/command/init.lua
+++ b/lua/neo-tree/command/init.lua
@@ -48,7 +48,11 @@ local function do_show_or_focus(args, state, force_navigate)
     end
     if force_navigate or not window_exists then
       -- close_other_sources()
-      manager.navigate(state, args.dir, args.reveal_file, nil, false)
+      manager.navigate(state, args.dir, args.reveal_file, function()
+        if state.winid and vim.api.nvim_win_is_valid(state.winid) then
+          vim.api.nvim_set_current_win(state.winid)
+        end
+      end, false)
     end
   end
 end


pynappo:
On second thought I'm seeing that this has been Neo-tree's behavior since the 3.0 tag, so a bit concerned that this could break someone's workflow. Will hold off on merging until 4.0 when i can change a couple defaults/systems (next few weeks, i hope?)

Overall still believe this change is correct though because it seems the least surprising that `:Neotree focus filesystem` will always focus. if people want the old behavior they can either use `show` instead of focus, or we could introduce a new (default?) action like `switch_or_focus` which has the current behavior: switch the source if it's not at the given position, otherwise focus (i can cover in a future pr)

Wikiwix:
> On second thought I'm seeing that this has been Neo-tree's behavior since the 3.0 tag, so a bit concerned that this could break someone's workflow. Will hold off on merging until 4.0 when i can change a couple defaults/systems (next few weeks, i hope?)

Sounds reasonable :)

> Overall still believe this change is correct though because it seems the least surprising that `:Neotree focus filesystem` will always focus. if people want the old behavior they can either use `show` instead of focus, or we could introduce a new (default?) action like `switch_or_focus` which has the current behavior: switch the source if it's not at the given position, otherwise focus (i can cover in a future pr)

I personally would not add a new action preemptively:
What would be the desired *outcome* when calling that action? The action's effect depends so much on Neovim's state that I don't think one can train useful muscle memory around it.

On top of that:
- I don't see the use case and two other users (whom I asked to confirm that I ran into a Neotree bug) also felt like this must be a bug.
- If users actually desire the `switch_or_focus` action, they can always define their own command.
- **If** there is a "typical" use for it, someone will raise an issue and it can still be added later.

PR #2020, opened 2026-04-23
Add tab targeting to command.execute

fedoraiver:
## Summary

This PR adds optional `tabid` targeting to `require("neo-tree.command").execute(...)` for sidebar positions.

When `tabid` is provided, the command runs against that tab's neo-tree state while restoring the caller's original tab/window context afterwards. This makes it possible for external plugins or local config to synchronize neo-tree open/close state across tabs without relying on user-visible tab switches.

## Why

Issue #1235 discusses shared neo-tree behavior across tabs. A full shared-instance implementation is a much larger state-model change. This PR takes the smaller step already suggested in that thread: expose an official way to target another tab through `command.execute`, so higher-level sync behavior can be built on top of the public API instead of config-local hacks.

## What changed

- add `tabid` support to `command.execute(...)` for sidebar positions
- preserve the caller's current tab/window after running commands against another tab
- guard `show` so it only restores focus when the original window is still valid in the same tab
- add coverage for showing and closing neo-tree in another tab without changing the current context

## Validation

- `nvim --headless --noplugin -u tests/mininit.lua -c "lua require('plenary.busted').run('tests/neo-tree/command/command_tabid_spec.lua')"`
- `nvim --headless --noplugin -u tests/mininit.lua -c "lua require('plenary.test_harness').test_directory('tests/neo-tree/command', { minimal_init = 'tests/mininit.lua', sequential = true })"`

Refs #1235.

### DIFF
--- a/lua/neo-tree/command/init.lua
+++ b/lua/neo-tree/command/init.lua
@@ -16,6 +16,40 @@ M._last = {
   position = nil,
 }
 
+---@generic T
+---@param tabid integer?
+---@param callback fun(): T
+---@return T
+local function with_tab_context(tabid, callback)
+  if not tabid or tabid == vim.api.nvim_get_current_tabpage() then
+    return callback()
+  end
+
+  if not vim.api.nvim_tabpage_is_valid(tabid) then
+    error("Invalid tabid: " .. tostring(tabid))
+  end
+
+  local original_tabid = vim.api.nvim_get_current_tabpage()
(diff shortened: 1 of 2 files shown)


pynappo:
Not reviewing this yet but just noting that i'd prefer `with_tab_context` to be renamed to `tab_call` or something, since Neovim has similar API functions `nvim_win_call` and `nvim_buf_call`

PR #2030, opened 2026-05-14
Fix missing bufnr

schnittchen:
I am encountering the error `nui.nvim/lua/nui/tree/init.lua:188: missing bufnr` coming from `lua/neo-tree/ui/renderer.lua:891`.

The error is intermittent, and I can resume after hitting enter, but it's quite annoying.

The error is triggered when

* launching nvim with `.` as command line argument
* then switching to another directory, like `:e lib/`
* with `hijack_netrw_behavior = "open_current"` (unsure if this is needed to trigger the bug)

I tried to manually pinpoint the bug, but my knowledge of nvim and lua was not sufficient. With the help of a stochastic parrot, iterating with debug prints to pinpoint the issue, a plausible fix was produced.

I read the contrib guidelines, but `mise deps` failed compiling openssl, so I skipped all the linting etc.. It's only a small patch, so this should not be a big issue I hope.

The first commit is the real fix, the second is more a defensive measure.

### DIFF
--- a/lua/neo-tree/setup/netrw.lua
+++ b/lua/neo-tree/setup/netrw.lua
@@ -39,6 +39,14 @@ M.hijack = function()
   utils.debounce("hijack_netrw_" .. dir_window, function()
     local manager = require("neo-tree.sources.manager")
     local log = require("neo-tree.log")
+    -- The hijack may have been triggered by a BufEnter that fired in a transient
+    -- window that no longer exists (this can happen during `:e <dir>` when the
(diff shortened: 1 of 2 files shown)


pynappo:
LGTM, we've actually had a test that's been logging that error but the test didn't fail so i never bothered to look at it haha. this seems to fix that error so thanks!

ci timeout is because i think mise updated so their new deps keyword shadows our deps task, no worries.

PR #2032, opened 2026-05-15
feat(document_symbols): add ignore_symbols option to filter symbol tree

### LINKED ISSUE #2031: document_symbols: add option to filter by symbol kind
When opening a file with many functions and variables mixed together, it's hard to quickly scan for functions.

Telescope already has `ignore_symbols`, aerial.nvim has `filter_kind`, VS Code Outline supports right-click filter by type. neo-tree's `document_symbols` doesn't have this.

Something like Telescope's API would be great:

```lua
document_symbols = {
  ignore_symbols = { "Variable", "Field" },
}
```

Ref: #879 (document_symbols tracking issue)

---

sth like

Image

jensenojs:
Allows hiding specific LSP symbol kinds (e.g. Variable, Field) from the document_symbols tree by matching against kinds.get_kind().name. Configured at source level:

    document_symbols = {
      ignore_symbols = { "Variable", "Field" },
    }

Matching is case-insensitive and recursive. Symbols without kind information are never filtered. Non-string config values are silently ignored to prevent runtime errors.

to close https://github.com/nvim-neo-tree/neo-tree.nvim/issues/2031

### DIFF
--- a/tests/neo-tree/sources/document_symbols/filter_symbols_spec.lua
+++ b/tests/neo-tree/sources/document_symbols/filter_symbols_spec.lua
@@ -1,0 +1,102 @@
+pcall(require, "luacov")
+
+local symbols = require("neo-tree.sources.document_symbols.lib.symbols_utils")
+
+local function make_sym(name, kind_name, children)
+  children = children or {}
+  return {
+    id = name,
+    name = name,
+    type = "symbol",
+    path = "/test.lua",
+    children = children,
+    extra = {
+      kind = { name = kind_name, icon = "?", hl = "" },
+    },
+  }
+end
+
+describe("filter_symbols", function()
+  it("returns nil when ignore is nil", function()
+    local syms = { make_sym("foo", "Function") }
+    local got = symbols._filter_symbols(syms, nil)
+    assert.are.same(1, #got)
+  end)
+
+  it("returns all when ignore is empty", function()
+    local syms = { make_sym("foo", "Function") }
+    local got = symbols._filter_symbols(syms, {})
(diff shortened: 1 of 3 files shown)


jensenojs:
图片

---

Thank you in advance for your time and effort; I hope you like this feature. Please feel free to let me know if there are any other adjustments you would like me to make. @pynappo

tests/neo-tree/sources/document_symbols/filter_symbols_spec.lua
```diff
@@ -0,0 +1,102 @@
...
+    local got = symbols._filter_symbols(syms, { variable = true })
+    assert.are.same(0, #got)
+  end)
+
+  it("handles case-insensitive matching", function()
+    local syms = {
+      make_sym("f", "Function"),
+      make_sym("v", "Variable"),
+    }
+    local got = symbols._filter_symbols(syms, { VARIABLE = true })
```
pynappo: clarify behavior here because this test case doesn't pass. I think the simplest fix to is remove this test case, and document that ignore must have lowercase keys for to properly take effect in filter_symbols. config.ignore_symbols should be copied and normalized to have lowercase keys as well.

pynappo (changes requested):

tests/neo-tree/sources/document_symbols/filter_symbols_spec.lua
jensenojs: Thank you so much for your time and effort. I’ll take care of this as soon as possible after work.

pynappo (approved):

PR #2034, opened 2026-05-15
fix: ci task + branch desync

pynappo:
accidentally pushed the wrong commits to v3.x. I don't have the right permissions to fix the branches with a force-push though so this is the next-best option


pynappo:
actually i think the best way to fix this is to just make a new v4 branch eventually lol. i'll just deal w/ this for now

PR #2036, opened 2026-05-22
fix(fs_actions): clear buffers when trashing

akioweh:
delete does this so i thought trash should do too?

not super sure about `stat`ing files in a loop for the `trash_nodes` variant but my thoughts were if some individual files failed to be trashed (there doesn't seem to be any communication of partial-failures i believe)?

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/fs_actions.lua
+++ b/lua/neo-tree/sources/filesystem/lib/fs_actions.lua
@@ -801,6 +801,7 @@ M.trash_node = function(path, callback, state)
     if state and restorefunc then
       table.insert(state.undostack, restorefunc)
     end
+    clear_buffer(path)
     internal_hooks.on_file_deleted(path, callback)
   end)
 end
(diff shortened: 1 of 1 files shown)


akioweh:
on a side note: the `trash_node` vs `trash_nodes` logic seem quite asymmetric...? e.g. the multi-node function does not fire pre-delete events and the undostack and post-delete event code are still ran even if `trash.trash` failed

pynappo:
I genuinely forgot to put this in when adapting the delete code lol so ty, I agree that this should be the expected behavior.

> the multi-node function does not fire pre-delete events and the undostack and post-delete event code are still ran even if trash.trash failed

Currently, trash.trash returns false upon any failure, partial or total. so yeah, stat'ing the files in a loop is currently the intended solution

this is also why the undostack and post-trash event code are still ran in the multi node variant - if trash trashed any files at all, it should add something to the undostack and fire some amount of events. in the single node variant I just assume that if the trash failed, there's no need to check further (but maybe i'm wrong on that?)

> (there doesn't seem to be any communication of partial-failures i believe)?

Yeah there probably should be some, tbh. Currently the supported trash backends that just use system commands will just log the output of the commands as warnings if they return non-zero error codes, but having more output wouldn't hurt in the case of these partial failures.

Will merge in a bit

pynappo (approved):

Issue #2028, opened 2026-05-04
BUG: Showing document_symbols does not initially reveal cursor position when follow_cursor is set to true


benward2301:
### Did you check docs and existing issues?

- [x] I have read all the docs.
- [x] I have searched the existing issues.
- [x] I have searched the existing discussions.

### Neovim Version (nvim -v)

NVIM v0.11.1

### Operating System / Version

Linux Mint 21.1

### Describe the Bug

With `follow_cursor` set to true, opening the `document_symbols` source (e.g. via `:Neotree show document_symbols`) does not reveal the symbol corresponding to the cursor position in the LSP buffer. Instead, the symbol tree remains collapsed until the cursor is moved.

I've fixed it with the following change and can open a PR:
```diff
diff --git a/lua/neo-tree/sources/document_symbols/init.lua b/lua/neo-tree/sources/document_symbols/init.lua
index 1a544a3..ec29f79 100644
--- a/lua/neo-tree/sources/document_symbols/init.lua
+++ b/lua/neo-tree/sources/document_symbols/init.lua
@@ -120,6 +120,19 @@ M.setup = function(config, global_config)
       event = events.VIM_CURSOR_MOVED,
       handler = follow_debounced,
     })
+    manager.subscribe(M.name, {
+      event = events.AFTER_RENDER,
+      handler = function(state)
+        if state.name ~= M.name or not state.tree or not state.lsp_winid then
+          return
+        end
+        local cursor = vim.api.nvim_win_get_cursor(state.lsp_winid)
+        local node_id = symbols.get_symbol_by_loc(state.tree, { cursor[1] - 1, cursor[2] })
+        if #node_id > 0 then
+          renderer.focus_node(state, node_id, true)
+        end
+      end,
+    })
   end
 
   -- Set up follow_tree_cursor: show symbol on cursor move in document_symbols buffer
```

### Steps to Reproduce

In your neo-tree config, set `follow_cursor` to true for the `document_symbols` source:
```lua
require('neo-tree').setup {
  document_symbols = {
    follow_cursor = true
```
Then, in a buffer with an LSP attached, run `:Neotree show document_symbols`.

### Expected Behavior

The symbol corresponding to the LSP buffer cursor position should be revealed when the symbol tree is opened.

### Your Configuration

```Lua
-- Prerequisite: lua-language-server installed and on $PATH.

local repro_root = vim.fn.stdpath('cache') .. '/neo-tree-repro'
[...]

PR #2041, opened 2026-05-31
fix(document_symbols): reveal cursor position on open

benward2301:
See issue https://github.com/nvim-neo-tree/neo-tree.nvim/issues/2028


benward2301:
@pynappo Thanks for the commit. I had tried using `follow_debounced`, but did not pass `afile`, so just duplicated instead. This is much better and works on my machine.

pynappo:
Thanks for verifying. i was gonna improve the behavior a bit further (currently, this doesn't do the follow if you switch to document_symbols while focusing Neo-tree) but i'll leave that for a future PR. Feel free to keep your issue open so i don't forget

PR #2047, opened 2026-06-05
fix(git): switch ignored files resolver to git-ls-files for consistency

Thaumy:
Commit ea7491ef3686b68cb1fbdb2d65bc1aeeb8ef0bb0 introduced a performance optimization using git-check-ignore to speed up resolving ignored files in large repositories. However, git-check-ignore behaves differently from git-status when dealing with directories that contain only ignored files.

With the previous optimization, such directories would initially be displayed (due to check-ignore logic) but would disappear upon reopening neo-tree (which relies on git status results). This inconsistency is confusing for users.

I measured the execution time of the relevant git commands emitted by neo-tree in a large repository (Linux kernel source tree):

- git status: ~230ms
- git ls-files: ~120ms
- git check-ignore: ~40ms

(The real command is very long so I only show the git sub command type.)

While git check-ignore is the fastest, git ls-files is still significantly faster than git status while maintaining behavioral consistency with git-status.

This removes check-ignore.lua, and switches the ignored files resolver to use git-ls-files for a simple fix.


pynappo (approved):
makes sense, tysm

PR #2053, opened 2026-06-23
fix(filesystem): follow_current_file logic error

julianbuettner:
Neotree did not `follow_current_file` in my neovim setup anymore, same thing happening to a colleague using Neotree.

I let a LLM investigate and apparently the bug is that "not" is stronger binding than comparison, so the line

```
not winpos == "current"
--equals
(not winpos) == "current"
-- equals
false == "current"
```

The suggested fix solved the problem for me.


pynappo (approved):
i'll merge b/c obviously the logic i wrote there was wrong, but for future reference can you write the reproduction steps needed to reproduce the issue? would be nice to have a test case to cover any future regressions

PR #2074, opened 2026-08-10
fix(git): discover nested repositories without the file watcher

ardaaltinors:
Nested repository discovery in `on_directory_loaded` sits inside the
`if state.use_libuv_file_watcher then` branch, so a folder holding a `.git`
entry is only registered as its own worktree when the file watcher is enabled.
`use_libuv_file_watcher` defaults to `false`, so under the default config a
nested repository is never discovered — every path below it resolves against
the outer worktree, and its files are reported as ignored whenever the outer
`.gitignore` excludes the directory. #2015 made resolution prefer the innermost
worktree, but with the watcher off the inner one never gets registered in the
first place.

To reproduce: an outer repo whose `.gitignore` contains `/inner/`, with an
independent repo checked out at `inner/`. On default settings every file under
`inner/` renders as gitignored; setting `use_libuv_file_watcher = true` fixes
it, which is what pointed at the coupling.

This moves the registration out of the watcher branch and gates it on
`enable_git_status`. Two details keep it from costing anything for people
without nested repositories:

- The tree root is skipped, since `filesystem.navigate` already runs a status
  for it. Without that it gets statused twice per fresh scan.
- The child has to be named exactly `.git` rather than merely end in `.git`, so
  a directory such as `mirror.git` no longer triggers a `git rev-parse` on
  every scan — it never becomes a registered worktree, so the existing lookup
  guard cannot suppress the repeat.

I counted git subprocesses on a plain repo with no nested repository (default
config, fresh open) and this branch matches `main` exactly: 2 `rev-parse`,
1 full `git status --ignored`, 1 `git status --untracked-files=no`, 1
`ls-files`. Same parity for a repo containing `sub/mirror.git/` and for a repo
opened through a symlink.

One thing worth flagging: `possible_worktree_roots` is drained only by
`mark_gitignored`, which `job_complete` skips when `hide_gitignored = false`,
so entries for nested repos can sit there unconsumed. That is pre-existing and
bounded — the `not git.worktrees[target_path]` guard stops appending once
registration completes — but this change makes it reachable on the default
config rather than only with the watcher on. Happy to address it here or
separately if you'd prefer.

Added `tests/neo-tree/sources/filesystem/git_nested_worktree_spec.lua`, which
runs both watcher settings. Against `main` the two `use_libuv_file_watcher =
false` cases fail and the two `= true` cases pass; with this change all four
pass. Full suite goes from 91 to 95 passing, no failures or errors.

### DIFF
--- a/lua/neo-tree/sources/filesystem/lib/fs_scan.lua
+++ b/lua/neo-tree/sources/filesystem/lib/fs_scan.lua
@@ -20,12 +20,10 @@ local M = {}
 --- how many entries to load per readdir
 local ENTRIES_BATCH_SIZE = 1000
 
----@type string[]
-local possible_worktree_roots = {}
-
 ---@param cwd string Filters what worktrees to view. Should be an ancestor of all the items.
 ---@param items neotree.FileItem[]
-local mark_gitignored = function(cwd, items)
+---@param possible_worktree_roots string[] Roots of newly-discovered nested repositories found while scanning.
+local mark_gitignored = function(cwd, items, possible_worktree_roots)
   if vim.in_fast_event() then
     log.warn("mark_gitignored cannot be called in a fast event", debug.traceback())
     return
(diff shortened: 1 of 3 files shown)


pynappo:
i'll take a quick look at what's going on w/ the windows test here, otherwise good changes ty

PR #2076, opened 2026-08-13
fix(document_symbols): make lsp_winid follow the current window

Jecrevin:
## Summary

When splitting the window (`:split`) or moving between windows with `<C-w>`, `state.lsp_winid` (the window whose symbols are shown) was not updated to follow the cursor. Worse, closing the tracked window with `<C-w>q` left a dangling window id that caused errors in `jump_to_symbol`, `show_symbol`, and `follow_symbol`.

## Changes

- `document_symbols/init.lua`
  - Added `sync_lsp_window()` which syncs `lsp_winid`/`lsp_bufnr`/`path` to the window the cursor is currently in (skipping the neo-tree window, floating windows, and non-real-file buffers).
  - Added `get_active_state()` to look up an open source state without creating a new one.
  - Subscribed to `VIM_WIN_ENTER` so `lsp_winid` follows the current window after `:split` / `<C-w>` window switches.
  - Subscribed to `VIM_WIN_CLOSED` to clear `lsp_winid`/`lsp_bufnr` when the tracked window is closed.
  - `M.navigate` now prefers the current window instead of always re-deriving via `get_appropriate_window`.
  - Guarded `follow_symbol` against an invalid `lsp_winid`.
- `document_symbols/commands.lua`
  - Added `get_valid_lsp_winid()` used by `jump_to_symbol` and `show_symbol`, which re-derives a window when the tracked one is invalid instead of erroring.
- Tests: added `lsp_winid` describe block covering window-switch following and the `<C-w>q` close case.

All `document_symbols` tests pass.


pynappo:
i'm surprised nobody else filed an issue/PR to fix this sooner, because this seems like obvious stuff the source should have done before. ig people (like me) don't use the source that much

i'll take a closer look tmrw but looks good to me at the moment. apologies for the wait.

i think i changed some internal types around to reflect some edge cases so the lua diagnostics fail now, i can fix it for you when i review

pynappo:
improved behavior on win_closed event slightly to allow commands to reopen the lsp buffer in a new window (e.g. if you :q all windows except neo-tree, neo-tree used to be useless, now you can reopen the last tracked buffer from neo-tree)

